PR #7, opened 2026-06-23
Overhauls codegen, Pt 1

jtoman:
Four major changes:
1. Resurrect main.py in the form of a proper "console-codegen"
2. Asyncify all of our code
3. New CEX handling API, uses an agentic analyzer for the codegen
4. Adds dedicated CEX remediation agent (codegen agent was *awful* at writing CVL fixes)


composer/prover/core.py
```diff
@@ -65,26 +80,113 @@ def make_prover_options(*, cloud: bool) -> ProverOptions:
...
+
+@dataclass(frozen=True)
+class FailingRule:
+    """A logical rule that failed verification, possibly with multiple
+    parametric instances. The ``rule_name`` is the base rule (``path.rule``);
+    ``instances`` lists the per-binding ``RuleResult`` records (one per
+    parametric arg, each with its own CEX trace). Non-parametric rules
+    have ``len(instances) == 1``.
+
+    Per-rule grouping is the natural scope for an analyzer's scratchpad —
```
shellygr, line 130: Unless you use assertions in hooks or cvl functions used in summaries. In these cases you want grouping by location of where the hook was invoked. It could be we don't generate such specs often

composer/prover/core.py
```diff
@@ -65,26 +80,113 @@ def make_prover_options(*, cloud: bool) -> ProverOptions:
...
+# CEX-analysis protocol
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class FailingRule:
+    """A logical rule that failed verification, possibly with multiple
+    parametric instances. The ``rule_name`` is the base rule (``path.rule``);
+    ``instances`` lists the per-binding ``RuleResult`` records (one per
+    parametric arg, each with its own CEX trace). Non-parametric rules
```
shellygr, line 127: Invariant base case, transient storage case also considered parametric args?

composer/prover/core.py
```diff
@@ -94,40 +196,121 @@ async def on_prover_link(self, link: str) -> None: pass
...
+        callbacks: CexProgressCallbacks,
+        report_dir: Path,
     ) -> str:
-        ...
+        async def _one(instance: RuleResult) -> tuple[RuleResult, str | None]:
+            await callbacks.on_analysis_start(instance)
+            analysis = await analyze_cex_raw(
+                self.llm, self.state["messages"], instance, tool_call_id
+            )
+            if analysis is not None:
```
shellygr, line 273: What happens if the analysis is None? What is the impact?

composer/prover/agentic_analyzer.py
```diff
@@ -0,0 +1,655 @@
...
+        callbacks: CexProgressCallbacks,
+        report_dir: Path,
+    ) -> str:
+        # Cluster the violated instances by rule name — the unit the
+        # per-rule analyzer iterates with a shared scratchpad. (Derived
+        # here rather than passed in: the handler interface only carries
+        # the full result set; grouping is opt-in per handler.)
+        failing_rules = group_failing(all_results)
+
+        # fs tools scoped to the report's ``inputs/.certora_sources``
```
shellygr, line 394: For cloud runs you assume we fetch zip outputs? This is a bit expensive, also in RAM, and while I don't mind the path structure, it's important to use POU latest fetch sources API to avoid zip download and extraction

composer/templates/cex_analyzer_aggregator_system.j2
```diff
@@ -0,0 +1,157 @@
...
+structural invariant linking `totalSupply` to the sum of `balances`.
+**Cause B** (rule `pauseBlocksTransfers`): The failure happens because the
+`paused` flag's persistence across the call is not modeled — it is havoced at
+the external call boundary.
+
+Expected result: Different constructs (a missing structural invariant on supply
+vs. a havoc on the `paused` flag at an external call), different directions of
+fix -> **separate partitions**.
+
+### Example 3
```
shellygr, line 90: The trickiest case is two root causes related to slightly different invariants. This should also result in separate partitions . E.g sum of balances over all tokens vs sum of balances over a single token

composer/templates/cex_analyzer_per_rule_system.j2
```diff
@@ -0,0 +1,137 @@
...
+output leads to the failed specification, root cause analysis is identifying
+the step where the execution "went wrong".
+
+During your analysis, be sure to take into account the CVL commands and features used
+by the rule. In particular, the rule may use "ghosts", "hooks", or "summaries", all
+of which control how the Prover reasons about the smart contract's execution, and thus
+the counter example it produces. You should use the `cvl_researcher` subagent to help
+understand the input CVL.
+
+**Important**: The Prover is a deeply complex tool, and you do not have any real understanding
```
shellygr: May dedup parts here shared with the other prompt

shellygr (approved):

composer/prover/core.py
jtoman: Yeah, you and I discussed using failing assert location as a better grouping strategy. There are different heuristics we can play with here; but the fundamental logic here (heuritistic grouping, aggregation within that group, then cross failure aggregation) seems to be solid. Agree?

composer/prover/core.py
jtoman: UHHH. I forget how those get parsed as RulePath's let me double check.

composer/prover/core.py
jtoman: LLM gets an empty counter example explanation. Honestly, this is an argument of moving everything to the agentic cex analyzer; this case becomes impossible.

composer/prover/agentic_analyzer.py
jtoman: I don't assume, I know we do as I wrote it that way. now that POU is OSS, porting to use that instead of my hand rolled `anonymousKey` based methods is probably better. One note; it would be ideal if we could make POU async friendly out of the box, rather than having to wrap in `to_thread` everywhere.

composer/templates/cex_analyzer_aggregator_system.j2
jtoman: That's what example 2 is pointing towards right? Not sure if there was something actionable here of if you're just agreeing.

composer/prover/core.py
shellygr: it is - although initially I thought that's a lot of code to facilitate the goal - how do we know the initial grouping is actually saving us time? maybe a single pass on all failures is better?

composer/prover/core.py
shellygr: also generic preserved vs specific preserved and we recently added constructor preserved too, and there are strong invariants where we assert the invariant again in every call site

composer/prover/agentic_analyzer.py
shellygr: fair point. adding to the wishlist

composer/templates/cex_analyzer_aggregator_system.j2
shellygr: no, it's a havoc cause vs missing invariant cause.
I meant a case where the two causes are due to missing invariants which are similar but not the same

jtoman: oh you're right, I was conflating two examples (across two different prompts) in my head, I'll add this example

composer/prover/core.py
jtoman: I guess we don't, my hypothesis is that if there a lot of failures, it's likely many will share the same root cause, and rather than have the LLM re-derive that root cause in parallel across all rules, do the "cheap" matching in our heuristic grouping. I will say that the old version did just fan out all the cex analyses in parallel, and then we had an adhoc pass to summarize root causes later. In practice, we would see something like 30 counter examples that collapsed down to three (or so) "fix these" points.

Now that we're doing the agentic analysis each of those CEX analyses is quite expensive (reading files, doing CVL research) so happy path "this is the same error" via this iterative process vs one final aggregation seems to be a big win. I confess that I never actually tried the "all parallel, one final aggregation".

composer/prover/core.py
jtoman: preserved et al aren't individual results though? they only constrain how each parametric case is proved right?

jtoman: added an integration test. It takes around 5 minutes to run, so I've added it to a nightly test that will ping us on slack if it breaks. keeps the "land PR" path snappy and quick.

tests/test_rule_skips.py
```diff
@@ -130,7 +130,7 @@ def _is_result_rejection(st: StateWithSkips) -> bool:
 
 
 # =========================================================================
-# Prover report handling
+# Prover report handlingq
```
shellygr: handlingq

pyproject.toml
```diff
@@ -74,23 +99,67 @@ conflicts = [
...
+console-autoprove = "composer.cli.console_autoprove:main"
+tui-autoprove = "composer.cli.tui_autoprove:main"
 console-foundry = "composer.cli.console_foundry:main"
 tui-foundry = "composer.cli.tui_foundry:main"
 autoprove-report-render = "composer.spec.source.report.render:main"
 tui-natspec = "composer.cli.tui_pipeline:main"
 cache-natspec = "composer.cli.cache_natspec:main"
 console-codegen = "composer.cli.console_codegen:main"
 tui-codegen = "composer.cli.tui_codegen:main"
+# certora_autosetup entry points (merged in from the AutoSetup repo).
```
shellygr: remove parenthesis

shellygr (approved):
tiny nits

PR #8, opened 2026-06-23
Codegen Overhaul, Pt 2

jtoman:
* Untangling some setup code
* better store discipline (no more splatting into top-level NS)
* modern resource management (standard connections et al)
* Refactoring out common "validation" logic into a common "chassis" module


composer/spec/cex_remediation.py
```diff
@@ -491,7 +497,7 @@ async def run(self) -> list[str | dict]:
     prebuilt = (
         bind_standard(builder, _RemediationState)
         .with_input(_RemediationInput)
-        .with_tools([mem_tool, read_system_document])
+        .with_tools([mem_tool, read_system_document, critic, ERC20TokenGuidance.as_tool("erc20_guidance")])
```
shellygr, line 500: is the cex_remediation tool designed to be CVL/EVM specific?

shellygr (approved):
a question only

composer/spec/cex_remediation.py
jtoman: Yes.

shellygr (approved):

PR #9, opened 2026-06-23
Codegen Overhaul, Pt 3: More cleanup, refactoring

jtoman:
* Remove dedicated audit db, move it to store
* takes this opportunity to finally centralize around a single file/document abstraction
* better recovery of in progress VFS
* removes trace html generation, no longer served by any live DB


composer/audit/store.py
```diff
@@ -0,0 +1,473 @@
+"""
+Audit archive backed by the LangGraph store.
```
shellygr, line 2: can you be specific what you refer to by "audit" here

shellygr (approved):
reviewed only the last commit. as a refactor it looks straightforward

composer/audit/store.py
jtoman: yeah, "audit" means "audit what the agent did" or "audit runs" not audit in the security certora audit sense.

PR #10, opened 2026-06-23
Code gen changes, Pt 4

jtoman:
Multi spec. Allow multiple spec files to gate the generated code. Adjust internal data structures to match; no CLI entry point exists as the only producer (the assistant) no longer exists.


composer/templates/system_prompt.j2
```diff
@@ -16,4 +17,281 @@ verification. As an extension, you know that simple, non-optimized code has a gr
...
+
+The task is complete only when:
+1. Every registered CVL spec file is independently stamped as VERIFIED via its own
+   `certora_prover(target_spec=<spec_vfs_path>)` run on the current code state.
+2. Every natural-language requirement is judged SATISFIED or LIKELY by the requirements
+   judge oracle.
+
+Follow this procedure:
+
+## Step 0 — Orient
```
shellygr, line 62: this is not a direct copy of the synthesis_prompt is it? the rewrite seems much more significant, is it just to support multiple spec files?

composer/templates/system_prompt.j2
```diff
@@ -16,4 +17,281 @@ verification. As an extension, you know that simple, non-optimized code has a gr
...
+Do NOT invent CVL syntax based on guesses. If you're not sure something is valid CVL, ask
+the researcher first.
+
+{% include "doc_ref_author.j2" %}
+
+## Code style for verifiability
+
+The high-level frame: **prover incompleteness manifests as timeouts; the antidote is
+isolation + summarization.** Write code so the awkward parts have a natural seam where a
+summary can attach.
```
shellygr: I wonder if we have a more thorough explanation of the rely-guarantee concept around in these prompts, as it seems a more likely candidate for the llm to pick up and understand (separation of concerns, modules, trust-but-verify assumptions) rather than adjectives such as "awkward" and "natural seam".

composer/tools/prover.py
```diff
@@ -98,7 +103,19 @@ class CertoraProverTool(WithAsyncDependencies[Command, ProverDeps], WithInjected
...
-    use_working_spec : bool = Field(description="Use the working copy of the spec instead of the master copy.")
+    target_spec: Optional[str] = Field(default=None, description=(
+        "The VFS path of the committed spec file to verify against. Required "
+        "when use_working_spec is false; the prover reads this spec from the VFS "
+        "and records a per-spec completion stamp on success. Ignored when "
+        "use_working_spec is true (the transient working draft has no VFS path "
+        "until committed, and runs against it never stamp)."
+    ))
+
+    use_working_spec : bool = Field(description=(
```
shellygr, line 114: question about the old code - in which case it is allowed to modify the spec?

composer/templates/system_prompt.j2
jtoman: No, I've rewritten it to be more like the other prompts

composer/tools/prover.py
jtoman: Only when the user approves. But I found that when the agent kept trying different mitigations it got REALLY annoying to approve each attempt at a fix. hence the temp "working spec" buffer which the agent can freely edit, and then we get an edit request only at the end.

shellygr (approved):

PR #13, opened 2026-06-24
Explicit CVL restoration

jtoman:
Instead of wasting tokens and attention asking the agent to put the CVL back via put cvl just ... do it for them.


composer/spec/cvl_generation.py
```diff
@@ -396,10 +396,11 @@ async def run_cvl_generator[S: CVLGenerationState, C: FeedbackToolContext, I: CV
 ) -> S:
     input_copy = in_state["input"].copy()
     last_attempt = await ctx.child(LAST_ATTEMPT_KEY).cache_get(_LastAttemptCache)
+    in_state_copy = in_state.copy()
     if last_attempt is not None:
-        input_copy.append("Your last working draft was (you will need to re-put this onto the VFS):")
+        input_copy.append("Your last working draft on this task is below; it has been put into your CVL buffer (no need to re-put it)")
```
shellygr: you probably don't need the parenthesis anymore

shellygr (approved):

composer/spec/cvl_generation.py
jtoman: yeah, it's preemptive prompting around anticipated behavior, I'll drop it.

PR #15, opened 2026-06-25
[Autosetup] Detect and warn the user if the main contract is abstract

jar-ben:
## Detect and warn when the main contract is abstract

When the main (verify-target) contract is abstract or an interface, it compiles to
no bytecode and can't be verified. AutoSetup previously failed with an opaque
`Compilation analysis failed` / `not among the files in the prover scene`, giving no
indication of the cause.

### Changes
- **`compilation_workarounds.py`**: detect the terminal no-bytecode case for the main
  contract (match the prover's `Contract X has no bytecode` against the verify target)
  and raise `AbstractMainContractError` with an actionable message, short-circuiting
  AutoSetup's import-patch retry (which can't recover an abstract contract).
- **`enhanced_config_manager.py`**: add `ConfigManager.extract_main_contract_from_config`
  to pull the verify-target name from a config dict.
- **`setup_prover.py`**: clarify the foundry-path "not in scene" error to call out
  abstract/interface (no bytecode) as the likely cause.

### Result
The user gets _"Main contract 'X' compiled to no bytecode: it is abstract … re-run with
a concrete implementation"_ instead of a generic failure.


certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -175,6 +183,20 @@ def log(self, message: str, level: str = "INFO") -> None:
         else:
             logger.info(message)
 
+    def _detect_abstract_main_contract(self, output: str, compilation_config: Dict) -> Optional[str]:
+        """Return the main (verify-target) contract name if the compiler reported it has
+        no bytecode — i.e. it is abstract (or lacks a constructor) and cannot be verified.
+
+        The prover prints ``Contract <name> has no bytecode`` for any contract in the
+        input files (harnesses/siblings included), so match the reported names against
+        the verify target specifically rather than assuming the first one is the main.
```
shellygr, line 192: what if an auxiliary contract is abstract? we still fail in this case, no?

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -391,6 +413,16 @@ def run_compilation_with_workarounds(
 
             # If no workaround applies, exit immediately (guardrail against infinite loop)
             if not workaround_applied:
+                # Terminal, non-recoverable case: the main (verify-target) contract has
```
shellygr: did you intent to put this block  in line 377, after we saw that returncode != 0 in the first retry? currently you check this only after all workarounds

certora_autosetup/utils/compilation_workarounds.py
jar-ben: There is a PR in progress (or perhaps already merged) that as a last resort drops a contract from `files` if we are not able to fix its compilation. Except for the main contracts of course. So, yes, you are right - we fail also if we specify an abstract contract in `files`, but we will have (in some situations perhaps acceptable) workaround for that. Also, note that the auxiliary contracts are selected either by the user (here AIComposer) via autosetup args or they are collected from foundry/hardhat. In the latter case, I believe we filter abstract contracts. And in the former case, I hope that AIComposer does not provide abstract contracts as auxiliary

certora_autosetup/utils/compilation_workarounds.py
jar-ben: It could be good to move it up, yes. will do that.

shellygr (approved):
not sure why I didn't approve before. good to ship

PR #16, opened 2026-06-25
track prover usage

naftali-g:
Track prover usage in Autosetup, and in AutoProver.

Note this uses `prover_api.get_statsdata(job_identifier)["run_id"]["start_to_end_time"]` which has the actual runtime, without upload + queued time

### DIFF
--- a/certora_autosetup/utils/prover_runner.py
+++ b/certora_autosetup/utils/prover_runner.py
@@ -296,27 +296,17 @@ class ProverRunner(ABC):
         """Get cache key for job specification."""
         return job_spec.get_cache_key(self.config_manager)
 
-    def _record_fresh_prover_runtime(self, job_identifier: Optional[str]) -> None:
-        """Fold a freshly-executed prover run's prover-REPORTED runtime (statsdata
-        ``run_id.start_to_end_time``, in ms) into the process prover-usage ledger.
+    def _record_prover_runtime_seconds(self, seconds: Optional[float]) -> None:
+        """Fold a freshly-executed prover run's runtime (in seconds) into the process
+        prover-usage ledger. ``None`` or non-positive is a no-op.
 
-        ``job_identifier`` is whatever ``ProverOutputAPI`` resolves to statsdata for
-        this runner — the emv-* folder path (local) or the cloud job URL (cloud).
-        Called from each runner's leaf executor, on the fresh-run path only: cache
-        hits short-circuit before the executor runs, so cached jobs are excluded —
-        they consumed no prover compute this run. Best-effort: any failure to locate /
-        read statsdata is a silent no-op (a run that crashed before producing statsdata
-        simply isn't counted)."""
-        if not job_identifier:
-            return
-        try:
-            stats = self.prover_api.get_statsdata(job_identifier)
-            # statsdata records every metric as a list of samples; start_to_end_time
-            # always carries exactly one — the run's total.
-            runtime = stats["run_id"]["start_to_end_time"][0]
-            record_prover_runtime_ms(int(runtime))
-        except Exception as e:
-            self.log(f"Could not record prover runtime for usage ledger: {e}", "DEBUG")
+        Cloud passes the server-reported runtime (job start→finish); local passes its
+        wall-clock duration — local runs are serialized (one prover at a time, no
+        queueing), so wall-time is the run time. Callers are on the fresh-run path
+        only: cache hits short-circuit before the executor runs, so cached jobs are
+        excluded — they consumed no prover compute this run."""
+        if seconds and seconds > 0:
+            record_prover_runtime_ms(int(seconds * 1000))
 
     async def _check_cache(self, cache_key: str, job_spec: ProverJobSpec) -> Optional[ProverResult]:
         """Check if we have a cached result for this cache key."""
--- a/certora_autosetup/utils/prover_runner.py
+++ b/certora_autosetup/utils/prover_runner.py
@@ -33,6 +33,7 @@ from .enhanced_config_manager import ConfigManager, FileContent, ProverJobSpec
 from .file_utils import atomic_write_json_fsspec
 from .job_problem_fixes import MAX_JOB_PROBLEM_FIXES, on_job_problem
 from .logger import logger
+from .prover_usage_ledger import record_prover_runtime_ms
 from .runner_types import (
     JobHandle,
     NotificationType,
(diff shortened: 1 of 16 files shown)


certora_autosetup/utils/prover_runner.py
```diff
@@ -295,6 +296,28 @@ def _get_cache_key(self, job_spec: ProverJobSpec) -> str:
...
+        this runner — the emv-* folder path (local) or the cloud job URL (cloud).
+        Called from each runner's leaf executor, on the fresh-run path only: cache
+        hits short-circuit before the executor runs, so cached jobs are excluded —
+        they consumed no prover compute this run. Best-effort: any failure to locate /
+        read statsdata is a silent no-op (a run that crashed before producing statsdata
+        simply isn't counted)."""
+        if not job_identifier:
+            return
+        try:
+            stats = self.prover_api.get_statsdata(job_identifier)
```
shellygr: for cloud runs, any reason you do not use `get_job_info`? it gives you start and finish time as two values, so you need to compute, but the plus side is that for partially successful jobs where statsdata.json was not generated but some rule results are available you do get the running time (it does not download a file, but takes from the cloud own metadata, while also not including upload and queue time, as we wanted).

for local runs it's okay to use statsdata... but then you can also take wall-time, I'm not mistaken if we run locally we're limiting to a single Prover instance anyway.

composer/diagnostics/timing.py
```diff
@@ -82,6 +83,12 @@ class RunSummary:
     prover_total_s: float = 0.0
     prover_total_calls: int = 0
     _active_prover_by_task: dict[str, tuple[float, int]] = field(default_factory=dict, repr=False)
+    prover_reported_ms_total: int = 0
```
shellygr: aren't the comments and fields upside down? isn't line 87 documenting line 86 and line 91 documenting line 90?

composer/prover/results.py
```diff
@@ -209,6 +212,27 @@ def read_and_format_run_result(s: Path) -> dict[str, RuleResult] | str:
...
+    """The prover's self-reported run time for this job, in milliseconds.
+
+    Read from ``<results_root>/Reports/statsdata.json`` (``run_id.start_to_end_time``)
+    — the prover engine's own start-to-end wall time, present and identical for cloud
+    (the extracted results archive) and local runs. This is NOT composer's client-side
+    ``elapsed`` (which also covers cloud queue / polling / result download). The value is
+    stored as a single-element list. Returns ``None`` on any failure (file absent,
+    malformed JSON, or the key missing) — usage capture must never break a prover run.
+    """
+    stats_path = results_root / "Reports" / "statsdata.json"
```
shellygr: ditto
I also don't understand why we try to read also from zipOutput? I think we should phase zipOutput usage out as much as possible

composer/spec/source/autosetup.py
```diff
@@ -248,3 +251,27 @@ def read_autosetup_usage(project_root: Path) -> list[TokenUsageDict]:
...
+    Returns ``None`` on any failure (file absent — autosetup skipped, full cache hit,
+    crash, replayed snapshot, or an AutoSetup too old to emit it — or malformed JSON):
+    missing external usage must never break the phase."""
+    usage_file = resolve_autosetup_prover_usage_file(project_root)
+    if usage_file is None:
+        return None
+    try:
+        return int(json.loads(usage_file.read_text())["ms"])
+    except (OSError, ValueError, KeyError, TypeError) as e:
+        _logger.debug(f"Could not read AutoSetup prover usage from {usage_file}: {e}")
```
shellygr: maybe info? warning?

shellygr (commented):
Overall approach is solid, but we should consider twice before resorting to statsdata.json for two reasons:
1. there's an alternative API that gives the same value even if a job did not produce statsdata, but we still know the run was 'useful'
2. avoiding yet another case of dependency on zipOutput

composer/diagnostics/timing.py
naftali-g: don't know why claude keeps doing this

shellygr (approved):

composer/diagnostics/timing.py
```diff
@@ -157,6 +172,21 @@ def token_usage_summary(self) -> dict[str, object]:
             ],
         }
 
+    def prover_usage_summary(self) -> dict[str, object]:
+        """Serializable run-total / per-phase prover-REPORTED runtime — the body of
```
jtoman: why do you keep SHOUTING reported?

composer/diagnostics/timing.py
```diff
@@ -157,6 +172,21 @@ def token_usage_summary(self) -> dict[str, object]:
...
+        ``prover_usage.json`` and of the ``prover_usage`` run tag. Runtime is the
+        prover's own start-to-end time (statsdata ``run_id.start_to_end_time``),
+        summed across every prover run (cloud and local alike), in milliseconds.
+        Minutes are trivially ``ms / 60_000`` and not stored."""
```
jtoman: no one cares claude, stop adding these pointless comments

composer/prover/cloud.py
```diff
@@ -106,6 +107,23 @@ async def poll_job(
     """
     return await asyncio.wait_for(_poll_job_inner(job, interval=interval, on_status=on_status), timeout=timeout)
 
+def _job_runtime_ms(job_data: dict) -> int | None:
+    """Prover execution time (ms) from the cloud job's ``startTime``→``finishTime`` —
```
jtoman, line 111: don't you mean "Prover-REPORTED execution time"? :P

composer/prover/core.py
```diff
@@ -194,6 +194,10 @@ async def on_prover_link(self, link: str) -> None: pass
     """Fires once per run as soon as the prover emits its result link. For
     cloud runs ``link`` is the prover UI URL; for local runs it is a
     filesystem path to the results directory."""
+    async def on_prover_runtime(self, ms: int) -> None: pass
```
jtoman, line 198: this comment is a lie, you only report the prover runtime on clouds, the local handling of statsdata doesn't appear to be implemented.

That said, you could just have the `local_cm` context manager do that parsing and yield the time too, I'm not sure why you aren't.

composer/prover/results.py
```diff
@@ -209,6 +212,28 @@ def read_and_format_run_result(s: Path) -> dict[str, RuleResult] | str:
         to_ret[r.name] = r
     return to_ret
 
+
+def read_prover_runtime_ms(results_root: Path) -> int | None:
```
jtoman: is this still used? It seems like it would be what you'd want to call in the local result context manager right??

composer/spec/source/autosetup.py
```diff
@@ -248,3 +251,27 @@ def read_autosetup_usage(project_root: Path) -> list[TokenUsageDict]:
         return []
 
     return [_to_token_usage(model, bucket) for model, bucket in by_model.items()]
+
+
+def read_autosetup_prover_usage(project_root: Path) -> int | None:
+    """Prover-REPORTED runtime (milliseconds) AutoSetup's subprocess prover runs
```
jtoman: hey there it is!

composer/prover/core.py
```diff
@@ -442,20 +449,26 @@ async def run_prover(
...
+    # into a tmpdir whose lifetime is bound to ``cloud_results``; local runs hand
+    # back a stable path. Either way the analyzer must complete before the
     # context manager exits.
-    async with results_cm as emv_path:
+    async with results_cm as (emv_path, cloud_runtime_ms):
         parsed = read_and_format_run_result(emv_path)
+        # Queue-free run time: cloud uses the job's startTime->finishTime (execution
+        # window, excludes queue); local has no cloud job and isn't queued, so it reads
+        # the engine's own start_to_end from statsdata.
+        runtime_ms = cloud_runtime_ms if prover_opts.cloud else read_prover_runtime_ms(emv_path)
```
jtoman: oh wait, you're reading it here. Why not just do this computation in `_local_results`? Is this to work around a parsing failure of job data in the cloud and fall back on parsing out of the local unzip?

jtoman (commented):
nits and questions

composer/diagnostics/timing.py
naftali-g: I was sure I erased that already

composer/prover/core.py
naftali-g: yeah doing that. also getting rid of the statsdata, will just use wallclock for local runs.

jtoman (approved):
thanks !

PR #19, opened 2026-06-26
De-couple composer from AutoSetup's on-disk contract

shellygr:
Resolves the `composer/spec/source/autosetup.py` `TODO: improve this once the repos are merged`.

composer hard-coded AutoSetup's output layout (`.certora_internal`, `autosetup_result.json`, `.CertoraProverLiteReports`, `llm_usage.json`) and re-derived the usage-file path. Now that `certora_autosetup` lives in this repo, it becomes the single source of truth.

### DIFF
--- a/composer/spec/source/autosetup.py
+++ b/composer/spec/source/autosetup.py
@@ -20,6 +20,8 @@ from composer.prover.core import ProverOptions
 
 from graphcore.utils import TokenUsageDict
 from composer.io.context import emit_custom_event
+# Locator for autosetup's on-disk llm_usage.json (certora_autosetup owns that layout).
+from certora_autosetup.utils.paths import resolve_autosetup_llm_usage_file
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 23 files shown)


composer/spec/source/autosetup.py
```diff
@@ -276,7 +235,12 @@ def read_autosetup_usage(project_root: Path) -> list[TokenUsageDict]:
     crash, replayed snapshot, or an AutoSetup too old to emit it — or malformed
     JSON): missing external usage must never break the phase.
     """
-    usage_file = _resolve_autosetup_usage_file(project_root)
+    # Local import: importing certora_autosetup.utils at module load would instantiate its logger
+    # singleton (a RotatingFileHandler) as a side effect. This runs only at runtime, so import here.
+    # (Not a cycle — certora_autosetup never imports composer.)
```
shellygr:
```suggestion
```

composer/spec/source/autosetup.py
```diff
@@ -276,7 +235,11 @@ def read_autosetup_usage(project_root: Path) -> list[TokenUsageDict]:
     crash, replayed snapshot, or an AutoSetup too old to emit it — or malformed
     JSON): missing external usage must never break the phase.
     """
-    usage_file = _resolve_autosetup_usage_file(project_root)
+    # Local import: importing certora_autosetup.utils at module load would instantiate its logger
+    # singleton (a RotatingFileHandler) as a side effect. This runs only at runtime, so import here.
+    from certora_autosetup.utils.paths import resolve_autosetup_llm_usage_file
```
naftali-g: So it gets instantiated here? is that better?

shellygr: Not really, I was assuming there's some race I'm not aware of and we're importing the module in some very particular point which is not here. I will see if I can move the logger instantiation anyway because module side effects are cursed.

naftali-g (approved):
lgtm

PR #21, opened 2026-06-26
End to End Integration tests

jtoman:
Adds an end-to-end integration test, based off an (updated) counter.py harness tape. The prover and database calls are real, the LLM calls are all mocked via the harness tape mechanism. The actual test simply asserts "nothing crashed"; no attempt is made (yet!) to ensure that files/reports ended up where they need to be. That's easy enough to add later, but it's worth getting this landed now. The setup required for getting the integration test env mocked out is unfortunate; (perhaps the memory backend should do its own DDL a la the RAG DB) and all the database provisioning is equally sadge, but for now, it works.

Because these tests are (relatively) long they run as a cron nightly at 3:07 UTC; they will send a message on slack to our dev channel to let us know (on top of sending whatever emails that may or may not be enabled).

### DIFF
--- a/tests/test_autoprove_integration.py
+++ b/tests/test_autoprove_integration.py
@@ -61,8 +61,11 @@ _SUMMARIES_REL = "specs/summaries/Counter_base_summaries.spec"
 
 
 async def _fake_autosetup_phase(*_args, **_kwargs) -> SetupSuccess:
-    """Stand in for the AutoSetup subprocess, which makes its own un-tapeable LLM
-    calls. Writes out the (trivial, no-op) summaries spec the generated CVL imports
+    """Stand in for the AutoSetup subprocess, which makes its LLM
+    calls that we, in autoprover land, aren't going to start taping.
+    It is also not an intersting unit of test for this workflow, so just use the
+    trivial, precomputed setups.
+    Writes out the (trivial, no-op) summaries spec the generated CVL imports
     against, then returns the config AutoSetup would have produced for Counter."""
     summaries = _SCENARIO / "certora" / _SUMMARIES_REL
     summaries.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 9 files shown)


tests/test_autoprove_integration.py
```diff
@@ -0,0 +1,192 @@
...
+    "solc": "solc",
+    "verify": "Counter:certora/specs/sanity-Counter.spec",
+    "wait_for_results": "none",
+}
+# AutoSetup's summaries spec, relative to certora/ (the SetupSuccess contract).
+_SUMMARIES_REL = "specs/summaries/Counter_base_summaries.spec"
+
+
+async def _fake_autosetup_phase(*_args, **_kwargs) -> SetupSuccess:
+    """Stand in for the AutoSetup subprocess, which makes its own un-tapeable LLM
```
shellygr: The calls could be taped. The reason is that we are not testing autosetup, so it's reasonable to assume a setup exists

tests/test_autoprove_integration.py
```diff
@@ -0,0 +1,192 @@
...
+        "// Auto-generated base summaries for Counter\n// No summaries needed for Counter\n"
+    )
+    return SetupSuccess(
+        prover_config=dict(_COUNTER_PROVER_CONFIG),
+        summaries_path=_SUMMARIES_REL,
+        user_types=[],
+    )
+
+# graphcore's Postgres memory backend doesn't self-create its schema; this mirrors
+# the memories_fs DDL in graphcore/tests/conftest.py (keep in sync if that moves).
```
shellygr, line 82: Brittle, no? Can we hoist?

shellygr (approved):
Good basis for starting....

tests/test_autoprove_integration.py
jtoman: Yes, I'll have to break a memory tools API (of which we are the only users so no great loss)

jtoman: I'm actually going to push this to a later task of removing a bunch of our old, now unused async code, which will let us get rid of some code that will make this refactoring easier.

tests/test_autoprove_integration.py
jtoman: yeah I'll adjust the comment.

PR #22, opened 2026-06-28
Handle cloud prover job failures gracefully instead of crashing the run

jar-ben:
## What
A failed cloud prover job (FAILED/CANCELED/HALTED) currently
raises an uncaught `RuntimeError` in `cloud_results`, which propagates through
`run_prover` → `verify_spec` and aborts the entire AutoProver pipeline — producing no final report.

## Changes
- `cloud.py`: add `CloudJobError`; `cloud_results` raises it (not bare
  `RuntimeError`) on a non-SUCCEEDED terminal status, poll timeout, or missing
  output archive.
- `core.py`: `run_prover` catches `CloudJobError` and returns a message — like
  its other prover-failure paths — so the agent gets feedback and can continue (without crashing the whole pipeline).
- Source terminal-status detection from POU's `JobStatus` enum
  (`convert_job_status`) instead of a hand-maintained string set, fixing drift
  (misspelled `CANCELLED`, missing `HALTED`/`SERVICE_UNAVAILABLE`/`UPLOAD_FAILED`,
  bogus `ERROR`/`TIMEOUT`).

## Why
Two real autoprover runs died this way during invariant-CVL
generation when a generated spec's prover job returned `FAILED`.

### DIFF
--- a/composer/prover/core.py
+++ b/composer/prover/core.py
@@ -479,7 +479,7 @@ async def run_prover(
                     diagnoses=[],
                 )
     except CloudJobError as exc:
-        return f"Prover cloud job did not produce results (status {exc.status.value}). Prover link: {exc.link}"
+        return f"Prover cloud job did not produce results (status {exc.status.value})."
 
     prover_report: dict[str, bool] = {}
     for i in parsed.values():
--- a/composer/prover/core.py
+++ b/composer/prover/core.py
@@ -41,7 +41,7 @@ from graphcore.graph import LLM
 from prover_output_utility import cloud_server_for_env
 
 from composer.prover.analysis import analyze_cex_raw
-from composer.prover.cloud import cloud_results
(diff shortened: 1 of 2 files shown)


composer/prover/core.py
```diff
@@ -449,34 +449,37 @@ async def run_prover(
...
+                    all_results, tool_call_id, callbacks, emv_path
+                )
+            else:
+                result_str = load_jinja_template(
+                    "rule_feedback.j2",
+                    rule_entries=[(r, None) for r in all_results],
+                    diagnoses=[],
+                )
+    except CloudJobError as exc:
+        return f"Prover cloud job did not produce results (status {exc.status.value}). Prover link: {exc.link}"
```
jtoman: this message gets returned to the agent right? Is including the link doing anything?

jtoman (approved):
one nit

PR #23, opened 2026-06-28
Widen the langgraph checkpointer/store DB connect window at startup

jar-ben:
## What

The langgraph checkpointer and store open their first Postgres connection at startup (in `checkpointer_context` / `store_context` → `setup()`). If that first connection is slow to establish, the pool's connection-acquire times out and the whole run fails before it begins.

This widens the connect window:

- `connect_timeout=10` on the connection string — bounds each individual connect attempt so a stuck connect fails fast.
- `timeout=120` on the checkpointer/store pools — gives `getconn` a long enough window to keep retrying the connect (each attempt bounded by `connect_timeout`) before giving up, covering a slow first connection.

## Why

A slow first connection at startup would otherwise surface as an acquire timeout and abort the run. The longer acquire window lets the pool retry the connect transparently. As a side effect it also lets a brief connection blip mid-run be ridden out rather than failing the run.

## Notes

The pool is `min_size=1, max_size=1`, so the acquire `timeout` also bounds how long a mid-run acquisition waits on a wedged connection before erroring — an accepted trade-off for the added startup/blip tolerance.

### DIFF
--- a/composer/workflow/services.py
+++ b/composer/workflow/services.py
@@ -138,6 +138,12 @@ def _adapt_async(obj: T, pairs: list[tuple[str, str]]) -> T:
         setattr(obj, async_name, new_async_method)
     return obj
 
+# Bound each connect attempt; give getconn a long window to keep retrying so a
+# slow first connection doesn't fail the run.
+_DB_CONNECT_TIMEOUT_SECONDS = 10
+_DB_POOL_ACQUIRE_TIMEOUT_SECONDS = 120.0
+
+
 def _get_composer_connection_string(
      *,
     user: str,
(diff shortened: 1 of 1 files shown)


composer/workflow/services.py
```diff
@@ -138,6 +138,12 @@ async def async_wrapper(*args, **kwargs):
         setattr(obj, async_name, new_async_method)
     return obj
 
+# Bound each connect attempt; give getconn a long window to keep retrying so a
+# slow first connection doesn't fail the run.
+_DB_CONNECT_TIMEOUT_SECONDS = 10
+_DB_POOL_ACQUIRE_TIMEOUT_SECONDS = 120.0
```
jtoman: I guess this is fine, but good question is why it's taking 2 minutes (?!?!?!?) to connect to a database in the same AWS enclave/conclave/whatever

H00N24: It's just for dev, as we're using auto-scaling Postgres with minimum of 0 ACUs the initial call can take some time (consequent scaling is quite fast). For stg/prod we're scaling from 0.5 ACUs, thus we won't have this issue.

jtoman: oic.

jtoman (approved):
approving from a code perspective. I'll leave the verdict on the reasonableness of the timeouts to @H00N24

H00N24 (approved):
LGTM, now thinking about it there is also an option that we would add a specific error code for this error and we would add a AWS Batch retry on this error.

PR #24, opened 2026-06-28
A tale of two fixes

shellygr:
1. For code without user-defined data types, the summary resolver would crash since it would send an empty block to the LLM. Handle this case to avoid it triggering. **Point to consider: make a more global fix in graphcore?**
2. Autosetup failed on a second run on the same project when parsing *_call_resolution.spec due to ASTExtraction output mishandling, trying to parse non-json as json. Context: It is part of the code that comments out irrelevant auto-injected summaries.

### DIFF
--- a/certora_autosetup/setup/summary_resolver.py
+++ b/certora_autosetup/setup/summary_resolver.py
@@ -118,7 +118,7 @@ def _parse_ast_payload(stdout: str, returncode: int, stderr: str) -> Optional[di
     """
     lines = stdout.splitlines()
     # Only LEADING diagnostic lines precede the JSON; never strip inside the payload.
-    while lines and (lines[0].startswith("Warning: ") or lines[0].startswith("Error: ")):
+    while lines and lines[0].startswith(("Warning: ", "Error: ")):
         lines.pop(0)
     body = "\n".join(lines).strip()
     # Empty stdout (or stdout that was nothing but diagnostics) means the jar gave us
--- a/certora_autosetup/setup/summary_resolver.py
+++ b/certora_autosetup/setup/summary_resolver.py
@@ -38,8 +38,6 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Sequence, Set, Tuple
 
-from certora_cli.Shared.certoraUtils import find_jar
-
 from certora_autosetup.parsers.method_parser import MethodParser
 
 # A logger callable matching SummarySetup.log(message, level).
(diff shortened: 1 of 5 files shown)


composer/spec/source/summarizer.py
```diff
@@ -251,10 +264,7 @@ def _validator(s: ST, _res: str) -> str | None:
             typechecked="",
             plan=None,
             curr_spec=None,
-            input=[
-                "The following types are available for use in your spec",
-                udts,
```
jtoman, line 256: I know that sending an empty string as the *entire* message payload could cause an error, but an empty string in the messages array is a new one. agreed on the fix at the graphcore level, something we can fold into the "invoke" wrappers I've added in https://github.com/Certora/graphcore/pull/21

jtoman (approved):
approving the AIC parts, leaving AS parts to my betters

composer/spec/source/summarizer.py
shellygr: have you applied this in graphcore#21? it seems not?

certora_autosetup/setup/summary_resolver.py
```diff
@@ -105,6 +105,36 @@ def _vmtype_to_str(vmtype: dict) -> str:
...
+    diagnostic lines to stdout *before* the JSON payload, which it then prints
+    exactly once as the final output (on both the success and syntax-error paths).
+    So we drop the leading diagnostic lines and parse the remainder — rather than
+    ``json.loads`` the whole stream, which fails at char 0 on the first warning.
+    Returns the decoded payload, or ``None`` when the jar produced no AST
+    (``ast`` is null — a hard syntax error left for the TypecheckerLoop backstop).
+    """
+    lines = stdout.splitlines()
+    # Only LEADING diagnostic lines precede the JSON; never strip inside the payload.
+    while lines and (lines[0].startswith("Warning: ") or lines[0].startswith("Error: ")):
```
naftali-g:
```suggestion
    while lines and lines[0].startswith(("Warning: ", "Error: ")):
```

certora_autosetup/setup/summary_resolver.py
```diff
@@ -105,6 +105,36 @@ def _vmtype_to_str(vmtype: dict) -> str:
...
+    # Empty stdout (or stdout that was nothing but diagnostics) means the jar gave us
+    # no payload — surface that with its stderr rather than an opaque JSONDecodeError.
+    if not body:
+        raise RuntimeError(f"{_AST_EXTRACTION_JAR} produced no JSON (exit {returncode}): {stderr.strip()}")
+    try:
+        payload = json.loads(body)
+    except json.JSONDecodeError as exc:
+        raise RuntimeError(
+            f"{_AST_EXTRACTION_JAR} output was not JSON after stripping diagnostics "
+            f"(exit {returncode}): {exc}; stdout_head={stdout[:300]!r} stderr={stderr.strip()[:500]!r}"
```
naftali-g: just print all stdout and stderr, no?

naftali-g (approved):
a nit and a comment, but approving to not block

composer/spec/source/summarizer.py
jtoman: not yet, since I didn't want to make this PR rely on a fix that hasn't landed yet. But there's no reason not to do it now anyway...

jtoman: done

PR #27, opened 2026-06-29
Generic Pipeline Refactor

jtoman:
The Foundry and Prover pipelines were basically structurally the same (system analysis -> properties -> formalization loop) but due to some backend specific changes they were two separate implementations with *significant* overlap.

This PR fixes this issue. There is now a unified "core" pipeline, parameterized over the `PipelineBackend` type. The `PipelineBackend` type describes the task phases to use, the application types, and various callbacks invoked by the core pipeline. In particular, PipelineBackend exposes an entry point called `prepare_system` which returns a `PreparedSystem` encoding the finalized application model to be used for the rest of the pipeline. In the prover pipeline, this callback will perform the harnessing steps; in the foundry pipeline it does nothing beyond return the `PreparedSystem` object.

The `PreparedSystem` object itself exposes another callback: `prepare_formalization` which is run in parallel with property inference. This phase is responsible for (asynchronously) preparing any setup for the formalization later. In the prover, this callback is responsible for invoking autosetup, invariant generation, and CVL summarization. Foundry is once again a no-op.

The result of `prepare_formalization` is another object called `Formalizer`. This object exposes a `formalize` method, which is responsible for the actual formalization (foundry or prover) and returning the result according to the protocol `BackendResult`. The core pipeline then handles report generation and result construction.

NB we chose three separate objects `PipelineBackend`, `PreparedSystem`, and `Formalizer` to make staging dependencies explicit. `Formalizer` is only produced after `prepare_system` is chained into `prepare_formalization`. The type safety is worth the hoop jumping in this author's humble opinion.

Other changes of note:
* Removes the slug de-duplication logic for components; instead, we push this up to the component generation phase and tell the LLM to pick different names in the (unlikely!) event the slugification of component names leads to collisions
* Removes the `methods` field of `PropertyFormulation` as nothing was consuming it. NB this is backwards compatible with any serializations that exist in databases; pydantic will happily ignore "unknown" fields.
* Generalizes the `ArtifactStore` logic and pushes more logic into the shared super class. This allows the pipeline core to interact with the artifact store using a uniform API (rather than the bespoke entry points exposed by `Foundry/ProverArtifactStore`.
* Centralizes around a single `GaveUp` type defined by the pipeline core
* Moves the foundry generation away from using the `CVLJudge` and `CVLGeneration` marker classes (lol)
* Moves some type definitions around so they are more "central" (reflecting their role).


composer/pipeline/core.py
```diff
@@ -0,0 +1,361 @@
...
+    *,
+    interactive: bool = False,
+    threat_model: Document | None = None,
+    max_bug_rounds: int = 3,
+) -> CorePipelineResult[FormT]:
+    spec, phases = backend.analysis_spec, backend.core_phases
+    source = run.source
+
+    # 1. System analysis (shared primitive, backend-parameterized; always yields SourceApplication).
+    analyzed = await run.runner(
```
shellygr, line 161: how well does this driver serves us if we wanted to cache system analysis results only and then dispatch both foundry and prover pipelines?

composer/pipeline/core.py
```diff
@@ -0,0 +1,361 @@
...
+
+    # 3. Pre-formalization setup runs CONCURRENTLY with extraction (neither needs the other) —
+    #    this preserves the prover's autosetup ∥ bug-analysis overlap, generically.
+    formalizer_task = asyncio.create_task(prepared.prepare_formalization(run))
+
+    batches = await _extract_all(prepared.main, backend.backend_guidance, run,
+                                phases["extraction"], interactive, threat_model, max_bug_rounds)
+    formalizer = await formalizer_task
+    if not batches:
+        raise ValueError("No properties extracted from any component.")
```
shellygr, line 187: I'd love to understand some time what's our policy regarding error handling here and whether we use `ValueError` and others in specific ways

tests/test_autoprove_integration.py
```diff
@@ -55,6 +55,8 @@
     "solc": "solc",
     "verify": "Counter:certora/specs/sanity-Counter.spec",
     "wait_for_results": "none",
+    "server": "production",
+    "prover_version": "master"
```
shellygr: making sure - on purpose?

composer/pipeline/core.py
jtoman: great question; we can make them share system analyses if we make them use the same cache key (currently one is called "system-analysis-foundry", the other called "system-analysis") and if they share the same root cache key (which I believe they do). Of course, there is some issue with raciness, but honestly I'm not going to lose too much sleep about that

composer/pipeline/core.py
jtoman: yeah! Great question. Sometimes we throw, sometimes we return exceptions (in the gather). Sometimes its a custom exception time, sometimes its ValueError/RuntimeError. We absolutely need to clean this up. I can fold that in to this PR, as its already a "touches everything" PR...

tests/test_autoprove_integration.py
jtoman: indeed not, workaround for a local run on my machine -_-

composer/pipeline/core.py
shellygr: I think it's worth the effort; but maybe then we should allow running system analysis as its own task separately and then have both pipelines enjoy the cached system analysis

jtoman:
pinging @naftali-g for his thoughts on my rework of the design doc stuff

naftali-g (commented):
design doc changes look fine to me, I'll let Shelly approve the whole thing

shellygr (approved):

PR #29, opened 2026-06-29
Makes composer portable

jtoman:
Abstract over the backend model provider via the new `composer.llm` module. The actual BaseChatModel construction is handled via the `ModelProvider` type, which also carries the provider kind. There is a bunch of wiring, dispatch, and feature detection login in this module.

The bigger change in this PR actually comes from the memory tool and the tool display.

First, we previously scraped anthropic's (custom) tool call content blocks (openai calls them "function calls") which completely broke display when using chatgpt. We instead use the (standardized) `tool_calls` property that LC provides for us.

Next, the memory tool we were sending to Claude was a custom name that claude was fine tuned on. OpenAI had no idea what to do with this custom tool name, and the schema that Claude used was actually pretty awful to formalize as a schema. Accordingly, we changed graphcore to export a "real" memory tool, and then have the memory backend generator returned by standard connections handle the choice of the right tool from graphcore.

One final wrinkle: the `content` type of the LC's Messages is `str | list[str | dict]` but this is a lie; openai does not allow you to have plain strings in a list, only dicts. SO we reintroduce our good friends `ainvoke` and `invoke` in graphcore, and use those to normalize around this problem.

The rest of the changes in this PR are just changing the plumbing of services to no longer assume "always anthropic".


jtoman:
Other big changes while reading this:

* more accurate live token prices
* changes to the file API to account for the shape of "uploaded file reference" blocks

composer/llm/anthropic.py
```diff
@@ -0,0 +1,173 @@
...
+        model_kwargs = (
+            {"cache_control": {"type": "ephemeral", "ttl": ttl}} if ttl else {}
+        )
+
+        return ChatAnthropic(
+            model_name=self.model_name,
+            max_tokens_to_sample=opts.tokens,
+            temperature=1,
+            timeout=None,
+            max_retries=8,
```
shellygr, line 172: 1. do we want to make any of these configurable? 
2. is it okay timeout isn't set?

composer/ui/message_renderer.py
```diff
@@ -41,43 +42,177 @@ def dot(style: str, text: Text | str) -> Text:
...
+# pricing. Once an individual call's input crosses this, the long
+# tier applies *for the full session* per OpenAI's terms; we
+# approximate that by switching on a per-message basis (a session
+# that drifts above 272K will mostly stay there).
+_OPENAI_LONG_CONTEXT_THRESHOLD = 272_000
+
+
+# Pricing tables transcribed from Anthropic + OpenAI rate cards.
+# Sources should be re-checked when new model families ship.
+_PRICING: list[tuple[str, _ModelPricing]] = [
```
shellygr, line 90: are we sure it's the only place in the code where we hold pricing information?

shellygr (commented):
2 nits and then ready to approve

composer/ui/message_renderer.py
jtoman: UH, I don't know any other place in the AP code, there might be something in AS I'm not aware of

composer/llm/anthropic.py
jtoman: they're hardcoded currently. I think when we've set timeout in the past anthropic's API flakiness has bit us.

shellygr (approved):

PR #30, opened 2026-06-30
Fix ommer_contracts property name in application context

oppenheim:
## Summary

- Renames `ContractComponentInstance.ommer_contract` to `ommer_contracts` to match `application_context_new.j2`.
- Restores the sibling-contract context block in prompts that include `application_context_new.j2` (property analysis, CVL generation, property judge, Foundry, etc.).

## Why

`application_context_new.j2` references `context.ommer_contracts`, but the Python model exposed `ommer_contract`. Jinja resolves the missing attribute as `Undefined`, which is falsy, so the entire "other contracts in the application" section was silently skipped. Agents still had source tools to discover sibling contracts, but lost structured metadata (descriptions, sorts, and paths in update mode) that system analysis had already produced.

## Test plan

- [x] Verified locally that rendering `application_context_new.j2` with a multi-contract `ContractComponentInstance` now includes the sibling contracts section.
- [ ] Run an AutoProve pipeline on a multi-contract project and confirm sibling contract names appear in bug-analysis / CVL-generation prompts.

Made with [Cursor](https://cursor.com)


composer/spec/system_model.py
```diff
@@ -233,7 +233,7 @@ def contract(self) -> ExplicitContract:
         return self._contract.contract
     
     @property
-    def ommer_contract(self) -> list[ExplicitContract]:
+    def ommer_contracts(self) -> list[ExplicitContract]:
```
jtoman, line 217: 🤦

jtoman (approved):

PR #31, opened 2026-06-30
Scene-timed summary emission via a unified `on_contracts_entered_scene`

jar-ben:
## Problem

Summaries were matched, emitted, and pruned only **upfront**, for the main contract and its
additional contracts. A library reached only *transitively* — via a contract pulled into the scene
later by call resolution — was therefore missed, or its summary referenced a contract not yet in the
conf scene and got permanently commented out by the typechecker loop. The two emission paths
(initial setup vs. the call-resolution hook) had also drifted: they ran the same steps in different
orders, with a separate append-only aggregator class and a one-shot upfront aggregator writer.
Several CVL-generation issues compounded this — reserved words used as parameter names, NONDET on
reference-return functions, and a duplicate-spec-path that gutted curated specs shared by multiple
registry keys.

## Change

- **One emission path.** Curated + LLM emission, the aggregator refresh, and the prune pass are
  unified into a single `SummarySetup.on_contracts_entered_scene`, used for both the initial scene
  and every batch added during call resolution. Curated matching runs first so the LLM step skips
  methods a curated summary already covers; the aggregator and prune are refreshed once per batch.
- **`run` → `configure`.** The old `run` conflated init with the first summarization. It is now a
  pure `configure` (file resolution, name parsing, settings capture); the caller drives
  `on_contracts_entered_scene` for the initial scene, mirroring the call-resolution path exactly.
- **State-based aggregator.** `generate_base_aggregator` and the separate `BaseSummariesAggregator`
  are replaced by `_add_aggregator_imports` / `_rewrite_aggregator`: one writer that re-emits the
  aggregator sorted from accumulated state. `summary_aggregator.py` is deleted.
- **Generation robustness.** Escape CVL reserved words in parameter names (declaration + dispatch);
  skip NONDET for reference-return functions with an `AUTO-DISABLED` breadcrumb.
- **Prune correctness.** Dedup the prune spec list (multiple registry keys can share one file); cache
  the `all_methods.json` parse; and make the AST-extraction step tolerate the jar's leading
  `Warning:` diagnostic lines (e.g. one per `using` alias) by parsing from the first JSON line.

A `TODO(precision)` records a known follow-up: emission is scoped to each contract's whole
compilation unit, not to the methods reachable from the main contract, so a contract pulled in as a
candidate link/dispatch target can get summaries for functions that are never called (the prover
reports these as "unused"). It is sound and the prover cost is negligible; the only cost is LLM
analysis of unreachable methods. Scoping to the reachable set is left as a follow-up.

## Validation

Ran autosetup end-to-end on real multi-contract projects via the CI job-report harness: curated +
LLM summaries register through the unified path, union-template specs are pruned to the functions
present in scene, no "contract not found" / "does not appear in code" aborts, no duplicate imports,
and each contract is analyzed exactly once across overlapping call-resolution paths. Summary unit
tests pass.


certora_autosetup/setup/setup_summaries.py
```diff
@@ -61,6 +61,27 @@
...
+    "ALL", "ALWAYS", "ASSERT_FALSE", "AUTO", "CONSTANT", "Create", "DELETE", "DISPATCH", "DISPATCHER",
+    "HAVOC_ALL", "HAVOC_ECF", "NONDET", "PER_CALLEE_CONSTANT", "STORAGE", "Sload", "Sstore", "Tload",
+    "Tstore", "UNRESOLVED", "as", "assert", "assuming", "at", "axiom", "builtin", "default",
+    "definition", "description", "else", "event", "exists", "expect", "fallback", "false", "filtered",
+    "forall", "function", "ghost", "good_description", "havoc", "hook", "if", "import", "in", "indexed",
+    "invariant", "lastReverted", "lastStorage", "links", "mapping", "methods", "new", "norevert", "old",
+    "onTransactionBoundary", "override", "persistent", "preserved", "require", "requireInvariant",
+    "reset_storage", "return", "returns", "revert", "rule", "satisfy", "sig", "sort", "strong", "sum",
+    "true", "unresolved", "use", "using", "usum", "void", "weak", "with", "withrevert", "xor",
+})
```
naftali-g: This list is more restrictive than what we really support. see `usable_keywords` in the cvl.cup:
```
usable_keywords ::=
    EXISTS:k      {: RESULT = (String) k; :}
  | FORALL:k      {: RESULT = (String) k; :}
  | SUM:k         {: RESULT = (String) k; :}
  | USUM:k        {: RESULT = (String) k; :}
  | USING:k       {: RESULT = (String) k; :}
  | AS:k          {: RESULT = (String) k; :}
  | IMPORT:k      {: RESULT = (String) k; :}
  | USE:k         {: RESULT = (String) k; :}
  | BUILTIN:k     {: RESULT = (String) k; :}
  | OVERRIDE:k    {: RESULT = (String) k; :}
  | SIG:k         {: RESULT = (String) k; :}
  | DESCRIPTION:k {: RESULT = (String) k; :}
  | INVARIANT:k   {: RESULT = (String) k; :}
  | PRESERVED:k   {: RESULT = (String) k; :}
  | WEAK:k        {: RESULT = (String) k; :}
  | STRONG:k      {: RESULT = (String) k; :}
  | ON_TRANSACTION_BOUNDARY:k   {: RESULT = (String) k; :}
  | OLD:k         {: RESULT = (String) k; :}
  | HOOK:k        {: RESULT = (String) k; :}
  | UNRESOLVED_LOWERCASE:k {: RESULT = (String) k; :};
```

naftali-g (approved):
I think the CVL_RESERVERD_WORDS is too restrictive, but other than that looks ok

PR #32, opened 2026-06-30
Add `CTHelpers.getCollectionId` curated summary

jar-ben:
## Problem

`CTHelpers.getCollectionId` (Gnosis Conditional Tokens) derives a collection id as a compressed
alt_bn128 curve point: it hashes its inputs to a curve point via a modular square root — a large
unrolled `mulmod` add-chain computing `x^((P+1)/4) mod P`. Left in the verification scene, loop
summarization folds that add-chain into a huge nested `mulmod`/`addmod` expression and exhausts
memory, so any contract that reaches `getCollectionId` cannot be verified.

## Change

A curated summary for `getCollectionId`:

- A registry entry in `function_summaries.json` (`ctHelpers_getCollectionId`) matching the function
  on the `CTHelpers` library.
- `certora/specs/summaries/CTHelpers.spec`: a **deterministic ghost** summary. `getCollectionId` is a
  deterministic function of its inputs, so an arbitrary deterministic ghost is a sound
  over-approximation, and it keeps the modular-sqrt add-chain out of the prover.

Injectivity is encoded via **left-inverse ghosts** rather than a 6-variable `forall`: asserting each
input component is recoverable from the id is equivalent to injectivity but quantifies only 3
variables and applies the ghost once, so the solver instantiates it linearly. The spec documents the
over-approximation boundary: it does not model the homomorphic composition of nested collections
(the `parentCollectionId` point addition), so properties asserting how collections compose under
that composition must keep the real implementation.

## Dependency / ordering

This is most useful together with scene-timed summary emission: when `getCollectionId` is reached
only transitively (through a contract pulled into the scene by call resolution), curated matching
must run at the contract-add event to emit it. That lands in the summary-subsystem PR — merge that
first, then rebase this onto master (the two changes touch disjoint files, so no conflicts).

## Validation

The ghost and injectivity axiom typecheck and apply as a summary; verified end-to-end on a project
where `getCollectionId` is reached transitively — the modular-sqrt blowup no longer reaches the
prover and the run completes.

### DIFF
--- a/certora_autosetup/certora/specs/summaries/CTHelpers.spec
+++ b/certora_autosetup/certora/specs/summaries/CTHelpers.spec
@@ -1,0 +1,36 @@
+// Curated summary for the Gnosis Conditional Tokens helper library (CTHelpers).
+//
+// CTHelpers.getCollectionId derives a collection id as a compressed alt_bn128 curve point: it hashes
+// (conditionId, indexSet) to a curve point via a modular square root (the private CTHelpers.sqrt, a
+// large unrolled mulmod add-chain computing x^((P+1)/4) mod P) and, when parentCollectionId != 0,
+// EC-adds the parent point. The function is a deterministic function of its inputs, so a deterministic
+// ghost summary is a valid over-approximation. Summarizing it also keeps the modular-sqrt add-chain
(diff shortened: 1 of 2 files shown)


naftali-g (approved):
lgtm

PR #34, opened 2026-07-01
Fix Yul stack-too-deep detector to survive solc line-wrapping

shellygr:
## Problem

An AutoProver run (`Certora/tokemak-v2-core-fv` @ `b0104fb9`) failed in the Autosetup **compilation-analysis** phase with `CompilationAnalysisError`, after only two workaround attempts:

1. `stack_too_deep_via_ir` fired (enabled `solc_via_ir_map[LMPStrategyInstance1]=true`) — recompile then failed with the **via-ir** form of the error, `YulException: … Stack too deep`.
2. The loop reported `⚠️ no specific workaround match — enabling use_relpaths_for_solc_json` (the catch-all), then `❌ no applicable workaround` → import-patch fallback → same error → give up.

The `yul_exception_add_optimizer` workaround (which would add `solc_optimize=200` alongside via-ir) **never fired**, so the optimizer was never tried.

## Root cause

`_detect_yul_exception_stack_too_deep` used:

```python
pattern = r"YulException:.*Stack too deep"
re.search(pattern, output, re.IGNORECASE)
```

solc hard-wraps its diagnostic text at a fixed width, so the real message splits the phrase across a newline:

```
YulException: Variable param_0 is 2 slot(s) too deep inside the stack. Stack too
deep. Try compiling with `--via-ir` ...
```

Without `re.DOTALL` (and with a literal space between `too` and `deep`), the pattern cannot match `Stack too\ndeep`, so the detector returned `False` and the optimizer workaround was skipped.

## Fix

Match across arbitrary whitespace (incl. newlines):

```python
pattern = r"YulException:.*?Stack\s+too\s+deep"
re.search(pattern, output, re.IGNORECASE | re.DOTALL)
```

Now `yul_exception_add_optimizer` triggers on the wrapped error, and its existing paired safety net (`yul_exception_stack_too_deep`, which removes via-ir + optimizer and disables autofinders if the error persists) still applies unchanged.

## Tests

`tests/test_compilation_workarounds.py` — asserts detection on the **verbatim wrapped** solc output from the failing run, on a single-line variant, and non-detection on unrelated output. The old pattern returns `False` on the wrapped fixture; the new one returns `True`.

## Validation still to do

- [ ] Build the AutoProver container from this branch and re-run on `tokemak-v2-core-fv` to confirm the optimizer path is attempted (note: it's still possible viaIR+optimizer won't fit `LMPStrategyInstance1` within stack limits — this fix ensures the intended path runs, not that the contract necessarily compiles).

### DIFF
--- a/certora_autosetup/utils/compilation_workarounds.py
+++ b/certora_autosetup/utils/compilation_workarounds.py
@@ -493,9 +493,16 @@ class CompilationWorkaroundManager:
         return None
 
     def _detect_yul_exception_stack_too_deep(self, output: str) -> bool:
-        """Detect YulException with stack too deep error."""
-        pattern = r"YulException:.*Stack too deep"
-        return bool(re.search(pattern, output, re.IGNORECASE))
+        """Detect YulException with stack too deep error.
+
+        solc hard-wraps its diagnostic text at a fixed width, so the phrase
+        "Stack too deep" is frequently split across a newline (e.g. the real
+        error reads "...Stack too\ndeep."). Match across arbitrary whitespace
+        (DOTALL + ``\\s+`` between words) so the wrap does not hide the error;
+        otherwise the via-ir / optimizer workarounds never trigger.
+        """
+        pattern = r"YulException:.*?Stack\s+too\s+deep"
+        return bool(re.search(pattern, output, re.IGNORECASE | re.DOTALL))
 
     def _detect_compiler_version_mismatch(
         self, output: str, contracts: List[ContractHandle]
--- a/tests/test_compilation_workarounds.py
+++ b/tests/test_compilation_workarounds.py
@@ -1,0 +1,52 @@
+"""Unit tests for CompilationWorkaroundManager detectors."""
+
+from pathlib import Path
+
+import pytest
+
+from certora_autosetup.utils.compilation_workarounds import CompilationWorkaroundManager
+
+
+# Verbatim solc output captured from a real AutoProver run (tokemak-v2-core-fv):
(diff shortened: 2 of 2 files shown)


shellygr:
## Validated in a real container run ✅

Built `autoprover-local` from this branch (fix confirmed baked in: `re.DOTALL` + `Stack\s+too\s+deep`) and ran `certora-autosetup` against `tokemak-v2-core-fv` (main). On a genuine wrapped `YulException: … Stack too\ndeep` during the LMPVault compilation-analysis, the loop now escalates correctly:

```
attempt 2  stack_too_deep_via_ir        → via-ir on ChainlinkOracle
attempt 3  yul_exception_add_optimizer  → "Detected YulException stack-too-deep with via-ir — adding solc_optimize 200"
           (YulException did NOT persist afterward → optimizer cleared it; the step-10 revert never fired)
attempt 4  use_relpaths_for_solc_json   → unrelated error, loop continues
```

`attempt 3` is exactly the step the old detector could never reach — with the old regex the wrapped phrase was missed, so the loop fell straight to the catch-all and dead-ended. It triggered here on `ChainlinkOracle` rather than the originally-reported `LMPStrategyInstance1` (scene nondeterminism), but it's the identical error class and code path.

Also cross-checked with a deterministic simulation driving the real `run_compilation_with_workarounds` with the captured output sequence: old detector → `[stack_too_deep_via_ir, use_relpaths]` → fail; new detector → `[stack_too_deep_via_ir, yul_exception_add_optimizer, yul_exception_stack_too_deep]`.

jar-ben (approved):

PR #35, opened 2026-07-01
Add hardened forge to the local image

shellygr:
## Why

The local image had **no forge**. Two paths need it:

- The autoprove compilation setup shells out to `forge remappings` to get a project's complete remapping set. Without forge, autosetup falls back to solc-direct with only the checked-in `remappings.txt`, which can be incomplete (e.g. lib submodules with no explicit remapping) — so real projects fail compilation-analysis with `Source "…" not found`.
- `console-foundry` / `tui-foundry` run `forge test`.

## What

- New `foundry-builder` stage: builds `forge` from Certora's **public** foundry fork (pinned by `FOUNDRY_SHA`) in `rust:1-bookworm`, then the final stage `COPY`s the binary to `/usr/local/bin/forge` and smoke-checks `forge --version`. Bookworm base ⇒ glibc matches the `python:3.12-slim` runtime.
- The entrypoint enables the fork's hardening guards (`FOUNDRY_DISABLE_EXTERNAL_CHEATCODES=true`, `FOUNDRY_FFI=false`) for the foundry test modes (`console-foundry`/`tui-foundry`), where the project's own `forge test` executes untrusted code that can use FFI / external cheatcodes. Those guards only affect test/script execution; the autoprove path invokes forge only for `forge remappings` (a static config query with no cheatcodes/FFI), so it neither needs nor is affected by them.

## Notes

- The Rust build runs under QEMU on Apple Silicon (`--platform=linux/amd64`), so the first build is slow (~30–60 min); cached after. `FOUNDRY_SHA` bumps trigger a cold rebuild (no cargo cache mount — kept simple intentionally).


scripts/Dockerfile
```diff
@@ -16,6 +16,25 @@
...
 
+# Hardened forge from Certora's public foundry fork, built from source and
+# copied into the final stage. Used by the autoprove path (`forge remappings`
+# during compilation setup) and by console-foundry (`forge test`). The fork adds
+# opt-in FFI / external-cheatcode guards that the entrypoint enables for
+# foundry-mode test execution. Bump FOUNDRY_SHA to move the pin.
+# NOTE: the --platform=linux/amd64 pin means this Rust build runs under QEMU
+# emulation on Apple Silicon — the first build is slow (~30-60 min); cached after.
+ARG FOUNDRY_REPO=https://github.com/Certora/foundry.git
+ARG FOUNDRY_SHA=69cdf8ddc3d19ee85ab3b3c389376e0b07e1179b
```
jtoman, line 27: I'm confused about the semantics of declaring an argument with a value and then declaring it again later. If that's a known docker file idiom okay then

jtoman (approved):
With the caveat I don't really know how docker files work, this plausibly looks like it builds our hardened foundry in the docker file.

scripts/Dockerfile
shellygr: apparently it's global/local scope thing. (FROM forming a scope)

PR #37, opened 2026-07-03
Skip NONDET summaries for reference-type returns

shellygr:
skip analyzing with the llm methods that have a reference type as a return value, as those would be omitted later anyway.


certora_autosetup/setup/setup_summaries.py
```diff
@@ -1741,6 +1741,16 @@ async def analyze_with_llm(
...
+                # Skip reference-type (struct/array/bytes/string) returns for NONDET recipes:
+                # NONDET is unsound for reference types and certoraRun rejects it at typecheck
+                # time, so there's no point spending an LLM call on these candidates.
+                if recipe.summary_type == "NONDET" and self._returns_reference_type(method):
+                    self.log(
+                        f"Skipping {method['contractName']}.{method['name']} for NONDET "
+                        "(return type is a reference type - unsound for NONDET)"
+                    )
+                    continue
+
```
jar-ben, line 1753: having this filter now, do we still need the check later?
```
            if self._returns_reference_type(method):
                sig = self._get_function_signature(method, is_wildcard=False)
                content.append(
                    f"    // AUTO-DISABLED (NONDET unsound for reference types): {sig}"
                )
```

shellygr: Technically speaking no, but the guardrail isn't wrong.

jar-ben (approved):

PR #38, opened 2026-07-03
Ignore build system's optimizer setting by default

shellygr:
## Summary
- The generated Certora config was blindly carrying a project's own `optimizer`/`optimizer_runs` setting into the conf. That setting tunes the project's *normal* build and has nothing to do with what Certora's own compilation of the same sources needs — inheriting it can break the prover (e.g. a huge `optimizer_runs` value) or is simply unnecessary.
- The ignore now lives in the shared `BuildSystemConfig._apply_common_solc_settings()` helper, so it applies to **every build system** (Foundry, Hardhat, and any future one), not just Foundry. `solc_optimize` is dropped at the base level rather than added-then-popped in `FoundryConfig`.
- `compilation_workarounds.py`'s escalation chain (`yul_exception_add_optimizer`, `stack_too_deep_via_ir`) already re-adds `solc_optimize` if compilation genuinely requires it. So: ignore it by default, let a real compilation failure bring it back via the existing workaround chain.
- Parsed `optimizer`/`optimizer_runs` values are left untouched for other consumers — in particular Foundry's explicit, per-contract `compilation_restrictions` override path (`apply_per_contract_settings` → `FoundryRestrictionManager`), which is a separate code path and still emits `solc_optimize` / `solc_optimize_map`. That mechanism is a deliberate, per-contract compilation signal from the project and is intentionally left as-is.
- `solc` and `solc_via_ir` are unaffected.

## Test plan
- [x] Verified manually against the edited code: a foundry.toml with `optimizer = true` (both boolean and `[profile.default.optimizer] enabled = true` forms) and a directly-constructed `HardhatConfig(optimizer=True)` both no longer produce `solc_optimize` in `to_certora_dict()`'s output, while `solc` / `solc_via_ir` are preserved and the parsed `optimizer` value is still available.
- [x] Full existing `test_foundry_compilation_restrictions.py` suite (31 tests) still passes — the per-contract restriction path is untouched.
- [x] Companion AutoSetup PR: `test_optimizer_default_ignored.py` (6 tests) covers Foundry (boolean + object form, no-optimizer, other-settings-unaffected) and Hardhat (the reviewer's "does not handle hardhat" case).


jar-ben (changes requested):
It does not handle hardhat. I would expect this to be in `base`? Also, why don't we just stop parsing it completely if we don't use it? AFAIK there are already fields we don't parse (because we don't use them).

shellygr:
Addressed. The foundry restrictions thing was surprising - let's keep sampling projects.

---

Claude answers: Good calls — addressed both.

- **Hardhat / `base`**: moved the ignore into the shared `BuildSystemConfig._apply_common_solc_settings()` helper. `solc_optimize` is now dropped at the base level, so it covers Hardhat and any future build system, not just Foundry. Reverted the Foundry-specific add-then-pop. Companion AutoSetup PR now has a Hardhat regression test too.

- **Why not stop parsing it entirely**: we do still use the parsed `optimizer_runs`. Foundry's `compilation_restrictions` path (`FoundryRestrictionManager`) uses the profile-level `optimizer_runs` as the per-contract fallback (`default_optimizer_runs`) when a restriction rule sets `via_ir`/a path but no explicit runs value. So the value is consumed elsewhere — only its emission into the base conf is suppressed. (We deliberately left the `compilation_restrictions` path honoring the optimizer as-is: it's an explicit per-contract compilation signal, unlike the global setting.)

jar-ben (approved):

PR #39, opened 2026-07-03
Stop false 'keccak/unstructured storage is unreachable' skips; gate skip justifications

shellygr:
Part 1 of 6 related spec-quality PRs derived from comparing generated specs against hand-written specs from a recent customer engagement (generated specs skipped properties claiming keccak-derived storage slots are unreachable in CVL — they are not).

## What
- `prop_inference.py` `CERTORA_BACKEND_GUIDANCE`: hash *collision/inversion* reasoning remains out of scope, but keccak-derived storage slots are compile-time constants — hookable via `Sload`/`Sstore` and exposable via harness getters.
- `property_generation_prompt.j2`: new `<storage_access>` block (slot-constant definitions, slot hooks, ghost mirroring, harness getters, KB pointers).
- `property_judge_system_prompt.j2`: "No Source Changes" → "No **Protocol** Source Changes"; harnesses are verification infrastructure. Harness-augmentation wording is injected by `property_feedback_judge(harness_augmentation: bool = False)` into both judge templates through one bind helper (no-op until the main-harness PR flips the argument).
- `property_judge_prompt.j2` Criteria 7: judge must check slot-hook / harness-getter / ghost+summary expressibility before accepting a "not expressible" skip, and audits access-shaped skip reasons whose declared category isn't `storage_access`; greenfield keeps unconditional rejection (author controls the stubs).
- `cvl_generation.py` `_RecordSkipSchema`: required `reason_category: Literal["storage_access", "hash_collision", "prover_limitation", "environment", "other"]` plus required `alternatives_considered`. The gate is purely structural: a `storage_access` skip cannot be recorded until `alternatives_considered` is non-empty — no text matching over LLM prose. Category and alternatives are surfaced verbatim to the judge and covered by the validation digest (mutating them stales the stamp; default `"other"` category and empty alternatives hash like legacy for cache back-compat).

- Regex skip gate (`_SKIP_GATE_CHECKS`/`_KECCAK_SLOT`) deleted — NLP-in-regex over LLM prose was brittle; the author now self-classifies via the `reason_category` enum and the gate checks structure only.
- Classification honesty moved to the judge — Criteria 7 gained one line auditing access-shaped reasons categorized as anything other than `storage_access` (judge polices semantics, not a regex).
- `reason_category` joins `_compute_digest` only when non-default, so cached pre-field skips keep their legacy digests and stamps.
- Mirrored `NotRequired` `harness_augmentation` fields on `FeedbackInherentParams`/`JudgeSystemParams` (aligned only by comments) removed — the flag now threads once through `property_feedback_judge` into both template binds, so the two judge prompts cannot disagree.
- Render tests assert harness wording is present/absent in both judge templates together for flag true and false, including through the injection helper itself.

## Cross-PR interactions
- `harness_augmentation` is flipped to true by the ap-main-harness PR (#40) changing exactly one Python argument at the `property_feedback_judge` call site; safe to merge in either order.
- The `reason_category` + structural gate + digest coverage + judge-audit design is the shared "structured exception ledger" pattern that #41 (vacuity repair) reuses for `acknowledge_vacuous_method`; #41 rebases on this PR's `_compute_digest` and judge-input plumbing.
- Touches `property_generation_prompt.j2`/`property_judge_prompt.j2` in disjoint sections from ap-vacuity-repair and ap-cvlmath; later-merging PR rebases mechanically.

## Tests
`uv run pytest tests/ -q -m "not expensive" --ignore=tests/test_rag_db.py --ignore=tests/test_token_usage.py` → 137 passed, 3 skipped, 3 deselected (ignored files fail identically on master: Postgres testcontainer / model-registry env). `tests/test_cvl_skips.py` rewritten for the enum gate (structural gating, digest back-compat, judge surfacing); `tests/test_template_render.py` gained the cross-template agreement tests. Pyright on changed files: 0 errors.

### DIFF
--- a/composer/spec/cvl_generation.py
+++ b/composer/spec/cvl_generation.py
@@ -7,7 +7,6 @@ Parameterized by:
 """
 
 import hashlib
-import re
 from dataclasses import dataclass
 from typing import Annotated, Callable, Literal, NotRequired, override, Awaitable, Any, Protocol
 from typing_extensions import TypedDict
@@ -47,6 +46,21 @@ CVL_JUDGE_KEY = CacheKey[CVLGeneration, CVLJudge]("judge")
 # Feedback types
 # ---------------------------------------------------------------------------
 
+# The author's self-classification of a skip reason. ``storage_access`` names the one
+# category that is a capability gap CVL can bridge (keccak slot constants, storage hooks,
+# harness getters, ghost mirroring) rather than an expressiveness limit, so it is the only
+# category the record_skip gate treats specially. The judge polices classification honesty
+# (Criteria 7 in property_judge_prompt.j2 audits access-shaped reasons categorized
+# otherwise as storage_access), so the enum is a declaration, not a trusted fact.
+ReasonCategory = Literal[
+    "storage_access",     # the state exists on-chain but is hard to *read* from CVL
+    "hash_collision",     # requires reasoning about collisions/preimages/inversion of a hash
+    "prover_limitation",  # a documented prover limit (e.g. quantifying over unbounded arrays)
+    "environment",        # depends on off-chain/environment behavior CVL cannot model
+    "other",
+]
+
+
 class SkippedProperty(BaseModel):
     """A property the agent explicitly decided not to formalize."""
     property_title: str = Field(description="The unique snake_case title of the property from the batch listing")
@@ -55,6 +69,11 @@ class SkippedProperty(BaseModel):
     alternatives_considered: list[str] = Field(
         default_factory=list,
         description="The alternative mechanisms the author considered before skipping, and why each does not work",
+    )
+    # Defaulted for back-compat with cached instances that predate the field.
+    reason_category: ReasonCategory = Field(
+        default="other",
+        description="The author's self-classification of the skip reason",
     )
 
 
@@ -158,6 +177,12 @@ def _compute_digest(curr_spec: str, skipped: list[SkippedProperty]) -> str:
     digester.update(curr_spec.encode())
     for s in skipped:
         digester.update(f"{s.property_title}:{s.reason}".encode())
+        # The judge audits reason_category as part of the skip's justification
+        # (feedback.py surfaces it), so changing it must stale the validation stamp.
+        # The default "other" contributes nothing, so digests of cached pre-field
+        # skips (which deserialize with the default) keep their legacy value.
+        if s.reason_category != "other":
+            digester.update(f"|cat:{s.reason_category}".encode())
         # The judge audits alternatives_considered as part of the skip's justification
         # (feedback.py surfaces them), so changing them must stale the validation stamp.
         # An empty list contributes nothing, so digests of cached pre-field skips (which
@@ -317,40 +342,14 @@ class _FeedbackSchema(WithInjectedState[CVLGenerationState], WithInjectedId, Wit
             )
         return tool_state_update(self.tool_call_id, msg)
 
-# Skip gate: reasons of the shape "the state is hard to *access*" name a capability gap that
-# CVL can bridge, not an expressiveness limit. Each entry pairs a trigger pattern (matched
-# against the skip reason) with the capability that must be addressed in
-# ``alternatives_considered`` and the evidence pattern used to detect that the capability was
-# actually discussed. A skip whose reason fires a trigger is rejected until every triggered
-# capability is addressed.
-#
-# A bare "keccak" also appears in the one hash-related skip that stays legitimate — reasoning
-# about collisions/inversion/preimages of the hash function itself — so the keccak trigger
-# requires the word near "slot"/"storage", the access-shaped usage this gate targets.
-_KECCAK_SLOT = r"keccak\S*.{0,40}(?:slot|storage)|(?:slot|storage).{0,40}keccak"
-_SKIP_GATE_CHECKS: list[tuple[re.Pattern[str], str, re.Pattern[str]]] = [
-    (
-        re.compile(rf"{_KECCAK_SLOT}|storage slot|unstructured", re.IGNORECASE),
-        "a precomputed keccak storage-slot constant (keccak256 is a built-in CVL function)",
-        re.compile(r"keccak|slot", re.IGNORECASE),
-    ),
-    (
-        re.compile(rf"{_KECCAK_SLOT}|storage slot|unstructured|cannot (be )?(access|observ)", re.IGNORECASE),
-        "Sload/Sstore storage hooks on the slot/path in question",
-        re.compile(r"sload|sstore|hook", re.IGNORECASE),
-    ),
-    (
-        re.compile(r"no (public )?getter|cannot (be )?(access|observ)|unstructured", re.IGNORECASE),
-        "a harness getter/helper or direct storage access (currentContract.<field>)",
-        re.compile(r"harness|getter|helper|currentContract|direct storage", re.IGNORECASE),
-    ),
-    (
-        re.compile(r"cannot (be )?(access|observ)", re.IGNORECASE),
-        "ghost state mirroring the value via hooks or summaries",
-        re.compile(r"ghost", re.IGNORECASE),
-    ),
-]
-
+# Skip gate: the author self-classifies each skip via ``reason_category``. A
+# ``storage_access`` skip ("the state is hard to *read*") names a capability gap that CVL
+# can bridge — precomputed keccak slot constants, Sload/Sstore hooks, harness getters,
+# ghost mirroring — not an expressiveness limit, so it cannot be recorded until
+# ``alternatives_considered`` is non-empty. The gate is purely structural (enum value +
+# non-empty field check); the judge audits the *substance* of the alternatives and
+# polices classification honesty (an access-shaped reason categorized otherwise is
+# audited as storage_access — see Criteria 7 in property_judge_prompt.j2).
 
 @tool_display(
     lambda p: f"Skipping property `{p.get('property_title', '?')}`",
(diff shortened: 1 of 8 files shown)


composer/spec/cvl_generation.py
```diff
@@ -305,14 +317,50 @@ async def run(self) -> Command:
...
+# against the skip reason) with the capability that must be addressed in
+# ``alternatives_considered`` and the evidence pattern used to detect that the capability was
+# actually discussed. A skip whose reason fires a trigger is rejected until every triggered
+# capability is addressed.
+#
+# A bare "keccak" also appears in the one hash-related skip that stays legitimate — reasoning
+# about collisions/inversion/preimages of the hash function itself — so the keccak trigger
+# requires the word near "slot"/"storage", the access-shaped usage this gate targets.
+_KECCAK_SLOT = r"keccak\S*.{0,40}(?:slot|storage)|(?:slot|storage).{0,40}keccak"
+_SKIP_GATE_CHECKS: list[tuple[re.Pattern[str], str, re.Pattern[str]]] = [
```
shellygr: claude, have you THOUGHT LONG AND HARD ABOUT THIS?

shellygr:
Reworked per the architecture review — delta since the previous version (aef046a → b7cf58e):

- The regex skip gate is gone: `_SKIP_GATE_CHECKS`/`_KECCAK_SLOT` deleted from `cvl_generation.py`. The author now declares a required `reason_category` enum (`storage_access` / `hash_collision` / `prover_limitation` / `environment` / `other`) on `_RecordSkipSchema`.
- The gate is purely structural: a `storage_access` skip is rejected until `alternatives_considered` is non-empty — no text matching over LLM prose.
- The judge polices classification honesty: `reason_category` is surfaced in its input, and Criteria 7 now audits access-shaped skip reasons categorized as anything else as if they were `storage_access`.
- Cache back-compat: category defaults to `"other"` and only non-default values join `_compute_digest`, so pre-field cached skips keep their legacy digests and validation stamps.
- The mirrored `NotRequired` `harness_augmentation` fields in `feedback.py` are removed; the flag threads once through `property_feedback_judge(harness_augmentation: bool = False)` into both judge-template binds, with render tests asserting the two templates agree for both flag values.
- `tests/test_cvl_skips.py` rewritten for the enum gate; suite: 137 passed, 3 skipped, 3 deselected.

composer/spec/prop_inference.py
```diff
@@ -88,8 +88,15 @@ class RefinementState(MessagesState):
 
 1. Attack vectors or invariants that reference off-chain events (like
    key compromising, phishing, etc.)
-2. Reasoning about hash function behavior or hash collisions (e.g.,
-   "invalid signatures should be rejected")
+2. Reasoning about hash function *collisions or inversion* (e.g.,
```
shellygr, line 91: We need to see how this updated section interacts with proper erc-7201 annotations. It may apply in cases where proper annotations do not exist or that we failed to parse them.

composer/templates/property_judge_prompt.j2
```diff
@@ -168,14 +168,57 @@ Verify that the specification covers all {{ properties | length }} listed proper
...
+
   *Exception*: if the author has filed an empirical rebuttal (see Step 1) against a prior-round
   formalization you proposed, accept that rebuttal per the Step 1 evaluation rules — do not
   re-insist on the same proposal. If the rebuttal establishes that no working formalization was
   found, the skip is VALID on the grounds that your prior proposal proved unworkable.
 {% if sort != "greenfield" %}
-  If the property being skipped could be formalized with Solidity code changes, the skip is VALID. Do *NOT* reject
-  skips with advice to change the Solidity code. The Solidity code is *IMMUTABLE* and suggestions to change the Solidity
-  code are INVALID.
+  If the property being skipped could be formalized only with *protocol* Solidity code changes, the skip is VALID.
```
shellygr, line 212: to adapt to John's edit agent

composer/templates/property_judge_system_prompt.j2
```diff
@@ -20,12 +20,29 @@
 {% endwith %}
 
 {% if sort != "greenfield" %}
-## No Source Changes
+## No Protocol Source Changes
```
shellygr, line 23: n.b. protocol source changes means we do not change the original files, but source changes involving harnesses are allowed

composer/spec/cvl_generation.py
```diff
@@ -322,6 +369,28 @@ class _RecordSkipSchema(WithInjectedState[CVLGenerationState], WithInjectedId, W
     reason: str = Field(
         description="Justification for why this property cannot be formalized"
     )
+    reason_category: ReasonCategory = Field(
+        description=(
+            "Classify the skip reason: `storage_access` — the state exists on-chain but is hard "
+            "to READ from CVL (private/internal fields, unstructured or keccak-addressed storage, "
+            "no public getter); `hash_collision` — the property requires reasoning about "
```
jtoman, line 376: again, private/internal fields are eminently readable from cvl via direct storage access, and this tool text will land in the attention window of the agent before it ever sees the CVL research articles about this being possible. This will make the problem worse, not better, the agent will start out thinking that private/internal storage fields can't be read and this is a problem to be overcome.

composer/spec/cvl_generation.py
```diff
@@ -322,6 +369,28 @@ class _RecordSkipSchema(WithInjectedState[CVLGenerationState], WithInjectedId, W
...
+            "access-shaped reason categorized as anything but `storage_access` will be re-audited "
+            "as `storage_access`."
+        )
+    )
+    alternatives_considered: list[str] = Field(
+        description=(
+            "The alternative CVL/verification mechanisms you considered before skipping, and why "
+            "each does not work for this property. Required (non-empty) when `reason_category` is "
+            "`storage_access`, where you MUST address the applicable mechanisms among: precomputed "
+            "keccak storage-slot constants (keccak256 is a built-in CVL function), Sload/Sstore "
```
jtoman, line 390: this is true (`keccak256` is built in), but doing `keccak256("whatever.namespace")` in cvl won't yield the same value as the (likely precomputed) constant generated by the solidity compiler. Further, there's no way to do `storage[keccak256("whatever")]`, the only way `keccak256` would help you here is if you wrote a *raw* storage acccess hook and did conditional branching on `slot == keccak256("...")`. But, again, this doesn't work, because the value computed by `keccak256` doesn't actually match whatever concrete value is being computed by the solidity compiler.

jtoman: the solution is to allow munging/editing around unstructured storage so our existing (and quite robust!) support for eip-7201 makes this problem go away.

For reference, on some project (I forget which), once I got the eip-7201 annotations in the right place (in autosetup) the cvl author was able to use the names we generated just fine.

So: I agree we should reject skips around unstructured storage, but the fix mechanism does not (and cannot) exist spec side.

composer/templates/property_judge_prompt.j2
```diff
@@ -168,14 +168,57 @@ Verify that the specification covers all {{ properties | length }} listed proper
...
+  Before accepting ANY skip justified as "not expressible in CVL" — in particular one claiming that state is
+  private, unstructured, keccak-addressed, or lacks a getter — you must verify that NONE of the following
+  mechanisms make the property expressible:
+  (a) a precomputed keccak storage-slot constant (`keccak256` is a built-in CVL function; ERC-7201 namespaces
+      and EIP-1967 slots are compile-time constants) combined with an `Sload`/`Sstore` hook on that slot;
+  (b) a harness getter or direct storage access (e.g., `currentContract.<field>`);
+  (c) ghost state mirroring the value via storage hooks or function summaries.
+
+  The author self-classifies each skip with a reason category (shown alongside the skip): if the
+  reason is access-shaped — state described as private, unstructured, keccak-addressed, or lacking
```
jtoman, line 180: I will note the "private" restriction is, again, totally invented by claude here, and trust all future such "limitations" are cleaned up rather than continuing to point them out

composer/templates/property_judge_prompt.j2
```diff
@@ -168,14 +168,57 @@ Verify that the specification covers all {{ properties | length }} listed proper
...
+  If the property being skipped could be formalized only with *protocol* Solidity code changes, the skip is VALID.
+  Do *NOT* reject skips with advice to change the protocol Solidity code. The *protocol* Solidity code is
+  *IMMUTABLE* and suggestions to change the protocol code are INVALID. Verification *harnesses* (contracts under
+  `certora/harnesses/`) are NOT protocol code:
+  {% if harness_augmentation | default(false) %}
+  needing a harness getter or wrapper (mechanism (b) above) does NOT make a skip valid under this rule.
+  {% else %}
+  a skip needing only a harness getter or wrapper is NOT validated by this rule — it falls under the
+  "missing harness support" labeling above.
+  {% endif %}
```
jtoman, line 221: this advice will collide with my editor agent advice soon

composer/templates/property_generation_prompt.j2
```diff
@@ -107,6 +107,27 @@ within the generated summary, match the callee contract against known implemente
...
 inputs, fall back on a sound model of the external behavior, using e.g., ghosts.
 
+<storage_access>
+Never conclude that contract state is "unreachable" (and never skip a property on that basis) merely because
+the state lives in keccak-derived storage slots (ERC-7201 namespaced storage, EIP-1967 proxy slots,
+`keccak256(name) - 1` patterns) or lacks a public getter. `keccak256` is a built-in CVL function, and these
+slots are compile-time constants. The available mechanisms are:
+
+* **Slot-constant `definition`s**: compute the concrete 32-byte slot value *once* (from the constant declared
+  in the source, or by evaluating the keccak expression yourself) and bind it with a CVL `definition`, e.g.,
```
jtoman, line 117: > evaluating the keccak expression yourself

We think claude can reliably simulate the sha3 algorithm???

shellygr: Not sure about the API we use but the chatbots get it right. But those are probably highly engineered behind the scene to write sandboxed code and execute it

PR #40, opened 2026-07-03
Main-contract augmentation harness: unstructured-storage getters, helper decomposition, verify-target threading

shellygr:
SG: this one is pretty risky as it introduces harnessing prominently. To be discussed.

---

Part 2 of 6 (spec-quality series). In a recent engagement, the hand-written specs verified a harness contract exposing internal state and splitting a monolithic oracle-report function into helper entry points; the pipeline had no equivalent, so whole components were skipped or given up.

## What
- New harness product: `certora/harnesses/<Main>Harness.sol` (`contract <Main>Harness is <Main>`) with (a) external view getters for private/unstructured storage (assembly `sload` of precomputed slot constants), (b) thin external wrappers decomposing monolithic functions.
- `harness.py`: `UnstructuredSlotSpec` / `HelperDecomposition` / `MainHarnessPlan` models; `AgentSystemDescription.main_contract_harness` and `SystemDescriptionHarnessed.main_harness` (optional-with-default → cached JSON still validates, covered by back-compat tests); `generate_main_harness()` with VFS confinement + delivery-path validation.
- `state_analysis.j2`: classifier flags hidden state (assembly/StorageSlot/ERC-7201 without getters) and monolithic externals → proposes a plan; each slot derivation must cite the source line (constant declaration / namespace string) it was computed from.
- Verify-target threading: `verify_contract_name()` accessor computed once in `pipeline.py` — after AutoSetup, so a fallback (below) is already reflected — and threaded to autosetup, prover tool, and conf writing. When the harness survives AutoSetup it becomes the verified contract.
- Raw-contract fallback: if the generated harness fails AutoSetup/compilation, the run logs the failure, retries setup against the raw main contract, clears the harness from the system description, and records the downgrade in `main_harness_fallback` — a broken harness degrades the run instead of killing it.
- Soundness guardrails (harness helpers expose transitions unreachable in production): author + shared-context guidance to exclude helpers from parametric rules/invariants (`filtered` idiom) and never witness `satisfy` through a helper; judge enforces a filter-legitimacy taxonomy — a `filtered` block is legitimate iff every excluded method is a documented harness helper or covered by a recorded vacuity acknowledgment.
- Harness API surfaced to author/judge/invariant prompts (`harness_api`); the `harness_augmentation` flag threads once through `property_feedback_judge()` into BOTH judge template binds (system prompt and Criteria 7 task prompt), keeping them in agreement by construction.
- Report disclosure: `HarnessDisclosure` (harness name/path, contract it extends, API one-liners) plus any `main_harness_fallback` reason land in the report schema and HTML, so consumers can reconcile the conf's verify target with the report's contract.
- `GiveUpTool` gains `sort` (`env_issue|needs_harness|prover_limit|other`) carried into reports, making remaining harness gaps measurable.

- Always-on confirmed (no operator off-switch); the review's soundness concerns are addressed pre-merge instead, via the three items below.
- Soundness guardrails added to author, judge, and shared harness context: helper-exclusion discipline for parametric rules/`satisfy`, plus the judge's filter-legitimacy taxonomy (documented helper or acknowledged-vacuous ledger entry — coherent with #41's judge wording).
- Raw-contract fallback replaces hard failure when the harness breaks AutoSetup; verify target is now computed post-AutoSetup so the downgrade propagates everywhere.
- Harness use is disclosed in the report (getters/helpers listed; fallback reason shown) rather than silent — required since the feature is always-on.
- `harness_augmentation` now injected into both judge binds from the single `property_feedback_judge` parameter (mirrored `NotRequired` fields removed); a render test asserts system and task prompts agree for both flag values.
- Cleanup per review: dead `ProverArtifactStore` main-contract field dropped; slot derivations must be source-cited in `state_analysis.j2`.

## Cross-PR interactions
- #39 (`ap-keccak-skip-gate`): consumes its `harness_augmentation | default(false)` template gate; this PR completes the both-binds threading of the flag (the `default(false)` is a render fail-safe, not the mechanism). Either merge order renders correctly; planned order is #39 → #40 → #41.
- #41 (`ap-vacuity-repair`): the judge's filter-legitimacy taxonomy references #41's acknowledged-vacuous ledger entries as the second legitimate filter reason; wording is written to compose with #41's judge edits.

## Tests
`uv run pytest tests/ -q -m "not expensive" --ignore=tests/test_rag_db.py --ignore=tests/test_token_usage.py` → 132 passed, 3 skipped (ignored = pre-existing env failures on master). Includes pydantic back-compat, accessor, template render, classifier empty-plan-rejection, guardrail-presence, judge-bind render-agreement (both flag values), fallback-path, and report-disclosure tests. Test fixtures are fully typed (`SolidityIdentifier` constructor idiom, real `PropertyFormulation` instead of duck-typed stubs) so pyright is clean on tests too.

### DIFF
--- a/composer/spec/feedback.py
+++ b/composer/spec/feedback.py
@@ -49,9 +49,6 @@ FeedbackTemplate = TypedTemplate[FeedbackInherentParams]("property_judge_prompt.
 
 class JudgeSystemParams(TypedDict):
     sort: Sort
-    # True when the run verifies through a main-contract augmentation harness;
-    # gates the harness-aware wording (the template defaults it to false).
-    harness_augmentation: NotRequired[bool]
 
 # Judge system prompt, shared between the natspec and source-mode flows. The fs
 # primitives are always documented; ``sort`` drives the rest (the template
@@ -59,22 +56,37 @@ class JudgeSystemParams(TypedDict):
 # ``sort == "existing"``, the only mode that wires those tools).
 FeedbackSystemTemplate = TypedTemplate[JudgeSystemParams]("property_judge_system_prompt.j2")
 
+
+def _bind_harness_augmentation(
+    prompt: TemplateInstantiation, harness_augmentation: bool
+) -> TemplateInstantiation:
+    """Inject the ``harness_augmentation`` flag into an already-instantiated template bind.
+
+    Both judge templates gate their harness-skip policy on this variable; injecting it
+    from the single ``property_feedback_judge`` parameter into BOTH binds is what keeps
+    the system prompt and Criteria 7 in agreement. The templates' ``default(false)`` is
+    a render fail-safe only, never the mechanism that decides the value.
+    """
+    return TemplateInstantiation(
+        prompt.template,
+        {**prompt.args, "harness_augmentation": harness_augmentation},
+    )
+
+
 def property_feedback_judge(
     ctx: WorkflowContext[CVLJudge],
     env: ServiceHost,
     prompt: InjectedTemplate[Properties] | TemplateInstantiation,
     props: list[PropertyFormulation],
     *,
+    harness_augmentation: bool = False,
     extra_inputs: list[str | dict] | Callable[[], list[str | dict]] | None = None,
     system_prompt: TemplateInstantiation | None = None,
-    harness_augmentation: bool = False,
 ) -> FeedbackToolContext:
 
     if system_prompt is None:
-        system_prompt = FeedbackSystemTemplate.bind({
-            "sort": env.sort,
-            "harness_augmentation": harness_augmentation,
-        })
+        system_prompt = FeedbackSystemTemplate.bind({"sort": env.sort})
+    system_prompt = _bind_harness_augmentation(system_prompt, harness_augmentation)
 
     builder = env.builder_heavy().with_tools(
         env.all_tools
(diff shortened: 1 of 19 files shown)


shellygr:
Delta since the pre-review version, for reviewers who saw the old branch (5 new commits, e9e2d5a..5735392):

- Soundness guardrails: author/judge/shared-context prompts now forbid helper-witnessed `satisfy` and helper-driven property weakening, and require helpers be excluded from parametric rules via `filtered`; the judge gets a filter-legitimacy taxonomy (a filter is legitimate iff it excludes documented harness helpers or acknowledged-vacuous methods).
- Raw-contract fallback: a generated harness that fails AutoSetup no longer kills the run — setup retries against the raw main contract, the harness is cleared, and the downgrade is recorded (`main_harness_fallback`). Verify target is computed post-AutoSetup so the fallback propagates to prover tool and confs.
- Report disclosure: reports now state that verification ran against `<Main>Harness`, list its getters/wrappers, and show any fallback reason (new `HarnessDisclosure` schema + HTML section).
- `harness_augmentation` now threads from the single `property_feedback_judge()` parameter into BOTH judge template binds (system + task prompt); mirrored `NotRequired` fields removed; render test asserts the two templates agree for both flag values.
- Cleanup: dead `ProverArtifactStore` main-contract field dropped; `state_analysis.j2` now requires each slot derivation to cite its source line.
- Tests: 120 → 132 passed (fallback path, disclosure rendering, guardrail presence, bind agreement).

shellygr:
Addendum to the delta summary above: one more commit pushed since (595da57).

- Test-only: typed the main-harness test fixtures so pyright is clean on tests as well (`SolidityIdentifier` constructor idiom for fixture literals; judge-bind injection test now uses a real `PropertyFormulation` instead of a duck-typed stub).
- No production-code changes; suite unchanged at 132 passed, 3 skipped.

PR #41, opened 2026-07-03
Vacuity detection + repair ladder: fix setup root causes instead of filtering methods

shellygr:
Part 3 of 6 (spec-quality series). In a recent engagement, a NONDET summary on a payable external dependency made a method revert on every path, and the generated specs filtered it out of nearly every rule instead of repairing the setup.

## What
- New `composer/prover/vacuity.py`: aggregates per-rule `SANITY_FAILED` results by instantiated method; flags methods vacuous in 100% of the rules that instantiate them, OR in ≥2 rules with **no healthy instantiation anywhere** (a passing sanity instance anywhere clears the method — legitimate mostly-reverting methods are not flagged); emits a `<vacuity_alert>` with the repair ladder (fix summary → mock → `optimistic_fallback` → only then `filtered`).
- Alert wired into the verify tool report; `verify_spec` withholds the PROVER validation stamp when a vacuous method sits in a `filtered` block — last resort is the structural `acknowledge_vacuous_method` ledger tool (`method`, `steps_attempted: list[summary_fix|mock|optimistic_fallback]`, `justification`; steps must be non-empty). No comment scanning: a filtered method that is known-vacuous and unacknowledged blocks the stamp, and a stale verdict only clears when a later run shows the method healthy.
- Agent-mutable setup is digest-covered: agent-set config, mocks, and acknowledgments join the PROVER-stamp validation digest (empty-contributes-nothing, so pre-field digests stay valid), so any post-stamp edit forces re-verification.
- Author tools: previously-dead `ConfigEditTool` registered as `edit_config`; new `SetProverFlag` (whitelist: `optimistic_fallback` add-only — may enable, never unset an autosetup recommendation; `loop_iter` 1–5; `global_timeout` ≤ 2× the autosetup default; `contract_recursion_limit`; `rule_sanity`/`optimistic_loop` deliberately excluded — detection depends on forced `rule_sanity: basic`). New `write_mock` tool confined to `certora/mocks`: output is held in graph state (namespaced `certora/mocks/<spec stem>/`), materialized onto the project tree only for the duration of a prover run, cleaned up afterward, and persisted via `ProverArtifactStore` so cache replay works from a fresh checkout.
- Judge evidence: vacuity verdicts, rule skips, and ledger acknowledgments reach the judge verbatim via an `extra_inputs` thunk; judge audits each acknowledgment on its merits. Criteria 5 rejects filters that paper over repairable summaries (NONDET-on-payable called out).
- Prompts: HAVOC/SANITY_FAILURE/filtered remediation rewritten as the ordered ladder.
- Autosetup-side prevention (NONDET guard + `optimistic_fallback` recommendation) is no longer in this PR: split out to #52 at the owner's request, reworked there with AST-based detection.

- Lexical escape hatch (comment scanning near `filtered` blocks) replaced by the structural `acknowledge_vacuous_method` ledger — shared "declare via typed tool schema, gate structurally, judge audits semantics" pattern with #39's `reason_category`.
- `write_mock` no longer mutates the shared project tree: writes were racing under concurrent batch generations; mocks now live in per-generation graph state, materialize per prover run, and persist through the artifact store.
- Agent-set config, mocks, and acknowledgments added to the PROVER-stamp digest — post-stamp edits can no longer ride a stale stamp.
- Vacuity evidence, rule skips, and acknowledgments now actually reach the judge (`extra_inputs`) instead of the judge seeing only the spec text.
- The ≥2-rules detection arm tightened to require no healthy instantiation — removes definitional false positives on methods that legitimately revert in most prestates (e.g. post-init initializers).
- `SetProverFlag` clamps tightened: `global_timeout` ≤ 2× autosetup default (was 7200), `loop_iter` ≤ 5 (was 8), `optimistic_fallback` add-only.

## Cross-PR interactions
- Touches `property_generation_prompt.j2`/`property_judge_prompt.j2` in disjoint sections from ap-keccak-skip-gate (#39) and ap-cvlmath (#42 — whose assets are now `Math.spec`/`CVLMathAbstract.spec`).
- Shares the structured exception-ledger pattern with #39 and reuses its digest and judge-`extra_inputs` plumbing; merge order #39 → #40 → #41 (the second lander of the digest change adds length-prefixing).

## Tests
Targeted: 88 passed (vacuity, config-edit, write-mock, skip suites). Full sweep: 175 passed, 3 skipped, with the two pre-existing env-failure files excluded (fail identically on master).

### DIFF
--- a/certora_autosetup/autosetup/autosetup.py
+++ b/certora_autosetup/autosetup/autosetup.py
@@ -903,6 +903,20 @@ class Autosetup:
 
             self.log(f"🔗 Running call resolution for {contract_handle.contract_name}")
             await call_resolution_phase.execute(max_iterations=10)
+
+            # Call resolution may recommend conf flags (e.g. optimistic_fallback when an
+            # unresolved low-level value transfer remains — no link/dispatcher can resolve
+            # those, and without the flag every caller can vacuously revert). The base conf
+            # was created before this phase ran, so merge into the existing conf here rather
+            # than at create_config time.
+            if call_resolution_phase.recommended_extra_flags:
+                self.log(
+                    f"Applying conf flags recommended by call resolution: "
+                    f"{call_resolution_phase.recommended_extra_flags}"
+                )
+                self.config_manager.apply_extra_flags(
+                    enhanced_config.path, call_resolution_phase.recommended_extra_flags
+                )
             return True
 
         except Exception as e:
--- a/certora_autosetup/autosetup/autosetup.py
+++ b/certora_autosetup/autosetup/autosetup.py
@@ -215,10 +215,6 @@ class Autosetup:
         llm = set(setup._methods_per_contract.keys()) & self._library_names
         return curated | llm
 
-    def _build_job_msg(self, contract_name: str, conf_file: Path) -> str:
-        """Build the msg string for a prover job."""
-        return ProverJobSpec.build_job_msg(self.config.orchestration_timestamp, contract_name, conf_file)
-
     def run(self, main_contract_handle: ContractHandle, skip_warmup: bool = False) -> AutosetupResult:
         """Execute the full autosetup pipeline.
 
@@ -811,7 +807,6 @@ class Autosetup:
                     contract_name=contract_name,
                     phase=f"Sanity Test Run - {contract_name}",
                     extra_args=autosetup.config.extra_args,
-                    msg=autosetup._build_job_msg(contract_name, test_config.path),
                 ))
 
                 if skip_warmup:
@@ -831,7 +826,6 @@ class Autosetup:
                     contract_name=contract_name,
                     phase="Sanity Warmup - warmup",
                     extra_args=autosetup.config.extra_args,
-                    msg=autosetup._build_job_msg(contract_name, warmup_config.path),
                 )
 
             async def prepare_and_run_warmup():
@@ -903,20 +897,6 @@ class Autosetup:
 
             self.log(f"🔗 Running call resolution for {contract_handle.contract_name}")
             await call_resolution_phase.execute(max_iterations=10)
-
-            # Call resolution may recommend conf flags (e.g. optimistic_fallback when an
-            # unresolved low-level value transfer remains — no link/dispatcher can resolve
-            # those, and without the flag every caller can vacuously revert). The base conf
-            # was created before this phase ran, so merge into the existing conf here rather
-            # than at create_config time.
-            if call_resolution_phase.recommended_extra_flags:
-                self.log(
-                    f"Applying conf flags recommended by call resolution: "
-                    f"{call_resolution_phase.recommended_extra_flags}"
-                )
-                self.config_manager.apply_extra_flags(
-                    enhanced_config.path, call_resolution_phase.recommended_extra_flags
-                )
             return True
 
         except Exception as e:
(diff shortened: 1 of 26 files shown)


shellygr:
Branch reworked per the architecture review — delta since the previous version (6 new commits, no history rewrite):

- Comment-scanning escape hatch is gone: filtered vacuous methods now require a structural `acknowledge_vacuous_method` ledger entry (typed steps-attempted + justification), same declare/gate/judge-audits pattern as #39's `reason_category`.
- `write_mock` no longer writes into the shared project tree (raced under concurrent batch generations): mocks live in per-generation graph state under `certora/mocks/<spec stem>/`, materialize only for the duration of a prover run, and persist via `ProverArtifactStore` for cache replay.
- Agent-set config, mocks, and acknowledgments now join the PROVER-stamp validation digest, so post-stamp edits force re-verification.
- Vacuity verdicts, rule skips, and acknowledgments now reach the judge via `extra_inputs` (previously the judge never saw them).
- Detection tightened: the ≥2-rules arm also requires no healthy instantiation anywhere, eliminating false positives on legitimately mostly-reverting methods.
- `SetProverFlag` clamps: `global_timeout` ≤ 2× autosetup default, `loop_iter` ≤ 5, `optimistic_fallback` add-only (can enable, never unset an autosetup recommendation).

certora_autosetup/setup/call_resolution.py
```diff
@@ -125,6 +125,25 @@ def dispatchers_added(self) -> int:
...
+    Such a call HAVOCs and — crucially — can always be *assumed to fail*, which makes every
+    caller able to revert on all paths (the vacuous-rule failure mode). ``optimistic_fallback``
+    assumes these calls succeed instead, hence the recommendation built from this predicate.
+
+    The selector check keeps high-level calls that merely look alike out: an unresolved
+    ERC20-style ``token.transfer(...)`` carries a resolvable selector, while native value
+    transfers have none (``[?].[?]`` in the report legend).
+    """
+    snippet = (call.call_site_snippet or "").replace(" ", "")
+    if not any(marker in snippet for marker in (".call{value", ".send(", ".transfer(")):
```
shellygr: oh no. another case that something HAS NOT BEEN THOUGHT LONG AND HARD ABOUT.

certora_autosetup/autosetup/autosetup.py
```diff
@@ -903,6 +903,20 @@ async def run_single_contract_call_resolution(
 
             self.log(f"🔗 Running call resolution for {contract_handle.contract_name}")
             await call_resolution_phase.execute(max_iterations=10)
+
+            # Call resolution may recommend conf flags (e.g. optimistic_fallback when an
+            # unresolved low-level value transfer remains — no link/dispatcher can resolve
+            # those, and without the flag every caller can vacuously revert). The base conf
+            # was created before this phase ran, so merge into the existing conf here rather
+            # than at create_config time.
+            if call_resolution_phase.recommended_extra_flags:
```
shellygr: I am surprised we do not include optimistic_fallback already.

shellygr:
Claude answers: heads-up on a scope split — the autosetup-side vacuity prevention (payable-NONDET guard + `optimistic_fallback` recommendation) moved out of this PR into #52, per the owner's request, and was reworked there with AST-based detection. This PR keeps all the composer-side work. Note: `tests/test_autosetup_vacuity_prevention.py` on this branch is stale (its replacement lives in #52) and is pending a `git rm` that requires owner action.

PR #42, opened 2026-07-03
Ship CVLMath relational math library; prefer relational properties over formula mirroring

shellygr:
Part 4 of 6 (spec-quality series). Human specs tame nonlinear arithmetic with mulDiv abstractions (CVLMath.spec / mulDivDownAbstractPlus); generated specs instead mirrored exact formulas behind piles of overflow requires — timeout-prone and tautological.

## What
- Two-tier math library shipped into every generated project, with a single source per tier:
  - **Exact tier — `Math.spec`**: certora_autosetup's bundled Math.spec is the one canonical source. The install helper copies it to `certora/specs/summaries/Math.spec` (and advertises it as a resource) only when AutoSetup's summary closure did not already place it there — one path, no drift, and a cached generated spec importing `summaries/Math.spec` keeps typechecking across runs (same-path diamond imports dedupe).
  - **Relational tier — `composer/spec/assets/CVLMathAbstract.spec`**: always installed and advertised as a `required=False` `CVLResource`. `mulDivDownAbstract`/`mulDivUpAbstract` with zero-preservation, exactness, monotonicity axioms; `down <= up <= down+1`; WAD/RAY definitions; vacuity-warning headers matching the existing `sqrtSummaryPrecise` style.
- The existing resources block renders import guidance with no template change.
- `property_generation_prompt.j2` `<math_properties>`: prefer relational properties (monotonicity, bounds, round-trip inequalities, additivity-up-to-rounding, zero-preservation); use the exact `...Summary` tier by default and switch to the `...Abstract` tier only on timeout. The block spells out the `require`-axiom vacuity caveat: no overflow-revert modeling, avoid under `satisfy`, and a fresh SANITY_FAILURE after adopting them means the axioms conflict with the rule's assumptions.
- Judge Criteria 1: relational arithmetic properties are not tautologies.

## Cross-PR interactions
- Disjoint prompt sections from #39 (ap-keccak-skip-gate) / #41 (ap-vacuity-repair); those PRs' structured exception-ledger rework does not touch this PR's files.
- Asset names changed in this rework to `Math.spec` / `CVLMathAbstract.spec` (there is no shipped `CVLMath.spec` anymore). #44 (ap-kb-articles) article 8 is being rewritten against this reality — merge this PR before #44.

## Tests
`tests/test_cvlmath_asset.py`: 7 passed, including emv.jar Typechecker syntax checks on both shipped spec names. The drift guard now hard-fails (not skips) if certora_autosetup's bundled Math.spec goes missing or is renamed — exactly the event that would break the install helper. Full sweep: 110 passed, 3 skipped with the two pre-existing env-failure files excluded.

- Single-sourced the exact tier from AutoSetup's bundled Math.spec — the composer-shipped `CVLMath.spec` copy duplicated proven summaries and could drift.
- `CVLMathAbstract.spec` is now always installed/advertised — the presence-conditional install and stale-copy unlink apparatus had nothing left to guard once there is one path per tier, so it was deleted.
- Drift-guard test hard-fails instead of skipping when the bundled Math.spec is absent — a skip would hide the exact breakage the guard exists to catch.
- The require-axiom vacuity caveat moved into the `<math_properties>` prompt block, not only spec headers — the generating agent must see it where it chooses the tier.


shellygr:
Reworked per the architecture review — delta for reviewers who saw the earlier version:

- The composer-shipped `CVLMath.spec` is gone. The exact `...Summary` tier is now single-sourced from certora_autosetup's bundled Math.spec, copied to `certora/specs/summaries/Math.spec` only when AutoSetup's summary closure didn't already place it there.
- The relational tier lives in `CVLMathAbstract.spec` and is now **always** installed and advertised (previously presence-conditional).
- The conditional-install and stale-copy unlink apparatus is deleted — with one path per tier there is nothing to drift or remove.
- The drift-guard test hard-fails (previously skipped) if the bundled Math.spec goes missing or is renamed.
- `<math_properties>` now names both tiers and spells out the `require`-axiom vacuity caveat in the prompt itself (no overflow modeling, avoid under `satisfy`, post-adoption SANITY_FAILURE = conflicting axioms) — previously only in spec headers.
- Note for #44: shipped asset names are now `Math.spec` / `CVLMathAbstract.spec`; merge this PR before #44.

composer/spec/assets/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+#: the bundled layout), which is what makes the single-name scheme work: its
+#: presence signals "the exact tier is already installed".
+MATH_SPEC_PROJECT_PATH = SUMMARIES_DIR / MATH_SPEC_NAME
+CVLMATH_ABSTRACT_PROJECT_PATH = SUMMARIES_DIR / CVLMATH_ABSTRACT_SPEC_NAME
+
+
+def cvlmath_abstract_spec_text() -> str:
+    """The packaged standalone relational-abstraction library source."""
+    return (files(__package__) / CVLMATH_ABSTRACT_SPEC_NAME).read_text()
+
```
jtoman, line 44: there's a way to do this with importlib; do that imo

composer/spec/assets/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+    """The canonical exact-tier ``Math.spec`` source, read from the
+    certora_autosetup package (the single source of the ``*Summary``
+    functions; composer deliberately ships no copy of its own).
+
+    Raises if the bundled file is missing/renamed — loud failure is wanted:
+    the install helper and AutoSetup's summary closure must agree on this
+    file's location for the single-name scheme to hold.
+    """
+    return (
+        files("certora_autosetup")
```
jtoman, line 56: wait, you know about the importlib thing, you're doing it here! do it for the bundled math imo

composer/templates/property_generation_prompt.j2
```diff
@@ -107,6 +107,38 @@ within the generated summary, match the callee contract against known implemente
...
+A rule that just re-computes the implementation's formula in CVL only proves the code matches its own
+source text; the relational forms capture the *intent* (no value creation, rounding favors the protocol)
+and are far easier on the solver.
+
+If the prover *times out* on a property because of exact nonlinear formulas, do NOT weaken or skip the
+property. Instead, summarize the underlying math function (e.g., `mulDiv`) with the relational
+abstractions from the CVLMathAbstract.spec resource (see the resources list, when available): the
+`...Abstract` variants replace the nonlinear formula with solver-friendly axioms (zero-preservation,
+exactness at the denominator, monotonicity, down/up rounding within 1) that suffice to prove relational
+properties. The exact `...Summary` models (faithful rounding and revert behavior) live in the Math.spec
```
jtoman, line 131: does this resource exist? I'm missing where that gets plumbed through (or communicated to the agent)

composer/templates/property_judge_prompt.j2
```diff
@@ -67,6 +67,10 @@ For example, `x + 1 > x` is always true, and doesn't say anything *meaningful* a
 is *NOT* tautological even if the implementation/design of `calculateInterest` means it always returns a non-zero value. In this case, the assertion provides a valuable
 "check" on the implementation's behavior.
 
+CLARIFICATION: *Relational* properties about arithmetic — monotonicity, bounds, round-trip inequalities
```
jtoman, line 70: did you see this failure mode? surprising...

jtoman (commented):
drive by comments

PR #43, opened 2026-07-03
Property portfolio: requireInvariant discipline, 6-lens coverage checklist, multi-call techniques

shellygr:
Some more guidance on property inference.

---

Part 5 of 6 (spec-quality series). Generated specs were ~80% unit-style rules; human specs concentrate on accounting invariants, multi-call sequences, and state-transition properties, and never assume state with bare requires.

## What
- `cvl_guidelines.j2` (read by author AND judge): new guideline 24 — a bare `require` over contract state is a spec smell; default is `invariant` + `requireInvariant`, bare requires only with a justification comment. Two carve-outs: solver-capacity bounds (e.g. `require totalSupply() <= 2^128`) are a recognized documented-trust-assumption class needing no invariant attempt, and `preserved` blocks defer to the stricter judge standard (`requireInvariant` or verifiable outside evidence — a comment alone is insufficient). Guideline 23 amended with a ghost-counter carve-out (information never stored in contract state); guideline 20 now names persistent counter ghosts and expression summaries on external-dependency calls as legitimate for that same class.
- `property_analysis_prompt.j2`: 6-lens coverage checklist (unit / valid-state / variable-transition / state-transition / multi-call / risk) — explicitly a brainstorming discipline, not a quota; swept lenses recorded in reasoning. No `PropertyType` schema change (caches survive).
- `property_generation_prompt.j2`: the `property_rules` mapping instruction now requires supporting invariants and companion witness rules to be listed under the property they support — rules unreferenced by any property are silently dropped from the final report.
- `cvl_additions.j2` (author-only): three `<advice>` entries — `storage s = lastStorage` / `f(e,args) at s` snapshot idiom (additivity/round-trip), `satisfy`-witness companion rules for implication antecedents (mapped in `property_rules` under the parent property), ghost counters via expression summaries (with persistent-ghost note and the caveats that summaries replace the callee and never apply to direct CVL calls).

- Guideline 24 carve-out for capacity/overflow-avoidance bounds — they keep the solver tractable rather than model behavior, so no invariant attempt is expected.
- Guideline 24 `preserved`-block policy now defers to the stricter judge standard (Criteria 4), removing the internal conflict where a comment sufficed for the author but not the judge.
- Companion `satisfy`-witness rules and supporting invariants must be mapped in `property_rules` under their parent property — orphan rules are dropped from the report (confirmed by reviewers).
- Ghost-counter/persistent-ghost defense made judge-visible via guideline 20, so the judge shares the vocabulary the author advice uses.

### DIFF
--- a/composer/templates/cvl_additions.j2
+++ b/composer/templates/cvl_additions.j2
@@ -25,4 +25,22 @@ where necessary.
 <advice>
 Reverting behavior is reasoned about by using the `@withrevert` annotation and examining the value of the `lastReverted` flag.
 </advice>
+<advice>
+Multi-call properties can compare executions from a common starting state using storage snapshots: `storage init = lastStorage;`
+snapshots the current blockchain state (and non-persistent ghosts), and `f(e, args) at init;` runs `f` starting from that snapshot
+instead of the current state. The standard additivity idiom: run `g(x); g(y);` sequentially from a snapshot, then run `g(x + y) at init;`
+from the *same* snapshot, and compare the resulting end states (e.g., final balances or shares).
+</advice>
+<advice>
+For implication-shaped assertions (`assert A => B;`), consider a companion witness rule containing `satisfy A;`. A `satisfy`
+statement checks that at least one execution reaches it with the condition true — a single witness execution suffices. This
+guards against vacuously-true implications whose antecedent `A` can never actually hold.
+</advice>
+<advice>
+Information that is never stored in contract state — such as how many times a function was called, or the cumulative amount
+transferred across calls — can be tracked with a ghost updated through an expression summary. For example:
+`function _.deposit() external => countDeposit() expect void;` where `countDeposit` is a CVL function that increments a ghost
+counter. This does not conflict with the direct-storage-access guideline: that guideline is about mirroring values already
+present in contract storage, whereas this idiom tracks information never stored in contract state.
+</advice>
 </cvl_advice>
--- a/composer/templates/cvl_additions.j2
+++ b/composer/templates/cvl_additions.j2
@@ -43,6 +43,9 @@ transferred across calls — can be tracked with a ghost updated through an expr
 counter. Two caveats: an expression summary *replaces* the summarized function's implementation (so only use it on calls whose
 real effects are not otherwise under test, e.g. calls from contract code to external dependencies), and summaries never apply
 to functions invoked directly from CVL — a ghost counting the verified contract's own entry points called from a rule stays 0.
+Mark such counter ghosts `persistent` when they must survive reverted calls or `at`-snapshot restores (non-persistent ghosts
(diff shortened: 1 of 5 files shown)


tests/test_template_render.py
shellygr, line 1: remove this test. does nothing

composer/templates/cvl_additions.j2
```diff
@@ -25,4 +25,29 @@ where necessary.
...
+`function _.deposit() external => countDeposit() expect void;` where `countDeposit` is a CVL function that increments a ghost
+counter. Two caveats: an expression summary *replaces* the summarized function's implementation (so only use it on calls whose
+real effects are not otherwise under test, e.g. calls from contract code to external dependencies), and summaries never apply
+to functions invoked directly from CVL — a ghost counting the verified contract's own entry points called from a rule stays 0.
+Mark such counter ghosts `persistent` when they must survive reverted calls or `at`-snapshot restores (non-persistent ghosts
+are rolled back along with the blockchain state, which would silently reset the count); this is compatible with the
+persistent-ghost guideline, since a call counter does not mirror contract state.
+This does not conflict with the direct-storage-access guideline: that guideline is about mirroring values already
+present in contract storage, whereas this idiom tracks information never stored in contract state.
+</advice>
```
shellygr:
```suggestion
```

composer/templates/cvl_guidelines.j2
```diff
@@ -35,11 +35,19 @@
...
+        contract state at all (e.g., counting how many times a function is called, or summing the amounts transferred across a sequence of calls).
+    24. A bare `require` that constrains contract state inside a rule or `preserved` block is a spec smell: the assumption may not actually hold in reachable states, silently
+        weakening the proof. The default discipline is to state the assumption as an `invariant`, prove it, and assume it via `requireInvariant`. Bare requires over contract
+        state are acceptable only when accompanied by a comment referencing a justification or a documented trust assumption. Exception: capacity/overflow-avoidance bounds on
+        supplies, balances, or timestamps (e.g. `require totalSupply() <= 2^128`) exist to keep the solver tractable, not to model contract behavior — they are a recognized
+        documented-trust-assumption class; a comment naming the bound's purpose suffices, and no invariant attempt is expected because such bounds are typically not preserved
+        by the contract. In `preserved` blocks the stricter standard applies — `requireInvariant` or verifiable outside evidence; a comment alone is insufficient.
```
jtoman, line 52: we discussed this on zoom, but we already have guidance for both the judge and author to not use `require` as a "make the rule pass" magic. Repeating it a third time will not fix the problem imo. The instructions you provide here are almost verbatim in the author prompt. It would be useful if we could find the run where this happened and see what reasoning/thinking tricks the agent used to talk itself out of the rules it already knew about, rather than just repeating the rule one more time and hoping that one makes it stick.

jtoman (commented):
my 2 c.

PR #44, opened 2026-07-03
KB: 8 articles on storage-slot access, harnessing, vacuity repair, and multi-call idioms

shellygr:
Part 6 of 6 (spec-quality series). The knowledge base had no coverage of the CVL capabilities the generated specs wrongly deemed impossible; these symptom-indexed articles make retrieval surface them at decision time.

## What
Eight additive entries to `CVL_HELP_MESSAGES` (`{title, symptom, body}` format, matching existing style), plus a declared cross-reference graph (`KB_CROSS_REFERENCES`) and a tier-ownership policy comment atop the article list:
1. Keccak-derived/unstructured storage slots (ERC-7201/EIP-1967 constants, slot hooks, harness-getter alternative)
2. Sload/Sstore hook grammar (slot/offset/KEY/INDEX patterns, ghost mirroring, direct-CVL-call caveat) — hook troubleshooting is now one-line pointers to the three pre-existing hook/ghost articles instead of restating them
3. No getter for private/internal state (harness getters, assembly sload, helper decomposition)
4. Vacuous rules / always-reverting methods (NONDET-on-payable root cause, repair ladder; loop cut-off correctly attributed to `optimistic_loop`); symptom quotes the literal prover failure tokens (`SANITY_FAILED`/`SANITY_FAILURE`, `rule_not_vacuous`, `<vacuity_alert>`) so embedding retrieval matches pasted output; scoped to method-level setup vacuity, with the pre-existing require-conflict vacuity article cross-linked rather than competed with
5. Unresolved `.call{value}` HAVOC (`optimistic_fallback` vs DISPATCHER vs mock trade-offs)
6. Multi-call sequences (`lastStorage` / `at` idioms)
7. Satisfy-witness rules for implications
8. Nonlinear mulDiv timeouts — rewritten against #42's single-source math tiers: always-installed `CVLMathAbstract.spec` for the relational tier, `Math.spec` for the exact `*Summary` tier, check `certora/specs/summaries/` before importing, inline `mulDivDownAbstract` fallback for projects predating the shipped assets

- Article 8 rewritten against post-#42 asset names (`Math.spec` / `CVLMathAbstract.spec`, no `CVLMath.spec` copy) — merge #42 before this PR.
- Article 4's symptom now carries the literal failure tokens agents paste from prover output, for retrieval precision.
- The two vacuity articles were deduplicated by scope (single-rule require-conflict vs method-level setup) and cross-linked.
- All informal cross-references replaced with the fixed phrase `article titled "<exact title>"` so a following agent can KBGet by exact key.
- Cross-references declared as a `KB_CROSS_REFERENCES` data literal; the drift-guard test checks declared facts instead of parsing prose (repo no-string-ops policy).
- Tier-ownership comment added atop `CVL_HELP_MESSAGES`: prompts own phase procedure, KB owns symptom-indexed concept knowledge, public docs derive from the concepts.
- Vacuity guidance here stays tool-agnostic; the acknowledgment-ledger mechanics belong to #39/#41 and are not duplicated in KB.

## Notes for review
- `python -m composer.scripts.kb_populate` needs local Postgres and was not run here; the AST-based test validates shape/uniqueness. Please reseed once after merge.

## Tests
`tests/test_kb_populate.py`: 3 passed (shape/uniqueness, plus cross-reference integrity: every declared reference names an existing article and its citing phrase appears in the declaring body). Diff is essentially additive (+407 / −2).


composer/scripts/kb_populate.py
```diff
@@ -498,6 +498,245 @@ class KBMessage(TypedDict):
...
+            "further hooks.\n"
+            "- Non-persistent ghosts are havoced together with storage on unresolved calls — mirror "
+            "values can be wiped even though the hook fired."
+        ),
+    },
+    {
+        "title": "No public getter for the state you need — write a harness",
+        "symptom": "The property needs a private/internal variable, an unstructured storage slot, or an intermediate step of a monolithic function, and the contract exposes no getter for it.",
+        "body": (
+            "Missing getters are a *harness* problem, not a CVL-expressibility problem. Three "
```
shellygr, line 592: here we need to mention that it is relevant if direct storage access is _also_ not applicable, for whatever reason

shellygr:
Delta since the previous version (post architecture review, 4 new commits):

- **Article 8 (mulDiv) rewritten** against #42's single-source math tiers: `CVLMathAbstract.spec` (always installed, relational) + `Math.spec` (exact `*Summary`), with an inline fallback axiom for projects predating the assets. Merge #42 before this PR.
- **Vacuity article retrieval precision**: symptom now quotes the literal prover tokens (`SANITY_FAILED`/`SANITY_FAILURE`, `rule_not_vacuous`, `<vacuity_alert>`); the two vacuity articles were split by scope (require-conflict vs method-level setup) and cross-linked instead of competing.
- **Cross-references formalized**: every reference uses `article titled "<exact title>"` and the graph is declared as a `KB_CROSS_REFERENCES` data literal; the new drift-guard test validates declared facts, no prose parsing.
- **Article 2 deduplicated**: hook troubleshooting is now one-line pointers to the three pre-existing hook/ghost articles.
- **Tier-ownership policy comment** added atop `CVL_HELP_MESSAGES` (prompts = phase procedure, KB = symptom knowledge, docs = public reference).
- Tests: `test_kb_populate.py` grew to 3 (adds cross-reference integrity check); full suite green.

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+            "}\n"
+            "```\n\n"
+            "**If the hook never fires, or the mirror value is unexpectedly wiped**, the usual "
+            "causes have dedicated articles:\n"
+            "- hooks fire only on *contract* code — see the article titled "
+            "\"Hook is not triggered by CVL code\";\n"
+            "- stores inside a hook body do not re-trigger hooks — see the article titled "
+            "\"Hook is not triggered recursively\";\n"
+            "- non-persistent ghost mirrors are havoced together with storage — see the article "
+            "titled \"Ghost variable has unexpected value after an external call\"."
```
jtoman, line 585: this almost directly contradicts advice elsewhere saying to use direct storage access where possible (ghosts/hooks which *purely* mirror storage state is cruft that serves no purpose over just doing direct storage access)

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+        ),
+    },
+    {
+        "title": "No public getter for the state you need — write a harness",
+        "symptom": "The property needs a private/internal variable, an unstructured storage slot, or an intermediate step of a monolithic function, and the contract exposes no getter for it.",
```
jtoman, line 590: again, direct storage access should be fine for private or internal variables!

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+    {
+        "title": "Rule passes vacuously — a method always reverts under the model",
+        "symptom": "The prover reports SANITY_FAILED / SANITY_FAILURE (the rule_not_vacuous sanity sub-rule fails), or the report contains a <vacuity_alert> block flagging a method as vacuous across rules, or a parametric rule's instantiation for one method always reverts even though the method works on-chain.",
+        "body": (
+            "**Root-cause checklist, in order of likelihood:**\n"
+            "1. **`NONDET` summary on a payable or side-effecting callee.** Canonical case: the "
+            "contract forwards ETH to a summarized `deposit()` (e.g. the beacon-chain deposit "
+            "contract). The `NONDET` summary erases the callee's state effects, so checks downstream "
+            "of the call fail on every path and the caller always reverts in the model — every rule "
+            "touching that path becomes vacuous. Inspect the auto-generated summaries first.\n"
```
jtoman, line 626: this is *such* a special case from your recent experience, calling it a "canonical" example is a huge stretch.

A better phrasing is to say the following:

A summary does not capture a side effect a later computation relies on. For example, `whitelistUser(address)` is summarized as NONDET, but the contract code contains:
```
whilelistUser(someUser);
performTrade(someUser); // <- always reverts, whitelists not updated
```

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+        "title": "Unresolved low-level call{value} havocs storage — optimistic_fallback vs DISPATCHER vs mock",
+        "symptom": "Call resolution shows an unresolved low-level `.call{value: ...}(\"\")` (or `send`/`transfer`) resolved as HAVOC; rules fail with arbitrary storage changes or become vacuous.",
+        "body": (
+            "**Default behavior:** an unresolved external call with an empty input buffer havocs "
+            "storage — the Prover assumes the unknown receiver may call back and change anything.\n\n"
+            "**Options, weakest-assumption first:**\n"
+            "- **`\"optimistic_fallback\": true`** (conf flag): unresolved empty-calldata calls no "
+            "longer havoc the caller's storage. The right fix when the call is a *pure ETH transfer* "
+            "to an arbitrary address. Unsound exactly when the receiver could reenter the protocol — "
+            "reentrancy is what the flag assumes away, so don't use it for reentrancy properties.\n"
```
jtoman, line 658: no agent can currently act on this; if we put this we will have agents spin out trying to figure out how to add this flag. Don't mention a remediation that the agent can't use imo

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+            "pessimistic default instead fails loudly with a loop-unwinding violation.)\n"
+            "4. **`lastReverted` overwritten** by a later call — see the article titled "
+            "\"`lastReverted` is overwritten by subsequent calls\".\n\n"
+            "**Repair order — fix the root cause before filtering:**\n"
+            "1. Replace the bad summary with a sound one (expression summary / ghost model of the "
+            "callee).\n"
+            "2. Write a small mock contract and link it into the scene.\n"
+            "3. `\"optimistic_fallback\": true` if the offender is an unresolved raw ETH transfer — "
+            "see the article titled \"Unresolved low-level call{value} havocs storage — "
+            "optimistic_fallback vs DISPATCHER vs mock\".\n"
```
jtoman, line 641: so objection below about unusable mitigations

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+            "contract forwards ETH to a summarized `deposit()` (e.g. the beacon-chain deposit "
+            "contract). The `NONDET` summary erases the callee's state effects, so checks downstream "
+            "of the call fail on every path and the caller always reverts in the model — every rule "
+            "touching that path becomes vacuous. Inspect the auto-generated summaries first.\n"
+            "2. **Conflicting `require`s** (including silent `require_*` casts) in the rule or a "
+            "preserved block — when only one rule is vacuous, see the article titled "
+            "\"Vacuity — rule passes but `assert false` also passes\".\n"
+            "3. **`optimistic_loop: true` with `loop_iter` too small:** paths needing more "
+            "iterations than `loop_iter` are silently assumed away — possibly all of them. (The "
+            "pessimistic default instead fails loudly with a loop-unwinding violation.)\n"
```
jtoman, line 632: same as objection below

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+            "    withdraw(e, shares);\n"
+            "    assert lastStorage[currentContract] == init[currentContract];\n"
+            "}\n"
+            "```\n"
+            "Whole-storage comparison (`==` on `storage` variables, optionally restricted per contract "
+            "with `s[c]`) is supported.\n\n"
+            "**Caveat:** rounding usually breaks exact equalities — prefer inequalities in the "
+            "direction that protects the protocol (\"the user never gains from split/round-trip\")."
+        ),
+    },
```
jtoman, line 703: I don't think this should be in the KB, not as the prompting around it is. Searching on a "symptom" of "how do I ... ?" just doesn't make sense. We should either add a sister tool for this kind of "how to do X" for various, common X patterns with richer search (FTS, full vector search) instead of trying to rephrase everything the agent needs to know as a "symptom".

Put another way, trying to express "here's how to do X" as an symptom: "I don't know how to do X" is not a natural way to find knowledge.

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+            "\n"
+            "methods {\n"
+            "    function MathLib.mulDiv(uint256 x, uint256 y, uint256 d) internal returns (uint256)\n"
+            "        => mulDivDownAbstract(x, y, d);\n"
+            "}\n"
+            "```\n\n"
+            "**Exact tier:** rules that depend on exact values or the overflow revert should instead "
+            "use the `*Summary` functions from `certora/specs/summaries/Math.spec` (present either "
+            "because AutoSetup's curated summaries shipped it or because the pipeline installed the "
+            "same canonical file). Check the `certora/specs/summaries/` directory listing (or the "
```
jtoman, line 761: agent has no idea what autosetup is

composer/scripts/kb_populate.py
```diff
@@ -498,8 +518,301 @@ class KBMessage(TypedDict):
...
+            "// the (slot ...) hook pattern takes a numeric literal; keep the derivation as a comment\n"
+            "hook Sstore (slot 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc)\n"
+            "    address newImpl {\n"
+            "    implMirror = newImpl;\n"
+            "}\n"
+            "```\n\n"
+            "**Alternative — harness getter:** add `contract XHarness is X` with an `external view` "
+            "getter that `sload`s the slot constant — see the article titled "
+            "\"No public getter for the state you need — write a harness\". Prefer hooks for "
+            "tracking *writes*, getters for reading current values inside assertions."
```
jtoman, line 550: except this doesn't help if this is used for anything but a single slot. We *really* don't want the agent trying to reason about these raw storage slots. Honest solution is to add the magic eip-7201 annotations and/or use the editor to make the computed storage slot eip-7201 compliant.

jtoman (changes requested):

composer/scripts/kb_populate.py
shellygr: I tend to disagree here that it cannot act upon it if receiving this from the cex analyzer?

composer/scripts/kb_populate.py
shellygr: We have exemplars on solc0.4. Good luck to the agent then. Agree the solution is not in kb articles

composer/scripts/kb_populate.py
jtoman: the agent has no way to change the prover config to add "optimistic_fallback" I mean. dispatcher vs mock, sure!

composer/scripts/kb_populate.py
jtoman: I guess my claim is that if direct storage access doesn't work, hooks won't work either, **unless** they get lucky and all the state they need is in constant slots.

shellygr:
no more kb

PR #51, opened 2026-07-04
Add curated CVL summaries for Solady SafeTransferLib, CREATE3, LibClone

shellygr:
Curated CVL summaries for three Solady libraries whose inline assembly breaks the prover:

- **SafeTransferLib** — reroute safe wrappers to direct token calls (mirrors the OZ SafeERC20 summary).
- **CREATE3** & **LibClone** — NONDET the deterministic deploy/clone/predict functions (opaque create2 assembly). Limitation: exact-address / duplicate-salt-revert properties become unverifiable.

Gated by library name in `function_summaries.json`. Version/overload variants are handled by the existing prune pass (name/arity) + TypecheckerLoop backstop, so overloads absent from a project go inert.

### DIFF
--- a/certora_autosetup/certora/specs/summaries/Solady/LibClone.spec
+++ b/certora_autosetup/certora/specs/summaries/Solady/LibClone.spec
@@ -1,0 +1,116 @@
+// Summarization of Solady's LibClone (minimal-proxy / ERC-1967 clone factory library).
+//
+// Every LibClone deploy/clone/create/predict function creates or derives an address via raw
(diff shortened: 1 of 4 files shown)


certora_autosetup/certora/specs/summaries/Solady/LibClone.spec
```diff
@@ -0,0 +1,116 @@
+// Summarization of Solady's LibClone (minimal-proxy / ERC-1967 clone factory library).
```
shellygr, line 1: these are quite aggressive. double check with previous projects

PR #53, opened 2026-07-04
Summarizer: reject view-class summaries on payable methods

shellygr:
The interface summarizer's LLM freely wrote `=> NONDET` for payable methods. That erases the callee's ETH effects; a caller that accounts for the transferred value (e.g. a post-call balance check) then reverts on every path, and its rules pass vacuously.

- `composer/cvl/summary_audit.py`: walks the typed CVL AST (no text matching), flags NONDET/CONSTANT/PER_CALLEE_CONSTANT/ALWAYS on payable methods, incl. wildcard and catch-all entries. Payability comes from the build's `all_methods.json`; missing inventory degrades to no-op.
- Summarizer's put tool runs the audit and rejects violations; raw/edit escape hatches removed from this flow so every write is auditable.
- One hard rule added to the summarization prompt.

Tests: 11 new; full suite 114 passed; pyright clean.

### DIFF
--- a/composer/cvl/summary_audit.py
+++ b/composer/cvl/summary_audit.py
@@ -1,0 +1,120 @@
+"""
+Audit for view-class summaries on payable methods.
+
+A view-class summary (``NONDET``, ``CONSTANT``, ``PER_CALLEE_CONSTANT``,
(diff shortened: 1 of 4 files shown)


shellygr:
found the root cause

PR #54, opened 2026-07-05
always regenerate the call-resolution spec

naftali-g:


shellygr (approved):

shellygr (commented):
Actually, just compared to the alternative solution here: https://github.com/Certora/AutoProver/pull/49/changes

Yours is simpler and I prefer it - but a point to consider - is there any meaning to the value of `self.config.skip_call_resolution`?

naftali-g:
> Actually, just compared to the alternative solution here: [#49 (changes)](https://github.com/Certora/AutoProver/pull/49/changes)
> 
> Yours is simpler and I prefer it - but a point to consider - is there any meaning to the value of `self.config.skip_call_resolution`?

hmm.. probably not really anymore...

PR #55, opened 2026-07-06
tweak prompt

jtoman:
rule outcome polarity got switched, same as in the foundry case harumph.

Also noticed the author agent using "rebuttals" to basically "ack" prior round feedback, this was due to loose prompting around the use of that field.


composer/templates/property_judge_prompt.j2
```diff
@@ -145,6 +151,9 @@ transfers a NO-OP, which in turn trivializes any property that reasons about bal
 {% if sort != "greenfield" %}
 Audit the rules to determine if any are likely to fail because they are *OBVIOUSLY* wrong; they assert that the contract does X, but the contract actually does Y
 (and Y is, with overwhelming probability, the correct outcome). For example, if a rule asserts that user's balance *INCREASES* after a withdrawal, this is almost certainly a spec error.
+
+As stated above, do *not* apply this rejection criterion to rules that will fail because they demonstrate a potential vulnerability in the contract code;
+if there is a real bug, a rule failing* is the *intended* result, as it communicates to the developer something must be fixed.
```
shellygr: Missing *?

shellygr (approved):

PR #60, opened 2026-07-08
RIP IDE Bridge

jtoman:
You will (probably not) be missed


naftali-g (approved):
Didn't even notice we had this

jtoman:
> Didn't even notice we had this

:')

PR #63, opened 2026-07-10
Harden AutoFoundry against invalid-Solidity build failures

shellygr:
## Problem

An LLM-authored Foundry draft that contains an invalid hex address literal (a shorthand/"leet" address whose spelling includes a non-hex character) fails to compile. `forge test` then exits non-zero and prints **no** parseable JSON, so `_parse_forge_json` returns `None` and the runner reports the whole campaign as a build failure. Because the publish gate only stamps on a clean *unseeded* run, the contract can never finish — one bad literal sinks the entire test file.

Two gaps enabled this: the generation prompt gave **no** guidance on writing test addresses, and on a build failure the runner dumped the raw solc output with no steer toward "fix the compile error first," so the agent burned its iteration budget without recovering.

## Change

- **Prompt (`foundry_property_generation_system_prompt.j2`)** — under *Authoring guidance*, steer test addresses to `makeAddr("name")` / `vm.addr(k)` (or a full valid checksummed hex literal), and tell the agent to fix a build failure before iterating on test logic or fuzz seeds. Generated tests already inherit forge-std `Test`, so these are in scope.
- **Runner (`composer/foundry/runner.py`)** — a small pure `_build_failure_hint(stderr)` helper leads the build-failure message with a general "fix the compile error first" instruction, lightly specialized toward the address idioms when solc reports the invalid-literal lexical error. Keyed off solc's own error code, **not** a regex over the generated source; the raw stderr/stdout is still surfaced and the publish gate is untouched.
- **Test (`tests/test_foundry_runner.py`)** — pure-function coverage of the hint (general default + address nudge) and `_parse_forge_json`, plus a behavioral test driving the runner's build-failure branch via a stubbed subprocess.

## Verification

- `uv run python -m pytest tests/test_foundry_runner.py` — 5 passing; `tests/test_foundry_entry_args.py` still green.
- `uv run pyright composer/foundry/runner.py` — 0 errors.
- End-to-end: pending a dev AutoFoundry re-run to confirm generated tests use `makeAddr`/valid addresses and that any residual build failure now leads with the hint.


composer/templates/foundry_property_generation_system_prompt.j2
```diff
@@ -103,6 +103,13 @@ err on the side of long-and-specific.
   fail (not just an inner call); see "Expecting test failure" below.
 * When you need to mock, `vm.mockCall` is the cheatcode; look it up with
   `foundry_cheatcodes_manual_search` if you haven't used it before.
+* **Use `makeAddr("name")` or `vm.addr(k)` for test actors/addresses; never
+  hand-write shorthand or "leet" hex.** If a specific literal is truly
+  required, use a full valid checksummed hex address. An invalid hex digit is
+  a hard solc build error that fails the **entire** build -- no tests run.
+* **If `forge_test` reports a build failure, fix the compile error before
+  anything else** -- before iterating on test logic or fuzz seeds. Invalid
+  address/number literals are a common cause.
```
shellygr, line 108: the tricky thing here is that indeed we may get a project with pre-existing failures and then it's legitimate to skip it. I also have no idea how common is the generated leetcode effect.

```suggestion
* **If `forge_test` reports a build failure, fix the compile error before
  anything else** -- before iterating on test logic or fuzz seeds.

```

composer/foundry/runner.py
```diff
@@ -359,6 +359,27 @@ class _ForgeSuiteResult(BaseModel):
...
+    lead = (
+        "The project failed to BUILD -- no tests ran. Fix the compile error "
+        "before iterating on test logic or fuzz seeds."
+    )
+    if "(8936)" in stderr:  # solc lexical error on a malformed number/hex literal
+        lead += (
+            " This looks like an invalid Solidity literal (e.g. a malformed hex "
+            'address). Use `makeAddr("name")` / `vm.addr(k)` for test actors, or '
+            "a full valid hex address."
+        )
```
jtoman: do we really need this? unless we have strong evidence the agent needs the extra help to understand the compilation failure, this feels like overengineering a targeted solution that is already subsumed by the general solution, i.e., telling the agent "your tests failed to compile, fix them"

composer/templates/foundry_property_generation_system_prompt.j2
```diff
@@ -103,6 +103,13 @@ err on the side of long-and-specific.
   fail (not just an inner call); see "Expecting test failure" below.
 * When you need to mock, `vm.mockCall` is the cheatcode; look it up with
   `foundry_cheatcodes_manual_search` if you haven't used it before.
+* **Use `makeAddr("name")` or `vm.addr(k)` for test actors/addresses; never
+  hand-write shorthand or "leet" hex.** If a specific literal is truly
+  required, use a full valid checksummed hex address. An invalid hex digit is
+  a hard solc build error that fails the **entire** build -- no tests run.
```
jtoman: ibid here, this feels like an oversteer.

tests/test_foundry_runner.py
```diff
@@ -0,0 +1,114 @@
...
+        assert "makeAddr" not in hint
+
+
+def test_build_failure_hint_adds_address_nudge_on_lexical_literal_error() -> None:
+    # solc error 8936 is the lexical error on a malformed number/hex literal.
+    hint = _build_failure_hint(
+        "Error (8936): Identifier-start is not allowed at end of a number."
+    )
+    assert "failed to BUILD" in hint  # still leads with the general guidance
+    assert "makeAddr" in hint
```
jtoman: etc.

composer/foundry/runner.py
shellygr: yes, was on the fence with this specific steering. we'll need to see if it's a common failure mode though

jtoman (approved):

PR #65, opened 2026-07-10
Fix packages construction

shellygr:
1. Unify remappings computation, forge remappings as first line, our workarounds second
2. Solve whitespace related issues in more places

### DIFF
--- a/certora_autosetup/build_systems/foundry.py
+++ b/certora_autosetup/build_systems/foundry.py
@@ -22,6 +22,7 @@ else:
 from certora_autosetup.build_systems.base import BuildSystemConfig
 from certora_autosetup.build_systems.manager import BuildSystemManager
 from certora_autosetup.utils.logger import logger
+from certora_autosetup.utils.remappings import build_packages_from_remapping_sources
 from certora_autosetup.utils.types import ContractHandle
 
 
@@ -176,11 +177,20 @@ class FoundryManager(BuildSystemManager):
             # Resolve paths relative to foundry.toml location
             config = self._resolve_paths(config, config_file.parent)
 
-            # Parse remappings and convert to packages format
-            if config.remappings:
-                config.packages = self._convert_remappings_to_packages(
+            # Build the packages list from the authoritative merged remapping sources
+            # (`forge remappings` + foundry.toml + remappings.txt + package.json) so the
+            # initial conf matches what the reactive source-not-found workaround would
+            # produce. Fall back to parsing foundry.toml's explicit remappings when forge
+            # is unavailable and nothing is found on disk.
+            packages = build_packages_from_remapping_sources(
+                base_dir=config_file.parent, log_fn=self.log
+            )
+            if not packages and config.remappings:
+                packages = self._convert_remappings_to_packages(
                     config.remappings, config_file.parent
                 )
+            if packages:
+                config.packages = packages
 
             self.log(
                 f"Parsed foundry config: solc={config.solc_version}, optimizer={config.optimizer}"
--- a/certora_autosetup/build_systems/foundry.py
+++ b/certora_autosetup/build_systems/foundry.py
@@ -177,18 +177,11 @@ class FoundryManager(BuildSystemManager):
             # Resolve paths relative to foundry.toml location
             config = self._resolve_paths(config, config_file.parent)
 
-            # Build the packages list from the authoritative merged remapping sources
-            # (`forge remappings` + foundry.toml + remappings.txt + package.json) so the
-            # initial conf matches what the reactive source-not-found workaround would
-            # produce. Fall back to parsing foundry.toml's explicit remappings when forge
-            # is unavailable and nothing is found on disk.
+            # Build the packages list from forge remappings + foundry.toml + remappings.txt
+            # + package.json for the resolved profile.
             packages = build_packages_from_remapping_sources(
-                base_dir=config_file.parent, log_fn=self.log
+                base_dir=config_file.parent, log_fn=self.log, profile=profile
             )
-            if not packages and config.remappings:
-                packages = self._convert_remappings_to_packages(
-                    config.remappings, config_file.parent
-                )
             if packages:
                 config.packages = packages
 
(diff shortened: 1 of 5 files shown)


certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -457,6 +468,9 @@ def _detect_stack_too_deep_errors(
             line = lines[i]
 
             # Pattern 1: Look for "Compiling <PATH> to expose internal function information"
+            # (two substrings on one line + a marker on the next). Not a proven line-wrap
+            # victim like _has_source_not_found; left as-is because the next-line lookahead
+            # and path extraction depend on the raw line structure.
```
shellygr: this is from certora-cli so not expected to wrap.

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -627,14 +644,21 @@ def _has_remappings_conflict(self, output: str) -> bool:
...
 
     def _has_unnamed_return_warning(self, output: str) -> bool:
         return "Unnamed return variable can remain unassigned" in output
 
     # Two-line signature emitted by the Certora wrapper when a library used by a
     # contract's constructor is not in the compile scope. Matching both lines makes
-    # the detector specific enough that no other prover error collides.
+    # the detector specific enough that no other prover error collides. (Both regexes
+    # search the whole output, so a marker split across a solc wrap could be missed —
+    # not observed in the sweep; unlike _has_source_not_found it isn't normalized here
```
shellygr: drop the part about the sweep.

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: You're right — that progress line is certora-cli output, not solc, so it will not hard-wrap. Removed the misleading comment (af696dd).

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: Dropped the sweep reference; the comment now just notes the two regexes search the whole output and aren't normalized because the \w+/\S+ captures rely on raw token boundaries (af696dd).

certora_autosetup/utils/remappings.py
```diff
@@ -0,0 +1,201 @@
...
+            earlier_source = remapping_key_to_source[key]
+            log_fn(
+                f"Package '{key}' has different paths in {earlier_source} "
+                f"('{remapping_key_to_path[key]}') and {source_name} ('{path}') "
+                f"— using {earlier_source}",
+                "WARNING",
+            )
+        return
+
+    packages.append(f"{key}={path}")
```
jar-ben: [nit] - can we construct `packages` at the end of `build_packages_from_remapping_sources()` from `remapping_key_to_path` instead of gradually building it in `_merge_remapping_entry`? I.e. separate data collection from "formatting".

certora_autosetup/utils/remappings.py
```diff
@@ -0,0 +1,201 @@
...
+                remapping_key_to_source=remapping_key_to_source,
+                warn_on_mismatch=True,
+                base_dir=base_dir,
+                log_fn=log_fn,
+            )
+
+    # Read package.json and add entries not already in remappings
+    package_json_path = base_dir / "package.json"
+    if package_json_path.exists():
+        package_data = json.loads(package_json_path.read_text())
```
jar-ben: no `try & except` here for malformed jsons?

certora_autosetup/utils/remappings.py
```diff
@@ -0,0 +1,201 @@
...
+    - if ``warn_on_mismatch`` and the stored path differs from the new one, log a
+      warning naming the actual earlier source from ``remapping_key_to_source``;
+    - otherwise silently skip.
+
+    Caller is responsible for stripping whitespace and confirming the entry contains
+    an ``=`` before calling.
+    """
+    raw_key, raw_path = entry.split("=", 1)
+    key = raw_key.rstrip("/")
+    path = raw_path.rstrip("/")
```
jar-ben: add `.strip()` for leading whitespace? (patterns like `@oz/ = lib/oz/`)

certora_autosetup/build_systems/foundry.py
```diff
@@ -176,11 +177,20 @@ def parse_config(self, config_file: Path, profile: str | None = None) -> Foundry
             # Resolve paths relative to foundry.toml location
             config = self._resolve_paths(config, config_file.parent)
 
-            # Parse remappings and convert to packages format
-            if config.remappings:
-                config.packages = self._convert_remappings_to_packages(
+            # Build the packages list from the authoritative merged remapping sources
+            # (`forge remappings` + foundry.toml + remappings.txt + package.json) so the
+            # initial conf matches what the reactive source-not-found workaround would
```
jar-ben: I don't like these cross references where two usages of a function (`build_packages_from_remapping_sources` in this case) talks about each other - it's hard to maintain and not needed I think.

certora_autosetup/build_systems/foundry.py
```diff
@@ -176,11 +177,20 @@ def parse_config(self, config_file: Path, profile: str | None = None) -> Foundry
...
+            # produce. Fall back to parsing foundry.toml's explicit remappings when forge
+            # is unavailable and nothing is found on disk.
+            packages = build_packages_from_remapping_sources(
+                base_dir=config_file.parent, log_fn=self.log
+            )
+            if not packages and config.remappings:
```
jar-ben: I would expect this to be a noop code, but it is not - and it is not due to a bug I think? `build_packages_from_remapping_sources` does not assume `profile` (i.e. when the user specifies non-`default` profile). I think `build_packages_from_remapping_sources` should assume `profile` (and then we can drop `_convert_remappings_to_packages`).

tests/test_compilation_workarounds.py
```diff
@@ -50,3 +50,43 @@ def test_detects_single_line_yul_stack_too_deep(manager: CompilationWorkaroundMa
 
 def test_ignores_unrelated_output(manager: CompilationWorkaroundManager) -> None:
     assert manager._detect_yul_exception_stack_too_deep(UNRELATED_OUTPUT) is False
+
+
+# Verbatim-shaped certoraRun output for the "Source ... not found" ParserError. solc
+# hard-wraps the diagnostic, so the two markers ('ParserError: Source "' and
+# "File not found") land on separate lines and defeat a raw substring check. These are
+# the ion-protocol / angstrom cases that must be detected so the source-not-found
```
jar-ben: Do we really need to talk about angstrom and ion-protocol here? also Royco below.

jar-ben (commented):
requesting some answers

certora_autosetup/utils/remappings.py
```diff
@@ -84,12 +89,13 @@ def build_packages_from_remapping_sources(base_dir: Path, log_fn: LogFn) -> List
...
-        foundry_remappings.extend(
-            foundry_data.get("profile", {}).get("default", {}).get("remappings", []) or []
-        )
+        # foundry.toml keeps remappings under `[profile.<name>]` (top-level keys belong to the
+        # default profile). Add the requested profile first so it wins over default under the
+        # first-writer-wins dedup, then default, then any top-level `remappings = [...]`.
+        profiles = foundry_data.get("profile", {})
+        if profile != "default":
+            foundry_remappings.extend(profiles.get(profile, {}).get("remappings", []) or [])
+        foundry_remappings.extend(profiles.get("default", {}).get("remappings", []) or [])
```
shellygr: elide these two lines

jar-ben (approved):

PR #66, opened 2026-07-10
autosetup: apply all applicable compilation workarounds per retry, stop no-op retry loops

shellygr:
This PR collects a bunch of problems that were encountered in many setups. It was validated against a set of about 80 projects with no regressions.

1. idempotency of certain compilation workarounds means we run workarounds and test them to no effect (e.g. warnings handlings)
2. Collect all applicable workarounds from the output
3. Support certora-cli 8.17.1 and enum ambiguity focused on `Math.Rounding` including tests
4. Option to stop after compilation analysis for testing
5. Make sure we do not invoke solc version which isn't installed
6. More precise detection of 'stack too deep' errors


certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -305,9 +305,22 @@ def run_compilation_with_workarounds(
...
                 name="unnamed_return_warning",
-                detect_fn=lambda output: "detected" if self._has_unnamed_return_warning(output) else None,
+                # The flag check must live in detect_fn, not in `enabled`:
+                # `enabled` is evaluated once when this list is built, but the
+                # warning text keeps appearing in the output even after the flag
+                # is set (the flag only stops the warning from failing the run),
+                # so an `enabled`-only check would re-fire this workaround on
+                # every retry. compilation_config is a superset of
+                # updated_config_dict (callers merge the latter in), so it is
+                # the right dict for the live check.
```
shellygr:
```suggestion
```

shellygr: Claude answers: kept — the loop rework was rebased on top of your suggestion, so the comment block stays removed.

certora_autosetup/certora/specs/summaries/OpenZeppelin/OZ_Math.template.spec
```diff
@@ -1,19 +1,7 @@
+// Stub: the OZ Math summary spec is rendered programmatically by
+// SummarySetup._render_oz_math_spec (setup_summaries.py) because its content
```
jar-ben: The `.spec` comments/files should not talk about the python code/autosetup that generates them.

certora_autosetup/setup/setup_prover.py
```diff
@@ -908,6 +910,14 @@ def generate_all_user_defined_types_json(self, build_data: Dict) -> int:
 
                                 # Add to collection if we found a valid type
                                 if type_name and qualified_name:
+                                    # canonicalId is "path/File.sol|Qualified.Name"; the
```
jar-ben: I don't think we need this comment here (we don't have such comments for the other fields we store). It looks like Claude explaining why the code is being added. 

Also, why don't we just store the `canonicalId` instead of parsing the source file from it? Do we actually need the source file anywhere?

certora_autosetup/setup/setup_summaries.py
```diff
@@ -115,6 +115,38 @@ def _cvl_safe_param_name(name: str) -> str:
 
 from certora_autosetup.parsers.type_analyzer import TypeAnalyzer, TypeCategory
 
+# Curated-summary keys whose materialized content depends on the WHOLE scene, not
+# just the batch that matched them — re-rendered on every scene growth.
+SCENE_SENSITIVE_TEMPLATE_KEYS: FrozenSet[str] = frozenset({"oz_Math_mulDiv"})
+
+
+@dataclass(frozen=True)
```
jar-ben: Do these rounding specific classes really belong to the general setup_summaries infra file?

certora_autosetup/setup/setup_summaries.py
```diff
@@ -766,105 +807,216 @@ def copy_summaries_folder(self, matched_function_keys: Iterable[str]) -> Path:
...
+        names: Set[str] = set()
+        for member in enum_members:
+            if isinstance(member, dict) and "name" in member:
+                names.add(member["name"])
+            elif isinstance(member, str):
+                names.add(member)
+        return names
+
+    @staticmethod
+    def _round_up_member(members: FrozenSet[str]) -> Optional[str]:
```
jar-ben: Same comment as above - I would not expect to see rounding specific functionality in the general setup_summaries.py infra.

jar-ben: But I see this is a pre-existing issue...

certora_autosetup/setup/setup_summaries.py
```diff
@@ -766,105 +807,216 @@ def copy_summaries_folder(self, matched_function_keys: Iterable[str]) -> Path:
...
+        ``all_user_defined_types.json``.
+
+        Every scene contract's row set covers the types in its own import closure,
+        so grouping rows by declaring source file yields the DISTINCT ``Rounding``
+        definitions in the scene: one definition -> the plain ``Math.Rounding``
+        spelling works; two or more -> the prover purges the ambiguous name and
+        every reference must be qualified by a contract that imports exactly one
+        definition (``C.Rounding``, certora-cli >= 8.17.1).
         """
+        user_types_file = Path(".certora_internal/all_user_defined_types.json")
```
jar-ben: Can we use a shared constant for this? don't hardcode the filepath here

certora_autosetup/setup/setup_summaries.py
```diff
@@ -766,105 +807,216 @@ def copy_summaries_folder(self, matched_function_keys: Iterable[str]) -> Path:
...
         try:
-            user_types_file = Path(".certora_internal/all_user_defined_types.json")
-            if not user_types_file.exists():
-                self.log("Warning: all_user_defined_types.json not found", "WARNING")
-                return None
-
             with open(user_types_file, "r") as f:
                 user_types = json.load(f)
+        except Exception as e:
+            raise Exception(f"Error reading user types file: {e}")
```
jar-ben: Where do we catch this exception? I.e. how much is killed by this failure? I would hope it kills/prevents only the summaries that depends on rounding.

certora_autosetup/setup/setup_summaries.py
```diff
@@ -907,7 +1059,15 @@ def _materialize_template(self, template_path: str, main_contract: str) -> str:
         dst = self.user_summaries_dir / versioned_rel_under
         dst.parent.mkdir(parents=True, exist_ok=True)
 
-        self.process_template_in_place(src, dst, main_contract)
+        if rel_under_summaries.name == "OZ_Math.template.spec":
```
jar-ben: Adding these special case handlings (like `if rel_under_summaries.name == "OZ_Math.template.spec"`) to the general setup_summaries.py will make the maintenance in the future really hard. Especially if we will have a bunch of such `if`s in the code and we will have to sync changes across all the occurrences. Having OZ/Math specific things in a dedicated file/class would help. 

It can be done later, but I am afraid we will just keep growing this already huge file and it will be harder and harder to maintain.

certora_autosetup/utils/solc_version_resolver.py
```diff
@@ -246,11 +256,20 @@ def resolve_pragma_to_version(
...
-        # Select highest matching version
-        highest_version = max(matching_versions, key=Version)
+        # Prefer the highest matching version whose solc binary is actually
+        # installed: the newest release on soliditylang.org may not exist in
+        # the environment yet (nor be supported by the prover toolchain), and
+        # resolving a floating pragma to it breaks every consumer the day a
+        # new solc ships. Fall back to the listed highest when no matching
+        # binary is installed (environments that fetch compilers on demand).
+        installed_versions = [v for v in matching_versions if _solc_binary_installed(v)]
+        pool = installed_versions or matching_versions
```
jar-ben, line 275: Can you update the docs of `resolve_pragma_to_version` (to say we prioritise installed solc versions)?

certora_autosetup/typechecker_loop.py
```diff
@@ -85,14 +85,42 @@ def _parse_typechecker_errors(
         external_matches = re.findall(external_method_pattern, error_output)
         matches.extend(external_matches)
 
-        # Pattern 2: Math.Rounding enum type errors
+        # Pattern 2b: ambiguous/unresolvable Rounding type with suggestions
```
jar-ben: where is the pattern `2a`?

certora_autosetup/utils/solc_version_resolver.py
```diff
@@ -91,6 +92,15 @@ def fetch_available_solc_versions() -> List[str]:
             return FALLBACK_VERSIONS
 
 
+def _solc_binary_installed(version: str) -> bool:
+    """Whether a solc binary providing `version` is on PATH, under either the
+    Certora naming convention ("solc8.35") or solc-select's ("solc-0.8.35")."""
+    names = [f"solc-{version}"]
+    if version.startswith("0."):
```
shellygr: you assume it's true here. otherwise it's going to be just certora naming convention in all cases. so write what you expect to get as version string. we have utility functions for converting version names already so no need to reinvent the wheel

certora_autosetup/typechecker_loop.py
```diff
@@ -375,6 +411,159 @@ def _is_in_summaries_folder(self, spec_file: str) -> bool:
...
+        return expanded
+
+    @staticmethod
+    def _rounding_members_by_qualifier(qualifiers: set) -> Dict[str, set]:
+        """Member names of the Rounding enum each qualifier contract sees, from
+        ``all_user_defined_types.json``. A suggested qualifier sees exactly one
+        Rounding definition (that is why the prover suggests it), so its rows
+        identify which definition — and which round-up member — it stands for."""
+        result: Dict[str, set] = {q: set() for q in qualifiers}
+        types_path = Path(".certora_internal/all_user_defined_types.json")
```
jar-ben: use a constant

certora_autosetup/typechecker_loop.py
```diff
@@ -375,6 +411,159 @@ def _is_in_summaries_folder(self, spec_file: str) -> bool:
                 return True
         return False
 
+    @staticmethod
+    def _expand_to_function_blocks(lines: List[str], reported: set) -> set:
```
jar-ben, line 347: Can we get this from the CVL AST instead of custom string parsing?

certora_autosetup/utils/solc_version_resolver.py
shellygr: Claude answers: fixed in 1009945 — the probe now documents its input contract (raw semver exactly as listed by fetch_available_solc_versions) and builds the Certora spelling via the shared convert_solc_version_to_certora_format instead of the hand-rolled slice. (format_solc_version in enhanced_config_manager would be the fuller utility, but enhanced_config_manager imports from this module, so the leaf converter avoids the cycle.)

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -395,39 +439,43 @@ def run_compilation_with_workarounds(
...
+                    updated_config_dict,
+                    compilation_config,
+                    config_file,
+                    contracts,
+                )
+                # If we disabled build_cache in config, also remove from CLI command
+                if workaround.name == "cached_autofinder_failure" and "--build_cache" in cmd:
+                    cmd.remove("--build_cache")
+                applied_this_pass.add(workaround.name)
+                if workaround.exclusive:
```
jar-ben: In the current code, it is _exclusive_ only if no workarounds triggered before, i.e. the order matters. Currently, `cached_autofinder_failure` is the only exclusive workaround and it is first in the `workarounds` list, so the exclusivity is enforced, but if it would not be first, it's no longer guaranteed to be exclusive. 

I would expect either a different naming (to allow non-exclusivity) or enforcing the exclusivity.

jar-ben (changes requested):

certora_autosetup/setup/setup_prover.py
shellygr: When you have Math.Rounding declared inside another file that imports some `Math`, we have to parse out the source file.

shellygr: turns out we get in another way from the ast, so did a deeper change for using the canonicalId for the grouping only

certora_autosetup/setup/oz_math_rounding.py
```diff
@@ -0,0 +1,275 @@
...
+        return "Ceil"
+    if "Up" in members:
+        return "Up"
+    return None
+
+
+def _load_math_rounding_rows(log: LogFn) -> List[Dict]:
+    """Rounding-enum rows declared inside a ``Math`` library, from the
+    compiled-scene type inventory. Any read failure degrades to an empty list
+    (=> "none" classification, no directional summary) rather than aborting
```
jar-ben: `rather than aborting the whole setup.` can be dropped

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -471,8 +473,25 @@ def run_compilation_with_workarounds(
...
                 applied_this_pass.add(workaround.name)
-                if workaround.exclusive:
-                    break
+
+            # Exclusive workarounds go first, regardless of list position: they
+            # invalidate the whole output, so when one applies the pass ends
+            # there and nothing else may act on it.
+            for workaround in workarounds:
+                if workaround.exclusive and workaround.enabled:
+                    detect_result = workaround.detect_fn(output)
```
jar-ben: [nit] the three lines can go to a helper function, because you use it also below. 

```
detect_result = workaround.detect_fn(output)
if detect_result is not None
apply_workaround(workaround, detect_result)
```

with a boolean return value saying if the workaround was actually applied or not (so that you can trigger the `break`)

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -612,6 +631,43 @@ def _detect_cancun_opcode_errors(self, output: str, contracts: List[ContractHand
...
+        segment: List[str] = []
+
+        def segment_hit() -> Optional[str]:
+            if current_path and marker in _normalize_ws("\n".join(segment)):
+                return self._get_contract_name_from_path(current_path, contracts)
+            return None
+
+        for line in output.split("\n"):
+            path = _path_from_compiling_line(line)
+            if path is not None and "to expose internal function information" not in line:
```
jar-ben, line 657: can you comment on when `"to expose internal function information"` appear / what are the implications?

jar-ben (approved):
looks ok, thank you for the changes. Leaving just a few minor comments

certora_autosetup/utils/compilation_workarounds.py
shellygr: it's matching on the certora-cli output in a very clunky way

PR #71, opened 2026-07-14
make the prover global_timeout overridable via AUTOPROVER_GLOBAL_PROVER_TIMEOUT

jar-ben:
`make_prover_options` now reads an optional `AUTOPROVER_GLOBAL_PROVER_TIMEOUT` env var for the cloud `--global_timeout` value (falling back to the 7200s default).


naftali-g (approved):
LGTM

PR #72, opened 2026-07-14
Code editor

jtoman:
Prototype for editor agent to allow the cvl author to request (limited) changes to the source code to work around prover limitations (unstructured/custom storage, refactoring, etc.)

Gated with a judge agent as usual.


composer/spec/source/munge/compile_check.py
```diff
@@ -0,0 +1,153 @@
...
+            f"files not present under config['files']: {absent}"
+        )
+
+    with accessor.materialize(state) as tmp:
+        folder = Path(tmp)
+        (folder / _CONF_NAME).write_text(json.dumps(config))
+
+        result, stdout = await run_prover_inner(
+            folder,
+            [_CONF_NAME, "--build_only"],
```
shellygr, line 131: do you want to go further and use --compilation_steps_only to make sure the changes are compatible with the spec? or is that handled elsewhere?

composer/spec/source/munge/edit_oracle.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+_ORACLE_SYSTEM = """\
+You judge whether a previously recorded finding about smart-contract source code is still accurate after the source was edited.
+
+You are given the finding (the original question and the answer that was recorded) and a unified diff from the source version the finding was written against to the current version.
+
+Decide whether anything in the diff contradicts or outdates the answer. If every claim the answer makes still holds against the changed source, it is still valid. If the diff changes something the answer depends on — a moved or renamed symbol, altered control flow, a removed branch, a changed signature — it is no longer valid, and you say in one sentence what changed.
+
+Edits are usually small and surgical, so most findings are unaffected; mark a finding invalid only when the diff actually touches what the answer relies on."""
```
shellygr: is it given the original or current version? some changes could look local but they are not.

composer/spec/source/munge/edit_oracle.py
```diff
@@ -0,0 +1,101 @@
...
+        "the answer. Left empty when it still holds.",
+    )
+
+
+def mk_oracle(
+    llm: BaseChatModel,
+    edit_store: EditStore,
+    sc: SourceCode,
+) -> MigrationOracle:
+    """Build a :class:`MigrationOracle` that decides, in one LLM call, whether a
```
shellygr: as above, point for future review - not sure it's a one-shot call

composer/spec/source/munge/munge_agent.py
```diff
@@ -0,0 +1,426 @@
...
+
+class CommonMungeDescription(
+    BaseModel
+):
+    """A holistic description of your changes"""
+    executive_summary: str = Field(description="An executive summary of your changes, fully covering each part of the diff with the existing source code")
+
+    how_to_apply: str | None = Field(description="What changes might need to be made by the upstream verification author to make effective use of your changes")
+
+    why_sound: str = Field(description="A precise, reasoned argument why your changes are *either* sound, OR an acceptable over-approximation")
```
shellygr, line 58: I read "over-approximation" as "sound". Did you mean "acceptable under-approximation"?

composer/spec/source/author.py
```diff
@@ -328,6 +355,196 @@ async def run(self) -> Command | str:
...
+    editor = editor_tool(
+        ctx=ctx,
+        env=source_env,
+        edit_tools=live_tools,
+        edit_store=edit
+    )
+    return [
+        editor,
+        # "commit_edit" is the name the editor's result message tells the author
+        # to call (see EditMungeTool.run's result_msg) — keep them in sync.
```
shellygr: 🤔 
isn't there a less fragile way to maintain this?

composer/spec/source/live_explorer.py
```diff
@@ -0,0 +1,229 @@
...
+an up-to-date answer.
+"""
+
+def _prover_output_dirs(p: PurePath) -> bool:
+    """Globally exclude prover outputs from the live tool surface AND the
+    materializer: they are never compilation inputs, and copying prior runs'
+    ``.certora_internal`` / ``emv-*`` trees into each prover-run
+    materialization would be enormously wasteful."""
+    return bool(p.parts) and (
+        p.parts[0] == ".certora_internal" or p.parts[0].startswith("emv-")
```
shellygr, line 149: what about .git and other directories that might not be of interest?

composer/spec/source/pipeline.py
```diff
@@ -183,7 +186,7 @@ async def prepare_formalization(self, run: PipelineRun) -> Formalizer[GeneratedC
         )
 
         invariant: tuple[list[PropertyFormulation], Delivered[GeneratedCVL]] | None = None
-        if invariants.inv:
+        if invariants.inv and False:
```
shellygr: ??

composer/spec/source/pipeline.py
```diff
@@ -208,7 +211,12 @@ async def prepare_formalization(self, run: PipelineRun) -> Formalizer[GeneratedC
                         description="Structural invariant CVL",
                         source=run.source,
                         spec_dir=SPECS_DIR,
-                        spec_stem=InvariantSpec().stem
+                        spec_stem=InvariantSpec().stem,
+                        # Invariants are assumed as preconditions by every
```
shellygr, line 258: are you sure about this?

composer/templates/munge_charter.j2
```diff
@@ -0,0 +1,61 @@
+## Your Charter
+
+You may make exactly three kinds of edit.
```
shellygr: did you mean
```suggestion
You may make exactly three kinds of edits.
```

composer/templates/munge_charter.j2
```diff
@@ -0,0 +1,61 @@
...
+
+### Exposure
+
+Add an external harness function so the specification can reach an internal or private computation
+directly. Harnesses are additive and thin: forward the arguments, call the target, return the
+result, no logic of their own. Leave the target's own visibility unchanged, and give the harness the
+target's effective mutability (a wrapper around a view computation is view). Name harnesses so they
+read as verification scaffolding, not product code. Note in your report that the harness joins the
```
shellygr, line 11: maybe noted later, but if we harness around state-modifying functions, we may want to omit those from the methods list via filter in parametric rules and invariants (not necessarily though)

composer/templates/munge_charter.j2
```diff
@@ -0,0 +1,61 @@
...
+contract's external surface.
+
+### Refactor
+
+Restructure code without changing behavior — typically, hoisting a computation the author names out
+of its surrounding function into a standalone internal function. Shape the new function so its
+signature carries its whole behavior: everything it consumes arrives as a parameter, everything it
+produces leaves as a return value, and state writes stay at the call site whenever feasible. The
+call site must behave identically after the hoist: same reads, same writes, same reverts, same
+evaluation order.
```
shellygr: also same calls if such are done in the refactored snippet

composer/templates/munge_charter.j2
```diff
@@ -0,0 +1,61 @@
...
+
+### Standardize
+
+Rewrite a prover-hostile construct into ordinary Solidity with the same logical behavior: inline
+assembly into source-level equivalents, hand-rolled storage packing into separate declarations,
+exotic bytes/string tricks into standard operations. Representation may change; logical behavior may
+not — same abstract state, same input/output/revert/event behavior, up to the change of
```
shellygr:
```suggestion
not — same abstract state, same input/output/revert/call/event behavior, up to the change of
```

composer/templates/munge_charter.j2
```diff
@@ -0,0 +1,61 @@
...
+64 bits stays a `uint64`, or the widening is called out in your report.
+
+## Lines You Do Not Cross
+
+- Never add a `require`, an `assert`, or an explicit `revert` — and never remove one. The revert
+  surface of existing code is behavior; preserve it. That a new guard only excludes unreachable
+  states is not a justification — reachability arguments are the verification's job, not yours.
+  This forbids guards, not arithmetic: the implicit overflow checks of ordinary Solidity math are
+  fine wherever you write new code — they need no justification and no mention in your report. The
+  one distinction that matters is deliberate wrapping: where the original computed modularly
```
shellygr: doesn't this fall under semantics preservation?

composer/templates/munge_charter.j2
```diff
@@ -0,0 +1,61 @@
...
+  (assembly arithmetic, an `unchecked` block), keep the rewrite `unchecked`; everywhere else, write
+  normal math and move on.
+- Never delete or stub out a function to make a problem disappear. When "pretend this function does
+  not exist" is genuinely right, CVL has a deleting summary for it — declared in the spec,
+  hard-erroring if the function is ever reached — which keeps the deletion visible and under the
+  spec's control. Refusing with that pointer is a correct outcome of your work.
+- Never fix a bug — not as a drive-by, and not when the request asks you to. A proof that fails
+  because the code is defective is the process succeeding: surfacing that defect is the intended
+  product of the verification effort, and repairing the source so the proof goes through destroys
+  the finding. If a standardization forces you to reproduce behavior that looks wrong — an
```
shellygr, line 65: if we already produce a findings report, we could include there findings stemming from the process of writing the rules and not just rules' results. happens also in real life

composer/templates/munge_editor_system.j2
```diff
@@ -0,0 +1,86 @@
...
+scheme, every caller of a function you restructure. Classify the fix against the charter; if no
+category fits, refuse. Make the edit minimally but completely, in one coherent pass.
+
+Request review before submitting. Address feedback by changing the code, not by arguing with it,
+and re-request review. If feedback would push you outside the charter, refuse rather than comply.
+
+## Reporting
+
+Your report accompanies both review and submission, and is consumed by the author without re-reading
+the diff — and later by an auditor with it.
```
shellygr: what is the last "it" here

composer/templates/munge_feedback_system.j2
```diff
@@ -0,0 +1,91 @@
...
+## What You Receive
+
+- The author's original request to the editor. It is provided so you can measure the edit against
+  what was asked; the request itself is not under review.
+- The editor's report: an executive summary, a soundness argument, and optional integration notes.
+- The diff of the edits.
+
+Your file tools show the source with the edits applied. Mind the asymmetry: the diff is your only
+view of the before-state — its removed lines are the original code — while the tools show only the
+after-state. Read the full bodies the diff excerpts; a hunk that looks clean in three lines of
```
shellygr, line 28: now all clear

composer/templates/munge_feedback_system.j2
```diff
@@ -0,0 +1,91 @@
...
+
+## What to Hunt, in Order of Blast Radius
+
+1. **Lost behavior.** In every removed region, account for each storage write, event emission,
+   external call, and revert path: where did it go in the new code? A translation that silently
+   drops one produces a proof about a contract that does not exist. Do not take the soundness
+   argument's word for it; trace each one yourself.
+2. **Added constraints.** New `require`/`assert`/`revert` statements, or checked arithmetic where
+   the original deliberately wrapped (assembly math, `unchecked` blocks). These narrow the state
+   space and can make unsound verification pass. That the request asked for one is not a defense.
```
shellygr:
```suggestion
   space and can make unsound verification pass. That the request asked for a constraint is not a defense of the edit.
```

composer/templates/munge_feedback_system.j2
```diff
@@ -0,0 +1,91 @@
...
+6. **Scope.** Every hunk must serve the request. Drive-by edits — renames, reformatting, refactoring
+   beyond what was asked — are rejectable even when harmless: every touched line is a trust
+   assumption someone must later audit.
+7. **Report fidelity.** The executive summary must cover every hunk — a change the summary omits is
+   a rejection by itself — and the soundness argument's category claim must match what the diff
+   actually does.
+
+## Do Not Reject For
+
+- Incidental overflow reverts from ordinary checked Solidity math in new code. These are expected
```
shellygr: so solidity to solidity new reverts due to math are okay but assembly/unchecked to checked solidity is not

composer/templates/property_generation_system_prompt.j2
```diff
@@ -7,6 +7,28 @@
...
+- **`commit_edit`** — apply a staged edit, by its id, to your working copy. From that point on your
+  file reads, the feedback judge, and the prover all see the edited source.
+- **`edit_history_log`** — list the edits you have applied, with each one's summary and diff.
+- **`revert_to_edit`** — roll your working copy back to an earlier applied edit, discarding the
+  edits applied after it.
+
+Treat source edits as a last resort, reached only after spec-side approaches have failed: every
+applied edit becomes part of the final deliverable, where a human auditor must read and accept it.
+After applying or reverting an edit, prover results and feedback verdicts obtained before the
+change no longer describe your working copy — re-run them before relying on them.
```
shellygr, line 34: are we stamping runs with the latest vfs digest/edit id?

shellygr (commented):
do we include edits in the final report? I missed it here

composer/spec/source/author.py
jtoman: I'm not sure what I follow; what fragility are you foreseeing here? The alternative is to have the edit agent make direct edits to the author's state (bad) or provide a way for the author agent to say directly what edits to apply, but that's just a backdoor to synthesize edits without the edit author. Having agents communicate results between each other with keyed ids is pretty effectively used elsewhere, and is both economical token wise and the only way to make sure an agent doesn't misrepresent or garble results.

composer/spec/source/munge/compile_check.py
jtoman: oh that's a great point, if we change a storage layout that the spec is directly accessing.

But now that I'm thinking it through; the editor can't change the spec. If we gate completion on "the spec still works", the editor can't propose changes that break the spec, even if the author fully intends to fix the spec in response. The agent being able to rollback its edits are the defense against the editor agent completely borking the spec.

composer/spec/source/munge/edit_oracle.py
jtoman: it is shown the *diff*. And you're right, a change in `foo()` may or may not be relevant to `gorp`, depending on whether `bar()` (called by `gorp`) itself calls `foo`.

I think I will make this prompt pessimistic to account for this wrinkle. Ugh, of course it wasn't easy.

composer/spec/source/munge/munge_agent.py
jtoman: I did! good catch

composer/spec/source/munge/edit_oracle.py
jtoman: why wouldn't it be? it's operating on raw `BaseChatModel` with ainvoke; we aren't calling a graph.

composer/spec/source/live_explorer.py
jtoman: blocked by the "global exclude" pattern we setup. Honestly these folders probably should be added to the global exclude as well.

composer/spec/source/pipeline.py
jtoman: :facepalm: debugging, ooooops

composer/spec/source/pipeline.py
jtoman: Depends on what "this" is; if you mean "assume as a precondition by every downstream spec": no the wording is sloppy.

If you mean "they must hold against the unedited source", we could let the invariant agent make edits that all future cvl agents run against, but that seemed like a bad idea for reasons I couldn't really articulate.

composer/templates/munge_charter.j2
jtoman: sure did

composer/templates/munge_charter.j2
jtoman: what do you mean?

composer/templates/munge_editor_system.j2
jtoman: In retrospect, I have no idea what this sentence is trying to say, I'll rewrite it.

composer/templates/munge_feedback_system.j2
jtoman: yeah exactly. If somehow one plus turns into two, I don't want a munge getting rejected with "YOU ADDED A REVERT".

composer/templates/property_generation_system_prompt.j2
jtoman: we should!

composer/spec/source/prover.py
```diff
@@ -284,10 +340,17 @@ async def verify_spec(
                 if rules is None and all_verified:
                     return tool_state_update(
                         tool_call_id=tool_call_id, content=result.result_str,
-                        prover_link=result.link, validations=stamper(state),
+                        prover_link=result.link,
+                        validations=stamper(state, state["version_history"]),
```
jtoman, line 388: this is where we include the version history (which implicitly includes the vfs).

composer/spec/source/munge/compile_check.py
shellygr: fair enough

composer/spec/source/munge/edit_oracle.py
shellygr: I meant that maybe it's better to allow for calling a graph?

composer/spec/source/author.py
shellygr: I meant the tool names; if those are so stable we can afford keeping them constant strings that's fine

composer/spec/source/pipeline.py
shellygr: I meant the latter. representation change could lead to changes in invariants

composer/templates/munge_charter.j2
shellygr: the whole discussion on artihmetics, checked vs. unchecked seems to be falling under the general umbrella of preserve semantics of the code. maybe it's needed to be very explicit from your experimentation, which is a fair argument

composer/templates/munge_editor_system.j2
shellygr: I think it means the report is consumed by 2 authors, one having the diff and one without, but it's a clunky way of conveying it

shellygr: maybe because I'm not a native speaker, llm-speak is sometimes unparseable to me

composer/spec/source/author.py
jtoman: oh oh oh, yes, let's make those constants.

composer/spec/source/pipeline.py
jtoman: worst case the author stops importing the invariants.spec, and if need be, hand picks those it needs, adapting as necessary right?

composer/templates/munge_charter.j2
jtoman: yeah I'm fairly certain I added this out of an imagined failure mode. I think I'll drop it entirely.

composer/spec/source/munge/edit_oracle.py
jtoman: mooted; I've basically ripped out all of this

composer/spec/source/munge/edit_oracle.py
jtoman: mooted (o7)

composer/spec/source/author.py
```diff
@@ -247,35 +254,64 @@ class RemoveFile(BaseModel):
...
-    NB that the link field *must* be at the top-level of the contract's storage. Link flags cannot be used
-    to link fields in structs.
+class SetStorageExtensionHarnesses(BaseModel):
+    """
+    Replace the `storage_extension_harnesses` prover flag: (contract, harness) pairs
+    declaring that `harness` describes storage extensions of `contract`. The provided
+    pairs replace any previously configured ones; an empty list removes the flag from
+    the configuration.
     """
-    type: Literal["add_link"]
```
shellygr, line 236: where did the add_link and friends go?

composer/spec/source/author.py
```diff
@@ -490,8 +524,15 @@ class _LastAttemptEdits(BaseModel):
...
     restored), and the window is two consecutive store puts.
+
+    ``config`` snapshots the prover config for the same reason the history is
+    snapshotted: ``config_edit`` changes (e.g. an added file registering an
+    editor-minted harness) must survive the restart together with the working
+    copy they describe. None on records written before the field existed, in
+    which case resume falls back to the initial config.
     """
     version_history: list[str]
+    config: dict | None = None
```
shellygr, line 552: does this take into account changes in base imported configs?

composer/templates/property_generation_system_prompt.j2
```diff
@@ -21,7 +21,10 @@ dedicated editor agent: you describe the problem you need solved, and you evalua
   file reads, the feedback judge, and the prover all see the edited source.
 - **`edit_history_log`** — list the edits you have applied, with each one's summary and diff.
 - **`revert_to_edit`** — roll your working copy back to an earlier applied edit, discarding the
-  edits applied after it.
+  edits applied after it; a null edit id discards every edit, returning to the unedited source.
+- **`config_edit`** — add or remove files in the prover's input set. Applying an edit does not
```
shellygr: but also storage extensions args?

shellygr (commented):
we're almost there

composer/spec/source/author.py
jtoman: link blocks!

composer/spec/source/author.py
jtoman: it has no way to modify those configs, so I'm not sure what you're referring to here.

composer/templates/property_generation_system_prompt.j2
jtoman: er, right.

shellygr (approved):

shellygr (approved):

PR #73, opened 2026-07-14
Command sandbox: confine untrusted native command execution (Landlock + seccomp)

ericeil:
## What this is

A **standalone command-sandbox mechanism** — a Python seam + a small trusted Rust launcher — that confines a single command to an unprivileged, in-process kernel sandbox (**Landlock + seccomp**). It is the first of the stacked PRs splitting `eric/crucible` (see `docs/pr-split-plan.md`).

```python
from composer.sandbox import run_local_command, rust_build_policy, SandboxPolicy
```

It has **no `composer` dependencies outside `composer.sandbox`** and is stdlib-only, so it can land and be consumed ahead of everything else on the branch.

## Why this code exists

For the Solana ecosystem, AutoProver will need to run commands that **compile and/or execute untrusted native code** — LLM-authored harnesses (`setup`/`action_*`/`build.rs`), the analyzed program's own `build.rs`/proc-macros, and/or the Cruicible fuzzing harness, for example. Without a sandbox, those will run with the **full ambient authority of the AutoProver process**: `ANTHROPIC_API_KEY`, `CERTORA*`/`AWS_*` cloud tokens, `PG*` DB creds, the network route to `169.254.169.254` (IMDS → IAM role creds on EC2), and the entire bind-mounted host project.

The outer container protects the **host from AutoProver**. It does **not** protect AutoProver's own secrets, egress, and filesystem from code running *inside* it. A separate trust boundary already ensures **the LLM authors only file *contents*, never the command line** — but that says nothing about what a command, once running, can *reach*. This mechanism is that missing boundary: each `RunCommand` runs with **no network, no inherited secrets, and only its declared inputs on disk**, proven by an escape test.

## Why we can't use an existing solution

**Namespace sandboxes (bwrap / nsjail) — they fight the container.** Their model (unprivileged user + mount + net namespaces, then `pivot_root`) is exactly what Docker's default seccomp + AppArmor block. Validated empirically (stock `python:3.12-slim`, uid 1000):

| Approach under Docker defaults | Outcome |
|---|---|
| unprivileged `bwrap` | ✗ userns creation blocked by default **seccomp** |
| `bwrap`, `seccomp=unconfined` | ✗ `mount --make-rslave` blocked by **AppArmor** `docker-default` |
| `bwrap`, `seccomp=unconfined` + `apparmor=unconfined` | ✓ works — but requires **weakening the whole container's LSMs** (rejected) |
| setuid `bwrap` | ✗ `capset` blocked (Docker drops `CAP_SETPCAP`) |

Making bwrap work means either stripping the container's own seccomp/AppArmor (widening the host-kernel attack surface across *all* of AutoProver — the opposite of the goal) or adopting a heavier runtime.

**gVisor / Kata — wrong cost, wrong layer.** They work, but (a) they impose their *heaviest* overhead precisely on our syscall/I/O-bound compile+fuzz workload, and (b) their benefit — protecting the host kernel — is an **infrastructure** boundary already provided on EC2 by the Nitro hypervisor. Not worth coupling this boundary to a deployment/runtime decision.

**Off-the-shelf Landlock tools — close, but each misses the hard part.**
- **[`landrun`](https://github.com/zouuup/landrun)** (Go, MIT): great for Landlock FS + env, but blocks network via **Landlock TCP-only rules (kernel ≥6.7)** — it does **not** block UDP/DNS, fails *open* on older kernels, and has no rlimits. It would need a seccomp companion anyway, so it doesn't save the hard part.
- **[`sandlock`](https://github.com/multikernel/sandlock)** (Python+Rust, Landlock+seccomp): the closest match, but requires **kernel ≥6.12** (above Amazon Linux 2023's 6.1), ships an **unstated license**, and carries far more surface than we need (MITM proxy, COW, notification supervisor).

## The chosen model: the process sandboxes itself

Instead of building a namespace *around* the command, the command **restricts itself** using two unprivileged kernel facilities — the model Chrome, OpenSSH, and systemd use. It needs **no namespaces, no capabilities, no root, and no `--security-opt`**, and runs in a **stock** container. The launcher composes two mature, permissively-licensed crates rather than hand-rolling primitives:

- **[`landlock`](https://crates.io/crates/landlock)** — default-deny FS ruleset; grant rw to the workdir, r+x to toolchain paths; closes the `/proc/<parent>/environ` secret leak.
- **[`seccompiler`](https://crates.io/crates/seccompiler)** (AWS Firecracker's seccomp-BPF compiler) — deny inet sockets (TCP **and** UDP/DNS **and** IMDS) and the same-uid secret vectors (`ptrace`, `process_vm_readv`).
- plus an **env allowlist** (scrubbed `execve`) and **rlimits** (`RLIMIT_AS`/`CPU`/`NPROC`/`FSIZE`).

Both confinements are preserved across `execve` and inherited across `fork`, so every descendant (`cargo`, `rustc`, each `build.rs`, the linker) runs confined. It wins for now on **kernel floor (5.13)**, **license clarity**, and **minimal surface** — and the `SandboxProvider` seam keeps `landrun`/`sandlock` swappable later with **no change to the policy or the gate**. Because it needs nothing from the container, the identical code path runs on a dev laptop, EC2, ECS, EKS, Fargate, and under `runc` or gVisor alike.

## What's in the PR (20 files, +2.4k)

- **Python `composer/sandbox/`** (stdlib-only): `policy.py` (tool-agnostic seam: `SandboxPolicy`/`SandboxProvider`/`LaunchSpec`, `none` passthrough, provider registry, fail-closed helpers), `command.py` (`run_local_command` — the `RunCommand` primitive), `launcher.py` (the `launcher` provider), `recipes.py` (`rust_build_policy` + `DEFAULT_ENV_PASSTHROUGH`), `config.py`.
- **Rust `rust/run-confined/`**: the trusted launcher; workspace root trimmed to `members = ["run-confined"]` (the pyo3 framework crates rejoin in later PRs), `Cargo.lock` regenerated for that subset.
- **Packaging/docs**: `scripts/Dockerfile` builds `run-confined` onto `PATH`; `.dockerignore`/`.gitignore` exclude the Rust build output; `docs/command-sandbox.md` (full design + threat model).

## Testing

`tests/test_sandbox{,_command,_config,_launcher,_run_confined,_escape}.py` → **42 passed, 0 skipped**. The **escape gate** exercises every vector (write outside workdir, read host files, read `/proc/<parent>/environ`, `ptrace` the parent, open an inet socket) — all denied — plus an unconfined control. The `run-confined` binary was built and genuinely exercised (Landlock ABI present on the CI kernel), not skipped.

## Notes for reviewers

- **Base:** branches from a clean ancestor of `master`; verified **zero file overlap** with the commits `master` has since gained, so it merges cleanly and the diff shows only the 20 sandbox files.
- **Stacked-PR context:** this is PR 1 of 4 (sandbox → ecosystem → rust framework → crucible app). The framework/Crucible consumers land in later PRs; this PR stands alone.


docs/command-sandbox.md
```diff
@@ -0,0 +1,512 @@
...
+`RunCommand` effect deliberately runs untrusted native code:
+
+- `cargo build-sbf` on the **user-supplied program** compiles it natively — running its
+  `build.rs`, its proc-macros, and (for a future Prover/CVLR backend) LLM-munged source.
+- `crucible run` compiles the **LLM-authored harness** (its `setup()`, `action_*`, `build.rs`)
+  and then runs it as a native LiteSVM-in-process binary (§7.2 — verified native, no SVM sandbox).
+
+So arbitrary code of the LLM's (and the analyzed program's) choosing executes with whatever
+ambient authority the AutoProver process has: `ANTHROPIC_API_KEY`, `CERTORA*` cloud tokens,
+`AWS_*`, `PG*`, the network route to `169.254.169.254`, and the entire bind-mounted host project.
```
shellygr: We should drop these explanations. Any implementor could be concerned about what LLMs could do, so the specifics of our setup are irrelevant

docs/command-sandbox.md
```diff
@@ -0,0 +1,512 @@
...
+Until this phase is green the backend may run only in a trusted, offline environment on trusted
+input (the gate scenario). This is the definition of done.
+
+---
+
+## 2. Threat model
+
+| | |
+|---|---|
+| **Asset** | AutoProver's ambient secrets (LLM/cloud API keys, DB creds), its network egress (incl. `169.254.169.254` metadata → IAM role creds on EC2), and host files outside the command's declared inputs. |
```
shellygr: ditto

docs/command-sandbox.md
```diff
@@ -0,0 +1,512 @@
...
+it assumes them.
+
+---
+
+## 5. Offline dependency resolution — split fetch (network, no exec) from build (exec, no network)
+
+The tension: `cargo build` needs its dependency crates, but the sandbox has no network. Resolution
+splits cleanly along the code-execution line:
+
+- **`cargo fetch` / `cargo vendor` download but never run build scripts** — no untrusted code
```
shellygr, line 163: in Solidity AP we already have a split between the 'dependency fetch' task and actual building, using a separate builder container that is not in this repo. does it make sense to move cargo fetch steps into the builder repo as well? let's sync over DM

scripts/Dockerfile
```diff
@@ -85,6 +95,8 @@ RUN mkdir -p $HOME/.cache/huggingface \
 # postgresql-client provides `psql` for the entrypoint's setup-db; git is
 # occasionally shelled out to by tooling. No openssh-client / SSH — the
 # Python install pulls from PyPI + the vendored graphcore submodule.
+# (The `RunCommand` sandbox uses in-kernel Landlock + seccomp — no package
```
shellygr: comment not useful here

scripts/Dockerfile
```diff
@@ -59,6 +59,16 @@ RUN set -eux; \
     ( cd "$target" && /venv/bin/sphinx-build -M singlehtml . tmp ); \
     cp "$target/tmp/singlehtml/index.html" /out/cvl.html
 
+# ---- Stage 1b: build the `run-confined` command-sandbox launcher ----
```
shellygr: do we want to use a docker-compose overlay here so that we can avoid folding rust into an image that is solidity-focused?

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    let status = created.restrict_self().map_err(|e| e.to_string())?;
+    if matches!(status.ruleset, RulesetStatus::NotEnforced) {
+        return Err("kernel did not enforce Landlock (need Linux >= 5.13)".to_string());
+    }
+    Ok(())
+}
+
+fn apply_seccomp(cfg: &Config) -> Result<(), String> {
```
ericeil, line 315: **[bug / critical] io_uring fully bypasses the seccomp network deny-list**

`apply_seccomp` uses a **default-allow** filter that only matches `socket(AF_INET|AF_INET6)`, `ptrace`, and `process_vm_*`. It does **not** deny `io_uring_setup` / `io_uring_enter` / `io_uring_register`.

On this kernel (7.0), a sandboxed process can:

1. `io_uring_setup` (allowed)
2. submit `IORING_OP_SOCKET` with `AF_INET` → returns a live TCP socket FD
3. submit `IORING_OP_CONNECT` to `1.1.1.1:80` → **connect succeeds**
4. same path to `169.254.169.254:80` → **IMDS TCP connect succeeds**

Direct `socket(AF_INET)` correctly returns EPERM; the escape test only exercises that path (`TcpStream::connect`), so the gate is green while the network guarantee is false.

This violates `docs/command-sandbox.md` guarantee #1 (no network / no IMDS) and the asset table (IAM role creds via IMDS).

**Suggestion:** Deny at least `io_uring_setup` (and preferably `io_uring_enter` / `io_uring_register`) under `network=false`. Cargo/rustc do not need io_uring for offline builds. Also extend the escape suite with an io_uring socket+connect probe (and IMDS), not only `socket(2)`.

Defense-in-depth: consider also installing Landlock net rules (`AccessNet::ConnectTcp|BindTcp`, ABI ≥4) so even a future seccomp miss cannot open TCP — noting Landlock net is TCP-only, so seccomp remains required for UDP/DNS.

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+        libc::SYS_ptrace,
+        libc::SYS_process_vm_readv,
+        libc::SYS_process_vm_writev,
+    ] {
+        rules.insert(nr as i64, Vec::new());
+    }
+
+    let filter = SeccompFilter::new(
+        rules,
+        SeccompAction::Allow,                     // default: allow syscalls we didn't name
```
ericeil, line 376: **[bug / high] Default-allow seccomp is too thin for the threat model**

```rust
SeccompAction::Allow,  // default: allow syscalls we didn't name
SeccompAction::Errno(...), // named + matched: deny
```

Beyond io_uring, live probes under the real launcher also allowed:

- **`socket(AF_NETLINK)`** — netlink is a network-adjacent control plane (routing/neighbor info; more with caps).
- **`socket(AF_VSOCK)`** — guest↔host channel on many VM/cloud setups; not covered by the “inet only” story.
- **`socket(AF_UNIX)`** — intentional for jobserver, but see abstract-namespace finding.

A hostile `build.rs` only needs *one* missed vector. Prefer either:

1. **Default-deny allowlist** of syscalls actually required by cargo/rustc/linker/fuzz harness (large maintenance cost), or
2. A **broader deny-list** that at minimum covers: all `io_uring_*`, `socket` for every domain except `AF_UNIX` (match `arg0 != AF_UNIX` rather than only denying INET/INET6), `bpf`, `perf_event_open`, `userfaultfd`, `kexec_*`, `init_module`, etc.

Right now the filter documents “deny inet sockets” but implements “allow everything else,” which is a much weaker property than the design doc’s “no network at all.”

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+fn set_no_new_privs() {
+    // Required before loading a seccomp filter (and by Landlock) for an unprivileged
+    // process; ensures no exec can regain privileges.
+    unsafe { libc::prctl(libc::PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) };
+}
+
+fn apply_landlock(cfg: &Config) -> Result<(), String> {
```
ericeil, line 260: **[bug / high] Missing Landlock scopes: signals + abstract Unix sockets (same-uid control plane)**

`apply_landlock` only handles `AccessFs` (ABI V5). It never installs **Landlock scopes** available since ABI V6:

- `Scope::Signal` — block signaling processes outside the sandbox
- `Scope::AbstractUnixSocket` — block connect to abstract UDS created outside the sandbox

Live probe under `run-confined` (same policy shape as production):

- `kill(SIGSTOP)` / `kill(SIGKILL)` on an arbitrary same-uid victim PID → **success**
- `connect` to abstract socket `\0autoprover-sb-test-socket` owned by a listener outside the sandbox → **received data**

Doc §6 claims the only same-uid secret vectors are `/proc/<parent>/environ` and `ptrace`/`process_vm_*`. That understates the model: the child shares AutoProver’s uid, so without scopes it can **SIGKILL the AutoProver parent**, stop other jobs, or speak to any abstract-namespace service on the same uid (language servers, agent helpers, docker credentials helpers, etc.).

On this host Landlock ABI is high enough (kernel 7.0). Use best-effort:

```rust
.scope(Scope::from_all(ABI::V6))  // BestEffort drops scopes on older kernels
```

and document residual risk on AMI kernels &lt; 6.7 (ABI &lt; 6). Add escape probes for `kill(ppid, SIGKILL)` and abstract UDS connect.

composer/sandbox/recipes.py
```diff
@@ -0,0 +1,158 @@
...
+
+    With ``offline`` (the default — the sandbox has no network, §5), ``CARGO_NET_OFFLINE=1``
+    is set in the child env. That one var forces *every* cargo invocation offline,
+    including the nested ``cargo`` that ``crucible run`` spawns to build the harness —
+    so the deps must already be warm in ``CARGO_HOME`` (see :func:`warm_cargo_cache`,
+    run *outside* the sandbox first).
+    """
+    home = Path.home()
+    rustup = Path(os.environ.get("RUSTUP_HOME", home / ".rustup"))
+    cargo = Path(os.environ.get("CARGO_HOME", home / ".cargo"))
```
ericeil, line 123: **[bug / medium] Entire shared `~/.cargo` is granted RO — leaks registry tokens and other files**

```python
cargo = Path(os.environ.get("CARGO_HOME", home / ".cargo"))
ro_candidates += [rustup, cargo, ...]
```

The private per-run `CARGO_HOME` correctly stops **write** poisoning, but the shared home stays on the RO list so `bin/cargo` remains visible. Landlock PathBeneath on `~/.cargo` makes **everything under it** readable, including:

- `~/.cargo/credentials.toml` / `credentials` (crates.io / private registry tokens)
- any other files the user stored under `.cargo`

Live probe: granting a fake cargo home containing `token = "crates-io-TOKEN-CANARY"` → sandboxed process read it back.

This is a secrets violation (guarantee #2 / asset table) whenever developers or CI store cargo credentials in the default location.

**Suggestion:** grant only the needed subtrees read-only, e.g. `~/.cargo/bin`, and (if required for offline) a **content-addressed registry cache** that cannot hold credentials — not the cargo home root. Alternatively strip/deny `credentials.toml` by not granting the parent of that file (grant `bin/` + `registry/cache/` + `registry/index/` separately). Extend the escape suite with a planted `credentials.toml` canary under the granted cargo path.

tests/test_sandbox_escape.py
```diff
@@ -0,0 +1,143 @@
...
+_MALICIOUS_RS = """
+use std::fs;
+use std::net::{SocketAddr, TcpStream};
+use std::time::Duration;
+
+fn probe(name: &str, result: &str) {
+    let _ = fs::write(format!("probe_{}.txt", name), result);
+}
+
+fn net(addr: &str) -> String {
```
ericeil, line 94: **[suggestion] Escape suite does not cover the bypasses that actually work**

The malicious program only checks:

- env var presence
- `/proc/<ppid>/environ`
- host file open
- `TcpStream::connect` (uses `socket(2)` + `connect(2)`)

It misses:

1. **io_uring** socket/connect (confirmed network + IMDS bypass)
2. **Abstract Unix** connect to a same-uid outside listener
3. **`kill(ppid, SIGKILL)`** / signal scope
4. **Cargo credentials** file under the RO-granted cargo home
5. **AF_NETLINK / AF_VSOCK**

Because the gate is the definition of done for Phase 6, green CI currently overclaims the threat-model guarantees. Please add probes for (1)–(4) at minimum before calling the boundary proven.

docs/command-sandbox.md
```diff
@@ -0,0 +1,512 @@
...
+  (`ptrace`, `process_vm_readv`/`writev`).
+- **env allowlist** — the launcher `execve`s with a scrubbed environment (PATH, HOME, CARGO_HOME,
+  RUSTUP_HOME, TERM, and benign build vars only). The `--clearenv` equivalent, done in-process.
+- **rlimits** — `setrlimit` for `RLIMIT_AS` / `RLIMIT_CPU` / `RLIMIT_NPROC` / `RLIMIT_FSIZE` (§7).
+
+Landlock and seccomp are **preserved across `execve`** (with `NO_NEW_PRIVS`) and **inherited across
+`fork`**, so the launcher applies them once and every descendant — `cargo`, `rustc`, each `build.rs`,
+the linker, the fuzz binary — runs confined.
+
+### The same-uid caveat, and why it is closed
```
ericeil, line 262: **[suggestion] Same-uid caveat section is incomplete**

> These are the only same-uid vectors to AutoProver's memory/env; both verified closed

Even after fixing io_uring, same-uid residual vectors that should be named here:

- **Signals** (SIGKILL/STOP the parent) — fixed by Landlock `Scope::Signal` (ABI ≥6)
- **Abstract Unix sockets** — fixed by `Scope::AbstractUnixSocket`
- **SCM_RIGHTS / AF_UNIX** to cooperatively malicious or overly trusting same-uid services
- **Readable grants that contain secrets** (full `~/.cargo`, full `/etc`)

Recommend updating §2 guarantees / §6 / §11 so operators know what is proven vs residual, and so the escape gate stays aligned with the prose.

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+    if let Some(v) = cfg.rlimit_nproc {
+        set(libc::RLIMIT_NPROC, v);
+    }
+    if let Some(v) = cfg.rlimit_fsize {
+        set(libc::RLIMIT_FSIZE, v);
+    }
+}
+
+fn set_no_new_privs() {
```
ericeil, line 245: **[nit] `PR_SET_NO_NEW_PRIVS` return value ignored**

```rust
unsafe { libc::prctl(libc::PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0) };
```

If this fails, seccomp filter install may fail (fail-closed) or behave oddly depending on kernel. Check the return value and `die(EXIT_SANDBOX_UNAVAILABLE, ...)` on failure for consistency with Landlock/seccomp error handling.

docs/command-sandbox.md
ericeil: Oh, that's interesting.  Yes, if we can build the user's code separately in a more locked-down container, that would be ideal.  I think we will still need the sandbox (both in the "builder" container, and later when we build the LLM-generated code), but some more defense-in-depth would be great here.  Will track this as a separate workitem for now.

ericeil (commented):
## Security audit summary

Focused review of PR #73 against the guarantees in `docs/command-sandbox.md` §2:

1. **No network** (incl. DNS / IMDS)
2. **No secrets** (env scrub + no out-of-band recovery)
3. **Minimal filesystem**
4. **Resource caps**
5. **Offline builds**

I probed a live `run-confined` binary on this host (kernel 7.0, Landlock enforcing). Several **intended** vectors are correctly closed (env scrub, Landlock outside-workdir, `/proc/<ppid>/environ`, `ptrace`/`process_vm_*`, hardlink-out, direct `socket(AF_INET)`).

However, **guarantee #1 is fully breakable today**, and several same-uid control-plane gaps remain. Highest severity first.

### Issue counts by severity

- bugs (security): 4
- suggestions: 2
- nits: 1

### Verified with live probes

| Vector | Result under sandbox |
|---|---|
| `socket(AF_INET)` | DENIED (EPERM) — as designed |
| **io_uring `IORING_OP_SOCKET` + `CONNECT` to 1.1.1.1:80** | **SUCCESS — full network bypass** |
| **io_uring connect to 169.254.169.254:80** | **SUCCESS (TCP connect)** |
| Abstract Unix socket to same-uid listener outside sandbox | SUCCESS |
| `kill(SIGKILL)` any same-uid process | SUCCESS |
| Read planted file under RO-granted cargo home (`credentials.toml`) | SUCCESS |
| AF_NETLINK / AF_VSOCK `socket()` | SUCCESS |
| Landlock outside file via normal open / io_uring OPENAT | DENIED |
| Env canary / ptrace / process_vm_readv | DENIED |

**Bottom line:** untrusted code inside the sandbox can reach the network (including IMDS) today via io_uring, despite the seccomp deny-list on `socket()`. That is a direct violation of the stated threat model.

scripts/Dockerfile
ericeil: Yes

ericeil:
Did security reviews and testing with both Grok and Opus, and fixed the problems they found.  Added a CI test to verify the sandbox against Amazon Linux 2023.

composer/sandbox/command.py
```diff
@@ -0,0 +1,156 @@
...
+    prov: SandboxProvider = provider if provider is not None else NoneProvider()
+    ensure_available(prov)  # fail-closed: raises before running if it can't confine
+    spec = prov.wrap(policy if policy is not None else SandboxPolicy(), program, list(args))
+    child_env = dict(spec.env) if spec.env is not None else None
+    if env_overlay:
+        # Overlay onto the effective env (the inherited parent env when the provider
+        # didn't set one — i.e. the `none`/unsandboxed path).
+        child_env = {**(child_env if child_env is not None else os.environ), **env_overlay}
+
+    async def _run() -> CommandResult:
```
chandrakananandi, line 131: do we know why we need async here?

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,347 @@
...
+            .cloned()
+            .ok_or_else(|| format!("{flag} requires a value"))
+    };
+    let parse_u64 = |s: &str, flag: &str| -> Result<u64, String> {
+        s.parse::<u64>().map_err(|_| format!("{flag} expects an integer, got {s:?}"))
+    };
+
+    while i < argv.len() {
+        match argv[i].as_str() {
+            "--rw" => cfg.rw_paths.push(PathBuf::from(take(&mut i, "--rw")?)),
```
chandrakananandi: why not use Clap: https://rust-cli.github.io/book/tutorial/cli-args.html#parsing-cli-arguments-with-clap? That's usually the way to do CLI arg parsing in rust. Maybe ask claude about it?

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,347 @@
...
+                return Ok(cfg);
+            }
+            other => return Err(format!("unknown flag {other:?} (did you forget `--` before the command?)")),
+        }
+        i += 1;
+    }
+    Err("missing `--` and command to run".to_string())
+}
+
+fn set_rlimits(cfg: &Config) {
```
chandrakananandi, line 225: do we know what these limits are needed for?

chandrakananandi (commented):
Had a couple of small questions.

composer/sandbox/command.py
jtoman: yes, because we are using async APIs within the body of `_run`; to wit, we're `await`ing  the asyncio.create_subprocess_exec below. The alternative would be to use the sync apis, but that would hang the event loop and starve llms while we wait for this command to execute.

rust/run-confined/src/main.rs
jtoman: Chandra, would you say there you're...

...clapping back?

composer/sandbox/command.py
chandrakananandi: oh i see i see.

composer/sandbox/command.py
```diff
@@ -0,0 +1,156 @@
...
+    stdout: str
+    stderr: str
+
+    def as_observation(self) -> dict:
+        """The ``Observation::CommandResult`` payload the IoC loop feeds back to Rust."""
+        return {
+            "exit_code": self.exit_code,
+            "stdout": self.stdout,
+            "stderr": self.stderr,
+        }
```
jtoman: typed dict imo

composer/sandbox/command.py
```diff
@@ -0,0 +1,156 @@
...
+        raise UnsafePath(
+            f"file path {rel!r} is absolute or traverses outside the workdir"
+        )
+    target = workdir / p
+    # Belt-and-suspenders: the resolved path must still live under the workdir.
+    try:
+        target.resolve().relative_to(workdir.resolve())
+    except ValueError as e:
+        raise UnsafePath(f"file path {rel!r} resolves outside the workdir") from e
+    return target
```
jtoman: it seems like this check is sufficient andyou can skip the p.is_absolute / `..` in p.parts check above?

composer/sandbox/command.py
```diff
@@ -0,0 +1,156 @@
...
+    sem: asyncio.Semaphore | None = None,
+    provider: SandboxProvider | None = None,
+    policy: SandboxPolicy | None = None,
+    env_overlay: dict[str, str] | None = None,
+) -> CommandResult:
+    """Write ``files`` into ``workdir``, then run ``program args`` there and capture output.
+
+    ``workdir`` persists across calls (a session materializes its crate once and
+    runs several commands against it). Concurrency is bounded by ``sem`` when
+    given — important because fuzzers are resource-hungry.
```
jtoman: claude loves to narrate these pointless "why" and "what we didn't do" details. No one needs this justification.

composer/sandbox/command.py
```diff
@@ -0,0 +1,156 @@
...
+    — as one unit under ``sem`` when given, so that when callers share a workdir the
+    file-write and the run don't interleave (a concurrent caller can't overwrite these
+    files between our write and our run). Path confinement complements the sandbox.
+
+    ``env_overlay`` sets extra env vars on the child on top of what it would otherwise
+    inherit — used by the *unsandboxed* prep steps (e.g. `cargo fetch` with a per-run
+    ``CARGO_HOME``); the sandboxed path's env is fully governed by the provider/policy.
+    """
+    prov: SandboxProvider = provider if provider is not None else NoneProvider()
+    ensure_available(prov)  # fail-closed: raises before running if it can't confine
```
jtoman: claude loves to keep mentioning this `fail-closed` term, I have never heard of it.

composer/sandbox/command.py
```diff
@@ -0,0 +1,156 @@
...
+            return CommandResult(-1, "", f"command timed out after {timeout_s}s")
+        rc = proc.returncode if proc.returncode is not None else -1
+        return CommandResult(
+            rc, out_b.decode(errors="replace"), err_b.decode(errors="replace")
+        )
+
+    if sem is not None:
+        async with sem:
+            return await _run()
+    return await _run()
```
jtoman: more compact
```
async with (sem if sem is not None else contextlib.nullcontext)):
   return await _run()
```

or, if you prefer to abuse `or` syntax:
```
async with (sem or contextlib.nullcontext):
   ...
```

but I personally hate abusing `or` like this

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @classmethod
+    def from_env(cls, **overrides) -> "SandboxConfig":
+        """Read the provider from ``$COMPOSER_SANDBOX_PROVIDER`` (default ``none``);
+        remaining fields come from ``overrides`` (e.g. a backend's ``extra_ro``)."""
+        return cls(provider=os.environ.get(_ENV_VAR, "none"), **overrides)
+
+    @property
+    def enabled(self) -> bool:
+        return self.provider != "none"
```
jtoman, line 81: you literally have `str | None` as a possible type, why u do this.

docs/command-sandbox.md
```diff
@@ -0,0 +1,558 @@
...
+
+**The mechanism sits behind a `SandboxProvider` seam, so it is swappable.** `run_local_command`
+never names a concrete tool. It holds a **tool-agnostic `SandboxPolicy`** (the *intent*: rw paths,
+ro paths, env allowlist, rlimits, network-off — §7) and a `SandboxProvider` that translates that
+intent into a concrete launch:
+
+```python
+class SandboxProvider(Protocol):
+    def wrap(self, policy: SandboxPolicy, program: str, args: list[str]) -> LaunchSpec: ...
+    def available(self) -> Availability: ...   # drives fail-closed (§7)
```
jtoman, line 132: fun drinking game: take a drink whenever claude writes "fail-closed". Can LLMs enjoy stims? 🤔

docs/command-sandbox.md
```diff
@@ -0,0 +1,558 @@
...
+  it happens **inside** the sandbox, `--offline`, against the already-warm cache.
+
+The harness `Cargo.toml` is **host-owned** (`CrucibleDep.render_deps`, pinned versions, §6.1), so
+its dep graph is fixed and vendorable deterministically. The program-under-test's `Cargo.toml` is
+user-supplied, but `cargo fetch` on it is still exec-free, so the same split holds for the build-sbf
+step. This also closes the build-time supply-chain vector: with offline + a pre-warmed cache, a
+malicious `build.rs` cannot pull a payload at build time.
+
+**Implementation (step 4).** "Offline inside" is one env var, not per-tool flags: the policy sets
+**`CARGO_NET_OFFLINE=1`** in the child env, which forces *every* cargo invocation offline — including
```
jtoman, line 177: I assume it is possible for malicious code to set this environment var back to 0 or unset it at some point. What is the implication of that? The user manages make cargo panic about not being able to talk to servers?

docs/command-sandbox.md
```diff
@@ -0,0 +1,558 @@
...
+- **seccomp-BPF self-filter** (`PR_SET_NO_NEW_PRIVS` + `SECCOMP_SET_MODE_FILTER`) — installing a
+  *stricter* filter on yourself is unprivileged and permitted by Docker's default profile. It denies
+  `socket` for every domain **except `AF_UNIX`** (so TCP, UDP/DNS, IMDS, netlink, vsock, … are
+  blocked while cargo's jobserver still works), denies **`io_uring_*`** (the classic way to create
+  sockets without calling `socket(2)`), and denies the remaining same-uid secret vectors
+  (`ptrace`, `process_vm_readv`/`writev`). On **x86_64** each deny is **mirrored onto its x32-ABI
+  syscall number** (`nr | 0x4000_0000`): the x32 calling convention runs under the same
+  `AUDIT_ARCH_X86_64` identity, so seccompiler's arch guard passes it through, and without the mirror
+  an x32-tagged `socket`/`io_uring`/`ptrace` would miss the exact-number rules and reach
+  default-allow — a full bypass (which libseccomp guards against automatically; seccompiler does not).
```
jtoman, line 251: do you understand this? I sure don't and I'm kind of a unix guy

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+    extra_rw: tuple[Path, ...] = ()
+    env_passthrough: tuple[str, ...] = DEFAULT_ENV_PASSTHROUGH
+    offline: bool = True  # sandbox has no network → force cargo offline (§5)
+    mem_bytes: int | None = None
+    cpu_seconds: int | None = None
+    nproc: int | None = None
+    fsize_bytes: int | None = None
+
+    @classmethod
+    def from_env(cls, **overrides) -> "SandboxConfig":
```
jtoman: This is pretty bleak! It is a PITA, but you should do:

```
class SandboxArgs(TypedDict):
   env_passthrough: NotRequired[tuple[str, ...]]
   offline: NotRequired[bool]
   # etc
```

and then have:
```
@classmethod
def from_env(cls, **overrides: Unpack[SandboxArgs]):
   ...
```

this keeps your signatures strongly typed, so you don't have to worry about some idiot going `SandboxConfig.from_env(lol="get wrekt")` and crashing at runtime.

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+    def enabled(self) -> bool:
+        return self.provider != "none"
+
+    def resolve_provider(self) -> SandboxProvider:
+        # Importing the launcher module registers the "launcher" provider; the seam
+        # itself never imports a concrete mechanism (docs/command-sandbox.md §6).
+        if self.provider == "launcher":
+            import composer.sandbox.launcher  # noqa: F401
+        return get_provider(self.provider)
+
```
jtoman: have you heard of our lord and savior importlib entrypoints? You should use those. this is literally what it was built for.

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+    def build_policy(self, workdir: str | Path) -> SandboxPolicy:
+        """The concrete policy for a command running in ``workdir``. The ``none``
+        provider ignores the policy, so a bare :class:`SandboxPolicy` suffices there."""
+        if not self.enabled:
+            return SandboxPolicy()
```
jtoman: oh I hate this. Returning a singleton that we know is garbage but we cross our fingers and hope it ignores it suuuuucks. There's gotta be a less roundabout way to accomplisht his.

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+            "allow_env": [f"{k}={v}" for k, v in policy.env_allowlist.items()],
+            "network": policy.network,
+            "rlimits": {
+                "mem_bytes": policy.mem_bytes,
+                "cpu_seconds": policy.cpu_seconds,
+                "nproc": policy.nproc,
+                "fsize_bytes": policy.fsize_bytes,
+            },
+            "timeout_s": timeout_s,
+        }
```
jtoman: typeddict please and thank you.

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+        For a real provider this resolves the ``run-confined`` path and is **fail-closed**
+        (``ensure_available`` raises if the launcher can't confine here). The ``none`` provider
+        yields ``run_confined=None`` — the backend runs the command directly (trusted input)."""
+        if not self.enabled:
+            return {"run_confined": None, "timeout_s": timeout_s}
+        provider = self.resolve_provider()
+        ensure_available(provider)  # fail-closed: raise before any untrusted code runs
+        policy = self.build_policy(workdir)
+        return {
+            "run_confined": getattr(provider, "binary", None),
```
jtoman: WAT

composer/sandbox/config.py
```diff
@@ -0,0 +1,98 @@
...
+            "ro": [str(p) for p in policy.ro_paths],
+            "rw": [str(p) for p in policy.rw_paths],
+            "allow_env": [f"{k}={v}" for k, v in policy.env_allowlist.items()],
```
jtoman: but like ... dicts exist? why are you doing this

composer/sandbox/config.py
jtoman: this a wild thing to do when provider is a protocol? Nothing even hints an implementer should add this attribute?

composer/sandbox/launcher.py
```diff
@@ -0,0 +1,114 @@
...
+    override = os.environ.get("RUN_CONFINED_BIN")
+    if override and Path(override).is_file():
+        return override
+    on_path = shutil.which(_BIN_NAME)
+    if on_path:
+        return on_path
+    # Dev fallback: composer/sandbox/launcher.py → repo root is parents[2].
+    repo_root = Path(__file__).resolve().parents[2]
+    cand = repo_root / "rust" / "target" / "release" / _BIN_NAME
+    return str(cand) if cand.is_file() else None
```
jtoman, line 44: I assume there is a way to include this binary in the python wheel and add the cargo build steps in pyproject. In which case you can use `importlib` to resolve that resource automatically instead of all this.

composer/sandbox/policy.py
```diff
@@ -0,0 +1,165 @@
...
+@runtime_checkable
+class SandboxProvider(Protocol):
+    """Maps a :class:`SandboxPolicy` + a command to a concrete :class:`LaunchSpec`.
+
+    The one seam every sandbox mechanism implements. Implementations are pure with
+    respect to :meth:`wrap` (argv construction only — no subprocess), so they are
+    trivially unit-testable; the actual confinement happens in the launched process.
+    """
+
+    name: str
```
jtoman: you almost certainly want this to be a `@property`. Also; why are you having this runtime checkable? You don't seem to inherit it in your only implementer.

composer/sandbox/policy.py
```diff
@@ -0,0 +1,165 @@
...
+    """Construct the provider registered under ``name``. Raises ``ValueError`` for an
+    unknown name (a config error, distinct from a provider being *unavailable*)."""
+    try:
+        factory = _PROVIDERS[name]
+    except KeyError:
+        raise ValueError(
+            f"unknown sandbox provider {name!r}; known: {sorted(_PROVIDERS)}"
+        ) from None
+    return factory()
+
```
jtoman: all of this disappears with entrypoints (I'm in the pocket of big entry point).

composer/sandbox/launcher.py
```diff
@@ -0,0 +1,114 @@
...
+                    f"(cargo build -p run-confined --release) or set RUN_CONFINED_BIN"
+                ),
+            )
+        try:
+            proc = subprocess.run(
+                [self._binary, "--probe"],
+                capture_output=True,
+                text=True,
+                timeout=_PROBE_TIMEOUT_S,
+            )
```
jtoman: probably worth making this async-able. yes its probably fast and won't block the event loop for long, but still.

composer/sandbox/policy.py
```diff
@@ -0,0 +1,165 @@
...
+    env: Mapping[str, str] | None = None
+
+
+@dataclass(frozen=True)
+class Availability:
+    """Result of :meth:`SandboxProvider.available` — whether the provider can
+    actually confine here (e.g. the launcher probes the kernel's Landlock ABI)."""
+
+    ok: bool
+    reason: str = ""
```
jtoman: this lets you expression nonsense like `Availability(ok=True, reason="everything is broken")`. Python typing does allow sum types, so you could go `type Availability = Literal["ok"] | Reason` where:

```
@dataclass
class Reason
   reason: str
```

or sth

composer/sandbox/recipes.py
```diff
@@ -0,0 +1,182 @@
...
+    shared ``registry/`` and ``git/`` trees are not required. A future shared
+    read-only cache optimization can add specific cache subtrees here without
+    re-opening the credentials file.
+    """
+    root = Path(cargo_home)
+    out: list[Path] = []
+    bin_dir = root / "bin"
+    if bin_dir.is_dir():
+        out.append(bin_dir)
+    return tuple(out)
```
jtoman: uh?

composer/sandbox/recipes.py
```diff
@@ -0,0 +1,182 @@
...
+    ro_candidates: list[Path] = [Path(p) for p in _SYSTEM_RO]
+    ro_candidates += [
+        rustup,
+        # Shared cargo: bin/ only — never the home root (credentials.toml).
+        *shared_cargo_ro_paths(cargo),
+        # cargo-build-sbf's downloaded sBPF platform-tools (layout varies by version).
+        home / ".cache" / "solana",
+        home / ".local" / "share" / "solana",
+    ]
+    ro_candidates += list(extra_ro)
```
jtoman: probably want `extend`

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,347 @@
...
+    rlimit_nproc: Option<u64>,
+    rlimit_fsize: Option<u64>,
+    program: String,
+    args: Vec<String>,
+}
+
+fn die(code: i32, msg: &str) -> ! {
+    eprintln!("run-confined: {msg}");
+    std::process::exit(code);
+}
```
jtoman, line 154: been there

rust/run-confined/src/main.rs
```diff
@@ -0,0 +1,347 @@
...
+    // Net TCP deny: ABI V4 — BestEffort; defense-in-depth next to seccomp.
+    let abi_fs = ABI::V5;
+
+    let mut ruleset = Ruleset::default()
+        .set_compatibility(CompatLevel::BestEffort)
+        .handle_access(AccessFs::from_all(abi_fs))
+        .map_err(|e| e.to_string())?
+        .scope(Scope::from_all(ABI::V6))
+        .map_err(|e| e.to_string())?;
+
```
jtoman, line 277: average rust code amirite

composer/sandbox/config.py
jtoman: I realize none is the literal id of a backend or provider so ignore me.

rust/run-confined/src/main.rs
ericeil: Yep, that's probably a better choice

composer/sandbox/command.py
ericeil: yep

composer/sandbox/command.py
ericeil: Heh, yeah, I wondered about that.  Says Claude: 

> "Fail-closed" is a security design principle: when the safety mechanism can't do its job, the system refuses to run rather than running unprotected.
>
> Here at command.py:122, ensure_available(prov) checks whether the selected sandbox provider can actually confine the child process (e.g. for the launcher provider: is run-confined present, does the kernel support Landlock + seccomp, etc.). If it can't, ensure_available raises SandboxUnavailable before the command is ever spawned. So the command never runs unsandboxed.
>
> The contrast is fail-open: if confinement were unavailable, quietly fall back to running the command with no sandbox. That's the dangerous default — you'd get a working command and silently lose the security boundary, which is exactly what you don't want for something running potentially-untrusted toolchains.

composer/sandbox/command.py
ericeil: cool

composer/sandbox/config.py
ericeil: Yep, that works quite nicely

composer/sandbox/config.py
ericeil: This is wonky indeed. This turns out to be here to support some Rust code that's not in this PR, and which wasn't ever fully decoupled from the specific run-confined provider. I'll work out a better boundary.

composer/sandbox/config.py
ericeil: Removing this

composer/sandbox/launcher.py
ericeil: Apparently it's possible, but currently the wheel is pure python and thus platform-independent.  The hope was to keep the Solana code testable on non-Linux (Mac) platforms, for which we currently have no sandbox facility (and if we add a Mac sandbox, it will be a different implementation).

It's possible that there's some way to work that all out, but this seems simple enough to me as-is.

composer/sandbox/config.py
ericeil: (Probably this will change again in a future PR)

composer/sandbox/recipes.py
ericeil: Will simplify this

docs/command-sandbox.md
ericeil: Yeah, Cargo will just fail (I've seen this case on accident :) )

docs/command-sandbox.md
ericeil: I take it to mean that untrusted code could spoof 32-bit syscalls, so we have to block those also.  But I'm not familiar at all with the details of Linux syscalls.

rust/run-confined/src/main.rs
ericeil: Their meanings are described in comments on the new Clap-based CLI options.  These are here so we can e.g. limit the memory usage of the Crucible process so it doesn't interfere with the main service.  I suspect we'll want more control over this some day, but these particular limits are easy to implement. :)

chandrakananandi (approved):
Thanks for getting this started!

pyproject.toml
```diff
@@ -146,6 +146,13 @@ certora-fixconf = "certora_autosetup.fixconf:main"
 autosetup = "certora_autosetup.autosetup.cli:main"
 fixconf = "certora_autosetup.fixconf:main"
 
+# Command-sandbox providers, discovered by composer.sandbox.policy.get_provider.
```
jtoman: at the risk of asking a stupid question: ... where is this function? I don't see `get_provider` in `composer.sandbox.policy` and I scanned all the files in `composer.module` and didn't see any entry point magic. Did you forget to patch that in perhaps?

jtoman (commented):
All the other python changes look quite cromulent. One glaring hole, which could be attributable to me failing to read...

pyproject.toml
jtoman: Dear viewer: I cannot, in fact, read

jtoman (approved):

pyproject.toml
ericeil: I updated the comment to point to the actual function that does this

PR #74, opened 2026-07-15
autoprove/foundry: fail the run when every component fails to generate

naftali-g:
Add `CorePipelineResult.all_failed` (true when no component produced a deliverable — all gave up or crashed) and `n_delivered`. The console autoprove/foundry entry points now return a non-zero exit code in that case.

### DIFF
--- a/composer/pipeline/ptypes.py
+++ b/composer/pipeline/ptypes.py
@@ -112,9 +112,8 @@ class CorePipelineResult[FormT: BackendResult]:
 
     @property
     def all_failed(self) -> bool:
-        """Every attempted component failed to generate or gave up — nothing was delivered.
-        The run is a total failure and its entry point returns a non-zero exit code. Guarded
-        on a non-empty outcome set so "all of nothing" is never reported as failure (the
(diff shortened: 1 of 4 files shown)


composer/pipeline/ptypes.py
```diff
@@ -104,6 +104,20 @@ class CorePipelineResult[FormT: BackendResult]:
...
+    @property
+    def n_delivered(self) -> int:
+        """Components that produced a deliverable — a successful formalization. Everything
+        else (a ``GaveUp`` or a crash) is a component that failed to generate."""
+        return sum(1 for o in self.outcomes if isinstance(o.result, Delivered))
+
+    @property
+    def all_failed(self) -> bool:
+        """Every attempted component failed to generate or gave up — nothing was delivered.
+        The run is a total failure and its entry point returns a non-zero exit code. Guarded
```
jar-ben: `and its entry point returns a non-zero exit code` - I would not talk about the entrypoint here, because the code is not here. Just describe what `all_failed` does.

jar-ben (approved):
nit comment change

PR #75, opened 2026-07-15
Stream .asts.json to fix autosetup OOM (SIGKILL) on larger trees

jar-ben:
Autosetup `json.load`ed the whole `.asts.json` (~2.5GB on a mid-size project) at four sites on the compilation-analysis path, each ballooning to 8-12GB of Python objects and getting OOM-killed.

New `stream_ast_files()` (`utils/file_utils.py`) streams the file one source file at a time via `ijson`; all four loaders (`generate_ast_graph`, `_build_declared_contracts_by_file`, `_extract_inheritance_and_abstract_from_ast`,
`auto_munges.detect_and_apply_code_access_patches`) route through it. Adds `ijson`.


certora_autosetup/setup/setup_prover.py
```diff
@@ -1140,46 +1138,50 @@ def _extract_inheritance_and_abstract_from_ast(self, ast_file_path: Optional[Pat
...
+            for _file_path, abs_path_dict in stream_ast_files(ast_file_path):
+                for _abs_path, nodes in abs_path_dict.items():
+                    for _node_id, node in nodes.items():
+                        if isinstance(node, dict) and node.get("nodeType") == "ContractDefinition":
+                            contract_defs.append({
+                                "id": node.get("id"),
+                                "name": node.get("name"),
+                                "abstract": node.get("abstract", False),
+                                "contractKind": node.get("contractKind", "contract"),
+                                "linearizedBaseContracts": node.get("linearizedBaseContracts", []),
```
naftali-g: do we want to make this a dataclass?

shellygr: if we had a pydantic model of the solc ast a lot of our work would be a lot easier... I wonder if such exists

shellygr: there might be one for typescript https://github.com/OpenZeppelin/solidity-ast

jar-ben: Perhaps I leave that for a follow-up PR? and in this one I add a dataclass for this local change

naftali-g: sounds fine to me

shellygr (approved):

PR #76, opened 2026-07-15
Typed pydantic models of the Solidity compact AST (--dump_asts) + autosetup migration

shellygr:
## What

A new subpackage `certora_autosetup/solidity_ast/`: a **full, hand-authored pydantic v2 model of the solc compact AST** as it appears in the `.asts.json` dump produced by `certoraRun --dump_asts`, plus migration of the two autosetup consumers of that dump onto it.

- **74 node models** (57 Solidity + 17 Yul), transcribed field-by-field from the OpenZeppelin `solidity-ast` JSON Schema (vendored at `solidity_ast/schema/schema.json`, v0.4.62, MIT — see `schema/NOTICE`). The schema unions solc versions >= 0.6: version-added fields are optional, genuinely-null fields are nullable.
- **Discriminated unions** (`Expression`, `Statement`, `TypeName`, `SourceUnitNode`, `ContractBodyNode`, `Node`, `YulStatement`, `YulExpression`) on `nodeType`, each with an `UnknownNode` fallback member so an unrecognized node type from a newer solc degrades per-node instead of failing the file.
- **Loader** (`AstDump.load` / `load_cached`) for the dump's 3-level structure `{original_file: {source_file: {node_id: node}}}`: validates each source's `SourceUnit` tree exactly once (the flat id map duplicates nodes nested in ancestors), derives the id index by traversal, keeps the raw map alongside. Vyper sources pass through raw behind an explicit `raw_kind="vyper"`; a source the models reject entirely degrades to `raw_kind="parse_failed"` with the raw map retained.
- **Traversal utilities**: iterative `walk`/`find_all`, typed `build_parent_map`, `iter_nodes_of_type` (typed-first + raw flat-map sweep for nodes the typed walk can't reach), and `build_parent_graph_json` — a byte-compatible relocation of the legacy parent-graph heuristic.

## Consumer migration (parity-pinned)

- `setup_prover.py`: declared-contracts and inheritance/abstract extraction now go through typed `ContractDefinition` access with a raw-scan fallback; `generate_ast_graph` delegates to the relocated builder — output file stays **byte-identical** (golden test).
- `auto_munges.py`: `.code` access detection iterates typed `MemberAccess` candidates with the raw sweep, so no access can be hidden by an unknown ancestor node or an unparsable source. Patches now read/target the source file the offsets refer to (previously a `.code` access in an in-scope *imported* file produced a patch against the compilation unit's main file at foreign offsets — silent source corruption) and are deduplicated across compilation units.

## Guardrails

- `tests/solidity_ast/test_schema_conformance.py`: machine-checks every model against the vendored schema — coverage in both directions, per-property presence/requiredness/nullability/enum/one-level kind (155 checks). Two deliberate, allowlisted deviations: `InlineAssembly.evmVersion`/`flags` are open strings (a closed EVM-fork enum would demote every assembly-containing file on the first post-schema solc release).
- Committed real fixtures generated via `certoraRun --dump_asts` for solc **0.6.12 / 0.7.6 / 0.8.30** (path-sanitized; regeneration script included, dev-only) + a hand-written Vyper-mixed fixture. Strict parsing asserts **zero unknown nodes and zero unmodeled fields** — the schema-drift alarm.
- Parity tests re-implement the legacy raw-dict algorithms inline and assert identical results, plus an end-to-end `detect_code_accesses` run producing correct patches.
- 421 tests pass, pyright clean (CI scope).

## Known pre-existing issue (not fixed here)

solc `src` offsets are byte offsets, but patch extraction/application slices Python strings by character index — any non-ASCII character before a `.code` access shifts the spans (detect and apply err consistently, so the mismatch check does not catch it). Fixing it properly touches `apply_code_access_patches`/`_inject_load_code_into_contract` and deserves its own change.

## Manual validation

```
uv run python -m certora_autosetup.solidity_ast <path/to/.certora_internal/all_asts.json>
```
Point it at any existing project dump: prints per-source parse status, node counts, unknown/unmodeled fields (should be none), and contracts with resolved inheritance.


shellygr:
@jtoman @jar-ben I do not expect you to read 200K loc. John you were excited about this so we can discuss design choices over DMs

certora_autosetup/solidity_ast/base.py
```diff
@@ -0,0 +1,136 @@
...
+- reference to a schema union helper
+  (Expression/Statement/TypeName/...) -> string forward ref to the union alias,
+                                         resolved by ``unions.model_rebuild`` wiring
+- schema enum property                -> shared ``Literal`` alias below when it matches a
+                                         helper definition, inline ``Literal[...]`` otherwise
+- python-keyword property name        -> trailing-underscore field with ``alias=`` (the only
+                                         known case is ``UsingForDirective.global``)
+"""
+
+from __future__ import annotations
```
shellygr: is it really useful here?

shellygr: Claude answers: You're right — removed it here. base.py has no forward references, and requires-python is >=3.12 so `X | None` (PEP 604) evaluates natively, so the future import was inert. Dropped it from base.py and the other non-forward-ref modules (loader, traversal, diagnostics, __main__) in a3b2d76. Kept it only in the six model modules (declarations, expressions, statements, types, yul, unions) where it's load-bearing — they use unquoted cross-module forward refs like `value: YulExpression` that only resolve when annotations are strings. 162 pure + 73 fixture tests pass.

certora_autosetup/solidity_ast/statements.py
```diff
@@ -0,0 +1,169 @@
+"""Statement nodes of the Solidity compact AST (members of the Statement union)."""
+
+from __future__ import annotations
```
jtoman: this needs to be expunged from all these files; it makes this code incompatible with future python versions.

shellygr: that bastard Claude lied to me when it said it removed it everywhere in a3b2d76...

shellygr: Claude answers: Done properly this time — 48bd338 removes `from __future__ import annotations` from every file in the PR, including the six model modules the earlier pass deliberately kept it in (that was a bad call: PEP 563 is a dead end and Python 3.14's PEP 649 lazy annotations make it a real forward-compat hazard, as you said). References to unions and not-yet-defined classes are now explicit string annotations; everything else evaluates natively on the >=3.12 floor. Full suite + pyright green, and a real dump re-validated clean.

jtoman (approved):
sure

PR #78, opened 2026-07-16
add --msg to prover runs

jar-ben:


composer/spec/source/prover.py
```diff
@@ -254,6 +254,15 @@ async def verify_spec(
 
                 summary = get_run_summary()
 
+                # Label the prover run with the --msg:
+                # the component (spec stem, else the main contract) and a 1-based
+                # iteration number. The spec stem carries a constant "autospec_"
+                # prefix that adds no signal, so drop it. prover_total_calls counts
+                # calls already completed.
```
naftali-g: why is claude so verbose in comments? I think this one can be safely removed

naftali-g (approved):

PR #79, opened 2026-07-16
Never pin the scene to hardhat's implicit default solc

shellygr:
## Problem

On a hardhat project whose `hardhat.config.js` has **no `solidity` entry** (e.g. a brownie-generated network-only stub), `npx hardhat config --json` reports hardhat's built-in default compiler (0.7.3). `HardhatManager` took that as the project's compiler and pinned the entire scene to `solc7.3` via `compiler_map` — on an all-`^0.8.0` codebase every file (including our own injected `DummyERC20Impl`) fails with `ParserError: Source file requires different compiler version`, and compilation analysis dies before any workaround can help.

Two additional defects kept the existing recovery path from firing:
1. `_detect_compiler_version_mismatch` matched the ParserError phrase within a single line, but solc hard-wraps its diagnostics (same class as the Yul stack-too-deep wrap bug fixed earlier).
2. The `compiler_version_mismatch` workaround was registered with `enabled=not solc_already_set`, i.e. disabled exactly when a build system pinned a wrong global solc.

## Fix

- `hardhat_config_extractor.js` additionally emits `solidityImplicitDefault: hre.userConfig.solidity === undefined`. When set, `HardhatManager` ignores the resolved compiler (logging why) and leaves `solc_version=None`, so compilation falls back to `DEFAULT_SOLC_VERSION` + per-contract pragma resolution.
- The mismatch detector matches the marker phrase across wrapped lines (`\s+` between words), then maps the match back to a line index for the existing path/pragma context extraction.
- The workaround is enabled even when a global solc is configured: the detector only fires when that compiler provably cannot parse a file, and `_seed_compile_maps` promotes the scalar into `compiler_map` up front, so a per-contract override is always safe.

## Validation

- 7 new unit tests (wrapped + single-line detection with a verbatim captured output, loop behavior with a wrong global solc, hardhat config extraction with implicit/explicit/legacy shapes); full suite passes (234).
- Extractor checked live in the mass-test container against a real brownie-stub project (`solidityImplicitDefault: true`, resolved 0.7.3 ignored) and a project with an explicit `solidity` block (`false`, 0.8.28 kept).
- End-to-end re-run of the failing project on this branch: compilation analysis now resolves the scene from pragmas (solc8.34) and completes solc compilation; the run proceeds to the next phase.

### DIFF
--- a/certora_autosetup/build_systems/hardhat.py
+++ b/certora_autosetup/build_systems/hardhat.py
@@ -229,6 +229,21 @@ class HardhatManager(BuildSystemManager):
                 )
             return str(path_obj)
 
+    @staticmethod
+    def _describe_solidity_version(solidity: Any) -> str:
+        """Best-effort human-readable version from a resolved `solidity` block (for logging)."""
+        if isinstance(solidity, str):
+            return solidity
+        if isinstance(solidity, dict):
+            compilers = solidity.get("compilers")
+            if isinstance(compilers, list) and compilers:
+                versions = {str(c["version"]) for c in compilers if isinstance(c, dict) and c.get("version")}
+                if versions:
+                    return ", ".join(sorted(versions))
+            if solidity.get("version"):
+                return str(solidity["version"])
+        return "unknown"
+
     def _extract_config_from_json(self, config_data: Dict[str, Any], config_type: str) -> HardhatConfig:
         """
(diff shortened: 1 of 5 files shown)


jar-ben (approved):
lgtm

PR #80, opened 2026-07-16
Add findings to the report

naftali-g:


composer/spec/source/report/collect.py
```diff
@@ -101,6 +101,25 @@ async def __call__(self, input: ReportComponentInput[R], /) -> dict[RuleName, Ve
...
+    failing trace excerpt (prover: the rule's ``cex_dump``). Both optional — a finding degrades to
+    property/group text when absent."""
+    analysis: str | None = None
+    counterexample: str | None = None
+
+
+class EvidenceFetcher(Protocol):
+    """Backend hook: given a violated rule's run link + name, return its `RuleEvidence` (or ``None``
+    when the backend has none). Prover reads the run-scoped CEX-analysis capture; foundry has none."""
+    async def __call__(self, link: str | None, rule_name: str, /) -> "RuleEvidence | None":
```
jtoman: absolutely no reason to quote this type annotation. a little curious you have the "positional only" marker here but I'm not mad about it.

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
+"""Synthesize Sherlock-``IssueIn``-shaped findings from violated rules.
+
+For each violated rule (a `RuleVerdict` with ``outcome == Outcome.BAD``) this reshapes the
+counterexample analysis the run already produced — looked up via the backend `EvidenceFetcher` — into
+a `Finding`. One structured LLM call per violation, fed the *distilled* analysis rather than the raw
+counterexample, so the expensive counterexample reasoning is not repeated. Best-effort per finding:
```
jtoman: what expensive reasoning? the cex analysis in `cex_analysis_raw` or whatever it is is single shot llm call?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+any failure drops that one finding, never the report.
+
+v1 scope: prover-only (a foundry BAD is an author-declared demonstration, not a discovered bug).
```
jtoman: I... don't think this is true? if you write an invariant test or a fuzz test, it's worth an explanation of why that invariant failure happened or why the fuzz test parameters chosen were problematic right? But for v1 this is fine, but I just want to make sure we don't buy this reasoning and avoid revisiting this for foundry in the future.

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+Severity is LLM-assigned from the default Sherlock rubric via the Impact × Likelihood matrix in the
+system template. ``IssueIn.locations`` is not produced here — a run knows only local paths and CVL-spec
+lines, so the submission layer reconstructs source locations from the engagement scope + counterexample
+(the accurate report-time locator is on ``FindingProvenance``: rule name, spec file, prover-run link).
```
jtoman: I'm sure I'll understand this more as I read the rest of the file, but having read it without further context, this last sentence is unintelligible.

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+class FindingDraft(BaseModel):
+    """The LLM-authored prose + severity for one finding. Code fills locations / PoC / references /
+    provenance around it. Mirrors the Sherlock ``IssueContent`` (minus the machine-filled fields)."""
+    title: str = Field(max_length=200, description="One-line, issue-specific title.")
+    severity: Severity = Field(description="critical | high | medium | low | informational.")
+    severity_reasoning: str = Field(
+        max_length=4000,
+        description="1-3 sentences: the Impact tier, the Likelihood tier, and the matrix cell they select.",
+    )
+    summary: str = Field(max_length=2000, description="1-3 sentence tl;dr.")
```
jtoman: "tl;dr" really?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+
+Severity = Literal["critical", "high", "medium", "low", "informational"]
+
+#: Bound the counterexample we inline (as prompt input and as ``proof_of_concept``).
+_MAX_CEX_CHARS = 8000
+
+
+class FindingDraft(BaseModel):
+    """The LLM-authored prose + severity for one finding. Code fills locations / PoC / references /
+    provenance around it. Mirrors the Sherlock ``IssueContent`` (minus the machine-filled fields)."""
```
jtoman: friendly reminder these doc strings on the class end up in the schema documentation sent to the LLM, so think carefully about whether this is the prompting you want for the LLM (saying stuff like "machine-filled fields" and what not is not something that should be relevant to the llm imo).

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+    severity: Severity = Field(description="critical | high | medium | low | informational.")
+    severity_reasoning: str = Field(
+        max_length=4000,
+        description="1-3 sentences: the Impact tier, the Likelihood tier, and the matrix cell they select.",
+    )
+    summary: str = Field(max_length=2000, description="1-3 sentence tl;dr.")
+    description: str = Field(max_length=50000, description="Full technical description grounded in the counterexample.")
+    impact: str = Field(max_length=20000, description="Concrete consequence if exploited.")
+    attack_path: str | None = Field(default=None, max_length=20000, description="Step-by-step exploit path, or null.")
+    assumptions_and_uncertainties: str | None = Field(default=None, max_length=10000, description="Assumptions / uncertainties, or null.")
```
jtoman: so, I know my claude code LOVES to truncate long strings. Is there a compelling reason here to add all these max lengths? a 50k budget is pretty generous for the description, but like... why not infinite? What are we protecting against?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+
+class FindingDraft(BaseModel):
+    """The LLM-authored prose + severity for one finding. Code fills locations / PoC / references /
+    provenance around it. Mirrors the Sherlock ``IssueContent`` (minus the machine-filled fields)."""
+    title: str = Field(max_length=200, description="One-line, issue-specific title.")
+    severity: Severity = Field(description="critical | high | medium | low | informational.")
+    severity_reasoning: str = Field(
+        max_length=4000,
+        description="1-3 sentences: the Impact tier, the Likelihood tier, and the matrix cell they select.",
+    )
```
jtoman: a question: why let the llm compute the severity itself? Why not have it output the impact and likelihood and have us compute the severity itself?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+    """A violated rule may formalize several properties; pick a stable representative (first by
+    title) to describe the finding. They describe the same violation, so the choice affects only
+    which prose seeds the write-up."""
+    return sorted(props, key=lambda p: p.title)[0] if props else None
+
+
+def _trim(text: str | None) -> str | None:
+    if not text:
+        return None
+    return text if len(text) <= _MAX_CEX_CHARS else text[:_MAX_CEX_CHARS] + "\n…(truncated)"
```
jtoman: did you ask for this? or is this claude suffering from its "NO LONG TEXT" psychosis?

I see no real compelling reason to truncate this text short of like, optimizing token costs, in which case these strings are small potatoes compared to the token consumption elsewhere

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+    rules: list[RuleVerdict],
+    properties: list[FormalizedProperty],
+    groups: list[PropertyGroup],
+    fetch_evidence: EvidenceFetcher | None,
+    llm: BaseChatModel,
+) -> list[Finding]:
+    """One `Finding` per violated rule (concurrent, best-effort). Returns ``[]`` for a non-prover
+    backend or when nothing is violated."""
+    if backend != "prover":
+        return []
```
jtoman: It will not surprise you to learn that I hate keying off the backend id like this. There must be a better way to let backend's opt out from this. In fact, I'm surprised `fetch_evidence == None` doesn't imply "don't build findings".

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # Reverse index a violated rule -> the properties it breaks -> their audit group (for prose context).
+    props_by_ref: dict[RuleRef, list[FormalizedProperty]] = {}
+    for p in properties:
+        for ref in p.rule_refs:
+            props_by_ref.setdefault(ref, []).append(p)
+    group_by_key: dict[PropertyKey, PropertyGroup] = {}
+    for g in groups:
+        for k in g.members:
+            group_by_key.setdefault(k, g)
```
jtoman: so this is "first wins"? Is that stable across runs? Is that the right choice at all?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+                "autoprove_report_findings_prompt.j2",
+                contract_name=contract_name,
+                rule_name=rule.name,
+                property_title=prop.title if prop else None,
+                property_description=prop.description if prop else None,
+                property_sort=prop.sort if prop else None,
+                group_title=group.title if group else None,
+                group_description=group.description if group else None,
+                analysis=ev.analysis if ev else None,
+                counterexample=_trim(ev.counterexample) if ev else None,
```
jtoman: any particular reason you aren't just passing in the prop/group here?

Also, I could have sworn the integrity rules we wrote for the collection/grouping meant that each rule ref is assigned at least one property/group. When could we get orphans like this?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+
+    findings = await asyncio.gather(*[_one(r) for r in bad])
+    return [f for f in findings if f is not None]
+
+
+def _pick_property(props: list[FormalizedProperty]) -> FormalizedProperty | None:
+    """A violated rule may formalize several properties; pick a stable representative (first by
+    title) to describe the finding. They describe the same violation, so the choice affects only
+    which prose seeds the write-up."""
+    return sorted(props, key=lambda p: p.title)[0] if props else None
```
jtoman: UH. This isn't *really* true though and kind of a bad assumption imo. Suppose we have two extracted properties:

1. The `allowDeposits` function must revert if msg.sender is not in `adminRoles`
2. The `allowDeposits` function must revert if `allowedActor` is `address(0)`

and both of these properties are formalized by the single rule `allowDeposits_revert_characteristic`:

```cvl
rule allowDeposits_revert_characteristic(env e, address allowedActor) {
   allowDeposits@withrevert(e, allowedActor);
   assert lastReverted <=> (allowedActor == 0 || !isAdmin(e.msg.sender));
}
```

And the CEX finds a case where `allowedActor == 0` but here we select property #1, i.e., "revert on not an admin role". Won't this mean we end up asking this finding llm call thing to explain why "allowedActor == 0" violates "revert on admin role", which like, it doesn't?

composer/spec/source/report/findings.py
```diff
@@ -0,0 +1,153 @@
...
+    return Finding(
+        title=draft.title,
+        severity=draft.severity,
+        content=IssueContent(
+            summary=draft.summary,
+            description=draft.description,
+            impact=draft.impact,
+            attack_path=draft.attack_path,
+            assumptions_and_uncertainties=draft.assumptions_and_uncertainties,
+            proof_of_concept=_trim(ev.counterexample) if ev else None,
```
jtoman: f--- stop it claude!

composer/spec/source/report/render.py
```diff
@@ -143,6 +165,7 @@ class ReportTemplateParams(TypedDict):
     prover_runs: list[RunView]
     rule_counts: list[ChipView]
     group_counts: list[ChipView]
+    findings: list[FindingView]
```
jtoman, line 172: <3

composer/spec/source/report/schema.py
```diff
@@ -157,6 +157,55 @@ class CoverageReport(BaseModel):
...
 
+# ---------------------------------------------------------------------------
+# Findings — violated rules surfaced as audit issues.
+#
+# `Finding` carries the Sherlock Audit Engine "Submit Issue" fields (schema ``IssueIn``,
+# POST /v1/engagements/{id}/issues) that a run can populate at report time: ``title``, ``severity``,
+# and the ``content`` prose. One finding is emitted per violated rule (a `RuleVerdict` with
+# ``outcome == Outcome.BAD``); its prose is synthesized from the rule's counterexample analysis and
+# the property/group it breaks (see ``report/findings.py``). Sherlock length caps are mirrored but
+# not enforced (report.json is lenient — the submission endpoint is the validating gate).
```
jtoman: oh THIS is where the length caps come from? But ... wait what API are we targeting here?

composer/spec/source/report/schema.py
```diff
@@ -157,6 +157,55 @@ class CoverageReport(BaseModel):
...
+# ---------------------------------------------------------------------------
+
+class IssueContent(BaseModel):
+    """A finding's free-text sections — the ``IssueIn.content`` shape."""
+    summary: str = Field(max_length=2000, description="Short tl;dr of the finding, in one to three sentences.")
+    description: str = Field(max_length=50000, description="Full technical description of the vulnerability and how it manifests.")
+    impact: str = Field(max_length=20000, description="Concrete consequence if exploited (funds at risk, DoS, data exposure, ...).")
+    attack_path: str | None = Field(default=None, max_length=20000, description="Step-by-step exploit path, if applicable.")
+    assumptions_and_uncertainties: str | None = Field(default=None, max_length=10000, description="Assumptions relied on and anything the submitter is unsure about.")
+    proof_of_concept: str | None = Field(default=None, max_length=65536, description="Executable PoC, or the prover counterexample demonstrating the issue.")
```
jtoman: this looks super similar to the fields found in the `findings.py` FindingDraft schema, common basemodel class?

composer/spec/source/cex_capture.py
```diff
@@ -0,0 +1,40 @@
...
+
+The autoprove prover tool already runs an LLM analysis of every violated rule during the run
+(``TrivialFanoutCexHandler`` -> ``analyze_cex_raw``); that text is otherwise consumed only as agent
+feedback and discarded. This in-memory, run-scoped store captures it keyed by rule name
+(last-write-wins across prover iterations) so the report phase can reshape the *final* iteration's
+analysis into a finding without re-reasoning about the counterexample.
+
+In-memory is sufficient: the report phase runs in the same process as formalization within
+``composer.pipeline.core.run_pipeline``. A ``BaseStore``-backed variant (cf.
+``composer.prover.report_store``) would be the resume-safe upgrade.
```
jtoman: I strongly suspect we'll someday want to use the agentic handler for the cex in autoprover, so make sure we're not baking in an assumption about always using the TrivialFanoutCexHandler here.

That said, why *not* use the `BaseStore` version? seems like we'd want to save this sort of thing?

composer/templates/autoprove_report_findings_system.j2
```diff
@@ -0,0 +1,50 @@
...
+
+  | Likelihood \ Impact | High     | Medium | Low    |
+  | High                | critical | high   | medium |
+  | Medium              | high     | medium | low    |
+  | Low                 | medium   | low    | low    |
+
+Use `informational` ONLY for a property break with no real-world exploit path — a code-quality or
+specification observation, not a vulnerability. A prover counterexample proves the property CAN be
+broken, so never downgrade a genuine break to `informational` just because it looks hard to reach;
+instead lower the Likelihood tier and let the matrix decide.
```
jtoman: if we expect informational to never be the correct output, why even include it in the output schema?

This also suggests we shouldn't let the agent choose this, but pick the two axes on this matrix?

jtoman (changes requested):

composer/spec/source/report/findings.py
```diff
@@ -1,153 +1,159 @@
...
-            ev = await fetch_evidence(rule.prover_link, rule.name) if fetch_evidence else None
-            prop = _pick_property(props_by_ref.get(rule.ref, []))
-            group = group_by_key.get(prop.key) if prop else None
+            ev = await fetch_evidence(rule.name)
+            # Every rule in the report is referenced by >=1 property (collect drops orphans), so pass
+            # ALL of them: the rule may jointly formalize several, and only the counterexample says
+            # which one actually broke — that is the LLM's job, not ours to pre-guess.
+            props = props_by_ref.get(rule.ref, [])
+            # A rule's properties may share a group (some may be in none); dedupe by slug, first-seen order.
+            rule_groups = list({g.slug: g for p in props if (g := group_by_key.get(p.key)) is not None}.values())
```
jtoman, line 129: neat trick

composer/spec/source/report/findings.py
```diff
@@ -1,153 +1,159 @@
...
             user = load_jinja_template(
```
jtoman: strong preference for you to move this our `TypedTemplate` mechanism, it will let you automatically get template fuzz testing (among other things).

composer/spec/source/report/findings.py
```diff
@@ -1,153 +1,159 @@
...
-        severity=draft.severity,
+        severity=severity_for(draft.impact_level, draft.likelihood_level),
         content=IssueContent(
-            summary=draft.summary,
-            description=draft.description,
-            impact=draft.impact,
-            attack_path=draft.attack_path,
-            assumptions_and_uncertainties=draft.assumptions_and_uncertainties,
-            proof_of_concept=_trim(ev.counterexample) if ev else None,
+            **{f: getattr(draft, f) for f in AuthoredContent.model_fields},
```
jtoman, line 173: what on god's green earth???

`**AuthoredContent.model_dump(draft)` would also work (probably), but I'm not sure its that much more readable/less gross. Whatever.

composer/spec/source/cex_capture.py
```diff
@@ -1,40 +1,48 @@
...
+
+    async def record(self, rule_name: str, analysis: str, counterexample: str | None = None) -> None:
+        """Store one rule's analysis under ``rule_name``, which must be the bare top-level rule name
+        (``RulePath.rule``) — exactly what the report's ``RuleVerdict.name`` carries (POU's
+        ``context[0]``) — so the findings join is an exact match."""
+        await self.store.aput(
+            self.namespace,
+            rule_name,
+            CexAnalysis(analysis=analysis, counterexample=counterexample).model_dump(mode="json"),
+        )
```
jtoman: Wait. I didn't notice this in the last round: this keys on the `path.rule`. But for a parametric rule like:

```
run whatever(method f) { ... }
```

that's just `whatever`. The `f` instantiation is part of the RulePath method (or contract if it exists too). Keying on `rule` alone *might* be what you want, but it means the last CEX within any rule gets stored, and all the others get silently overwritten...

composer/spec/source/pipeline.py
```diff
@@ -151,11 +153,15 @@ async def fetch_verdicts(
...
+    def findings_evidence(self) -> EvidenceFetcher | None:
+        # Prover runs capture per-rule counterexample analysis, so this backend opts into findings.
+        return self._fetch_evidence
+
+    async def _fetch_evidence(self, rule_name: str) -> RuleEvidence | None:
         # The run captured each violated rule's counterexample analysis as it happened; hand the
-        # final-iteration analysis to the report's findings synthesizer. None -> the finding degrades
-        # to property/group text.
-        rec = self._analysis_store.get(rule_name)
+        # final-iteration analysis to the findings synthesizer. None -> the finding degrades to
```
jtoman: ditto here; do we want one finding per rule? What if a rule is broken for different reasons across different instantiations?

jtoman (commented):
all of this looks much, much better. One major question about the grouping/granularity of saving/restoring CEX. Sorry I didn't notice it before...

composer/spec/source/cex_capture.py
naftali-g: right. we need per rule+method

composer/spec/source/report/findings.py
naftali-g: done

composer/spec/source/report/findings.py
```diff
@@ -46,6 +48,27 @@ def severity_for(impact: ImpactLevel, likelihood: LikelihoodLevel) -> SeverityTi
...
 
+class FindingsSystemParams(TypedDict):
+    """``autoprove_report_findings_system.j2`` takes no parameters — the instructions are static.
+    Declared empty rather than skipped so the template is still covered by the strict-render fuzz
+    test: adding a ``{{ ... }}`` without declaring it here then fails."""
+
+
+class FindingsPromptParams(TypedDict):
+    """The full, typed context of ``autoprove_report_findings_prompt.j2``. Every key is required;
+    the nullable ones carry ``None`` when the backend supplied no evidence (the template guards them
```
jtoman: > "nullable ones"
> *looks inside*
> no nullable types

okay claude

composer/spec/source/cex_capture.py
```diff
@@ -1,48 +1,65 @@
...
 
+from composer.prover.ptypes import RulePath
+
+#: ``asearch`` defaults to 10, and a rule can be instantiated once per method of a wide interface.
+_LIMIT = 1000
+
+
+def _label(path: RulePath) -> str:
+    """Names one instantiation; "" when the rule is not parametric. Uses the explicit `RulePath` fields
+    rather than ``pprint()``, which drops the contract whenever a method is set."""
```
jtoman, line 22: that sounds like a bug in pprint() tbh, but not for fixing here.

composer/spec/source/cex_capture.py
```diff
@@ -1,48 +1,65 @@
...
+        """Every captured instantiation of ``rule_name``, in a stable order (backends disagree on
+        search order, and the report should not vary between runs)."""
+        items = await self._search(rule_name)
+        return sorted((CexAnalysis.model_validate(i.value) for i in items), key=lambda a: a.label)
+
+    async def forget_rule(self, rule_name: str) -> None:
+        """Drop ``rule_name``'s records, so an instantiation that failed in an earlier iteration but
+        passes now cannot be written up as a current failure."""
+        for item in await self._search(rule_name):
+            await self.store.adelete(self.namespace, item.key)
```
jtoman, line 62: I hate that we have to do element wise ops like this instead of an (arguably) simpler `delete from ...` but whatever, that's the api we're dealing with.

jtoman (approved):
I'm finally done giving you a hard time about this PR :)

PR #81, opened 2026-07-16
autosetup: fetch sanity sources from source-files API, not the output tar

jar-ben:


shellygr (approved):

shellygr:
does this one require a bump of ProverCLI? or is that the next PR you have on the pipe?

PR #83, opened 2026-07-17
Honor the project-declared EVM version + Yul-steps escalation rung

shellygr:
## Motivation

Projects can pin an EVM target older than solc's default (`evmVersion: "paris"` vs the `shanghai` default of solc >=0.8.20). The difference is not cosmetic: shanghai codegen can fail with YulException stack-too-deep where the declared target compiles, besides producing bytecode for a different target than the project deploys. Autosetup never read the declared version — a corpus project failed compilation analysis in all three PR #66 validation sweeps purely because of this (isolated experimentally: with the declared `paris`, every solc/optimizer/Yul-steps combination compiles; without it, none does).

## Changes

**Honor the declared EVM version end-to-end**
- Parse foundry.toml `evm_version` (incl. profile merge) and hardhat `settings.evmVersion`; emit scalar `solc_evm_version` from the shared build-system conf builder.
- Workaround manager seeds the declared scalar into a total `solc_evm_version_map` (certora-cli forbids scalar+map together and requires maps to cover every file), collapsing back to the scalar on finalize. The cancun workaround now overrides only the detected contract while the rest keep the declared version.
- New reactive `invalid_evm_version` workaround: if solc rejects the declared name ("Invalid EVM version requested." — same text for unknown and too-new names), drop the setting and fall back to solc defaults. No solc-version/EVM-name table to maintain: the compiler is the authority.
- Merged-conf reconciliation (removal propagation, map-supersedes-scalar with totality fill) + `solc_evm_version_map` fill in `sync_compiler_maps_with_files` for files injected after compilation analysis; `solc_evm_version` joins fixconf's strippable low-priority keys.

**Hardhat config extractor: survive `moduleResolution: nodenext`**
The ts-node registration forces `module: commonjs`; a project tsconfig pinning `moduleResolution: "nodenext"` then fails TS5110 even in transpileOnly mode — silently degrading the WHOLE hardhat config (solc version, optimizer, viaIR, evmVersion) to defaults. Now overrides `moduleResolution: node` alongside.

**Yul ladder: try solc's default Yul steps before relaxing autofinders**
With optimizer + via-ir, certoraRun substitutes finder-friendly Yul steps (full inliner removed). These give less stack relief than solc's own pipeline and can themselves cause stack-too-deep in functions the default pipeline compiles (isolated: fails under the substitute steps at any runs value, compiles under `strict_solc_optimizer`). New escalation rung between adding the optimizer and relaxing the autofinder assertion.

## Validation

- 251 unit tests pass (`-m 'not expensive'`), pyright clean. New tests: foundry/hardhat evm_version parsing, seeding/normalize round-trips, invalid-version drop, cancun+declared reconciliation, merged-conf helper, sync fill, strict-steps rung ordering.
- End-to-end: both corpus contracts of the motivating project now complete compilation analysis (previously failed in all three sweeps after ~21 min of futile ladder walking each). Observed ladder: declared `paris` honored → stack-too-deep → +optimizer → +`strict_solc_optimizer` → success, with the autofinder assertion never relaxed.


certora_autosetup/build_systems/base.py
```diff
@@ -29,6 +29,9 @@ class BuildSystemConfig(ABC):
     optimizer: Optional[bool] = None
     optimizer_runs: int = 200
     via_ir: Optional[bool] = None
+    # Declared EVM target (foundry `evm_version`, hardhat `settings.evmVersion`);
```
jar-ben: I would drop this comment, since we now have more build systems than just foundry and hardhat.

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -179,6 +185,16 @@ def _seed_compile_maps(self, config: Dict, contracts: List[ContractHandle]) -> N
...
+        # requires every *_map to cover all files).
+        evm_version = config.pop("solc_evm_version", None)
+        if evm_version:
+            config.setdefault("solc_evm_version_map", {})
+            for c in contracts:
+                config["solc_evm_version_map"].setdefault(c.contract_name, evm_version)
+            self._evm_version_seeded = True
+
     def _normalize_compile_maps(self, config: Dict) -> None:
         """Collapse a uniform compiler_map / solc_via_ir_map back to its scalar if all values are same.
```
jar-ben: Update the comment

certora_autosetup/utils/enhanced_config_manager.py
```diff
@@ -1120,6 +1151,27 @@ def sync_compiler_maps_with_files(
...
                 modified = True
+            # Extend: the prover requires the map to cover every file, and a map
+            # entry cannot say "use this solc's default" — fill contracts newly
+            # injected into files (mocks/harnesses) with the declared scalar from
+            # the reference conf, else the map's most common value.
+            evm_map = conf_object["solc_evm_version_map"]
+            if evm_map:
+                fill = (
+                    self.reference_compiler_maps.get("solc_evm_version")
+                    or Counter(evm_map.values()).most_common(1)[0][0]
```
jar-ben, line 1162: I don't like this guessing, but getting the right evm version might be a non-trivial change, so perhaps let's go with the guessing heuristing for now and refine if we see issues.

jar-ben (commented):
Looks ok, just a few minor comments. Also, the Claude comments are (as usually) sometimes too verbose, but it's mostly acceptable. 

Unfortunately, it took me a while to review and the branch is now out of date with conflicts.

certora_autosetup/utils/enhanced_config_manager.py
shellygr: to be fair, it's a rare thing to see an evm_version set explicitly

jar-ben (approved):

PR #86, opened 2026-07-17
Make the system doc optional

jtoman:
It will be useful to make the system doc optional for the system analysis for reusing it in other contexts.

Also adds a TaskRunner abstraction.

### DIFF
--- a/composer/templates/application_analysis_prompt.j2
+++ b/composer/templates/application_analysis_prompt.j2
@@ -1,3 +1,5 @@
+{% from "application_analysis_macros.j2" import input_phrase %}
+
 # Background
 
 {% if sort == "greenfield" %}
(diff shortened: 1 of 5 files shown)


composer/templates/application_analysis_prompt.j2
```diff
@@ -111,10 +118,10 @@ In addition, each explicit contract should have an associated list of "Contract
 Each explicit contract is comprised of "contract components" which serve to organize related functionality.
 Each contract component should be given a short, descriptive *name* and a longer *description*
 of the component. This description should focus on *what* the component does, not *how* it does it.
-A system or design document may say that a contract implements a well-known interface (e.g., ERC20);
+{{ input_phrase(true, true) }} may indicate that a contract implements a well-known interface (e.g., ERC20);
```
shellygr: nit: named parameters?

shellygr (approved):

PR #88, opened 2026-07-18
autosetup: never emit a conf with both solc and compiler_map

shellygr:
## Problem

Autosetup can generate certoraRun confs that carry BOTH a scalar `solc` and a per-contract `compiler_map`. certora-cli hard-rejects the pair:

```
compiler map flags cannot be set with other compiler flags. Got {'compiler_map': ..., 'solc': ...}
```

Observed on foundry projects whose build artifacts were produced with a different solc than the `foundry.toml` pin (`_precompute_compiler_settings` maps every contract to the artifact version while the build system's scalar `solc` stays in the conf).

### DIFF
--- a/certora_autosetup/utils/compilation_workarounds.py
+++ b/certora_autosetup/utils/compilation_workarounds.py
@@ -150,13 +150,21 @@ class CompilationWorkaroundManager:
 
     def _seed_compile_maps(self, config: Dict, contracts: List[ContractHandle]) -> None:
         """Promote scalar ``solc``/``solc_via_ir`` into fully-populated maps.
+
+        When a map is already present (e.g. precomputed from build artifacts,
+        or in a user conf fed through fixconf) the scalar still acts as the
+        default for contracts the map doesn't cover, and is then dropped —
+        certoraRun rejects a conf carrying both a map and its scalar.
         """
-        if "compiler_map" not in config:
-            default = config.pop("solc", None) or self.solc_default_version
-            config["compiler_map"] = {c.contract_name: default for c in contracts}
-        if "solc_via_ir_map" not in config:
-            via_ir = config.pop("solc_via_ir", False)
-            config["solc_via_ir_map"] = {c.contract_name: via_ir for c in contracts}
+        default = config.pop("solc", None) or self.solc_default_version
+        config.setdefault("compiler_map", {})
+        for c in contracts:
+            config["compiler_map"].setdefault(c.contract_name, default)
+
+        via_ir = config.pop("solc_via_ir", False)
+        config.setdefault("solc_via_ir_map", {})
+        for c in contracts:
+            config["solc_via_ir_map"].setdefault(c.contract_name, via_ir)
 
     def _normalize_compile_maps(self, config: Dict) -> None:
         """Collapse a uniform compiler_map / solc_via_ir_map back to its scalar if all values are same.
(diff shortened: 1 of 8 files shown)


certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -376,12 +395,17 @@ def run_compilation_with_workarounds(
                 # tried, not in the same pass that just added it (the live
                 # "solc_optimize in config" check would otherwise see the value
                 # the previous workaround set seconds ago and stop asserting
-                # autofinder success without ever testing the optimizer).
+                # autofinder success without ever testing the optimizer). A
+                # solc_optimize_map counts as "optimizer already tried": it is
+                # project-configured, and the add rung never fires next to it.
                 detect_fn=lambda output: (
```
shellygr: i'm worried about this change as we could have it misconfigured for one of the contracts in the map.

certora_autosetup/utils/enhanced_config_manager.py
```diff
@@ -762,6 +762,31 @@ def extract_solc_version_from_pragma(
     # compilation_workarounds.
     # =========================================================================
 
+    # (scalar, per-contract map) conf-key pairs. certoraRun hard-rejects a conf
+    # that sets both members of a pair (certoraContextValidator:
+    # convert_to_compiler_map for solc/compiler_map, check_map_attributes for
+    # the rest), so a map always supersedes its scalar counterpart.
+    SCALAR_TO_MAP_KEYS = (
+        ("solc", "compiler_map"),
```
shellygr, line 770: may make sense to import names from certora-cli package

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: Valid concern — an entry in the map can be missing or "0" (optimizer off for that contract), and the previous revision would have skipped straight to relaxing the autofinder assertion without ever trying to enable it. Reworked in 9424974: the add-optimizer rung is now per-contract-aware — with a `solc_optimize_map` present it enables exactly the entries whose optimizer is off (missing/"0"), keeping explicit project runs values untouched and never setting the scalar next to the map. The escalation rung fires only once the optimizer is on globally or for every scene contract; one predicate (`_yul_optimizer_pending`) drives both rungs as complements, so there is no gap or overlap between them. Regression tests cover both the all-entries-on case (escalate) and the misconfigured-entry case (fill "0"→"200", keep "1", then escalate only if the error persists).

certora_autosetup/utils/enhanced_config_manager.py
shellygr: Claude answers: I looked into importing them: the names live in `certora_cli.CertoraProver.certoraContextAttributes` (`EvmProverAttributes.SOLC.get_conf_key()` etc.), but that module isn't cleanly importable from library code — certora_cli's internals use flat imports (`from Shared import ...`), so reaching it requires inserting the certora_cli package dir into `sys.path` (shadow risk for generic top-level names like `Shared`) and calling `Attrs.set_attribute_class(EvmProverAttributes)`, a process-global. I didn't want either as a runtime side effect of importing `enhanced_config_manager`. Compromise in 9424974: the pairs stay literal, and a new test (`test_scalar_to_map_keys_match_certora_cli`) resolves every name against the installed CLI's attribute definitions in a subprocess — an upstream rename/removal fails the suite instead of silently desyncing. If you'd rather have the runtime import anyway, say the word and I'll wire it with the sys.path shim.

jar-ben (approved):

PR #89, opened 2026-07-20
 AutoProve Plugin System, prop prompt improvements 

jtoman:
Adds a way to install "plugins" into the autoprover pipeline. Currently the only extension point is the property inference. We use importlib's entrypoints functionality for symbol discovery and instantiation.

* Pre-property plugins contribute `AnyPropertyGenerationInput`. These are sorted and concatenated with the task prompt into a cache friendly order. These plugins run concurrently with each other.
* Post-property plugins post-process the entire property list, and are run in plugin id order (I couldn't think of another sensible default ordering).

Via packaging magic, templates placed in "templates/" as a sibling to the loader class "just work" as arguments to the builders. The "host" (core) templates are accessed via "autoprover/" namespacing; there is some jinja black magic for this to work.

Sensible attempts at caching are made throughout.

Fixes a bug in autosetup where cancellation due to a sibling task failure would actually hang the whole program (at least until autosetup finished); `CancellationException` is a base exception, so it didn't go through the previous "except" handler, meaning we never killed "autosetup" during shutdown. In the `finally` block, we'd then wait for the (unkilled) process, hanging the orderly shutdown in the case of error until autosetup finished; not what we wanted.


composer/pipeline/core.py
```diff
@@ -146,14 +155,28 @@ def extract_task_id(idx: int) -> str:
...
-# ---- the driver --------------------------------------------------------------
 async def run_pipeline[P: enum.Enum, FormT: BackendResult, H, A: ArtifactIdentifier](
     backend: PipelineBackend[P, FormT, H, A],
     run: PipelineRun[P, H],
     *,
     interactive: bool = False,
     threat_model: Document | None = None,
     max_bug_rounds: int = 3,
+) -> CorePipelineResult[FormT]:
+    async with load_plugins(run) as plugins:
```
naftali-g, line 167: nit
```suggestion
    async with load_plugins(run) as plugin_manager:
```

naftali-g (approved):
Cool feature!

PR #90, opened 2026-07-20
Monitoring for budget pressure, prover abuse

jtoman:
Uses the monitoring feature of https://github.com/Certora/graphcore/pull/24 to add two monitors:
1. A prover monitor, which keeps the state of prover tool calls, and tracks whether the agent is "banging its head" against a set of rules. If so, it uses the new "reminder" channel to emit a "nag", urging the agent to either give up or try something new
2. A budget monitor, which accumulates the total cost of all LLM calls within a given "cost center". If we detect that the task is reaching a budget limit (defined as 80% of the allocated budget) we emit a "wrap it up" message. For verification tasks, we also halt any active feedback agents, and disable all feedback gating (with the idea that it's better to finish under budget with lower quality than blow the budget and return nothing).

Budget cancellation is co-operative, if the agent is overbudget/under budget pressure, it throws a special exception which is interpreted as the appropriate "I give up" type (only for the verification tasks).

As mentioned, there is a per cost center budget and a total "overall run" budget. The sum of cost center budgets does not necessarily have to be the overall budget, in fact, it is expected to be greater. This was chosen over a strict allocation of the total budget to different centers, with some arcane budget rebalancing for unused costs.

As part of this, and as predicted by @shellygr, we have moved the pricing information out of the display subsystem into the `llm/` module. We also more accurately track 1hr vs 5m cache costs, as this matters when computing budget pressure.


composer/spec/source/prover.py
```diff
@@ -235,8 +251,25 @@ async def verify_spec(
...
         if state["curr_spec"] is None:
             return "Specification not yet put on VFS"
+        
+        spec_hash = string_hash(
+            state["curr_spec"]
+        )
+
+        if len(state['prover_history']) > 0:
+            last_item = state["prover_history"][-1]
+            if last_item["sort"] == "run" and any(i == "TIMEOUT" for (_,i) in last_item) and last_item["spec_digest"] == spec_hash:
```
jar-ben: Why only timeouts? verified rules, violated rules, sanity failures, .., are also not transient

jar-ben: `last_item["spec_digest"] == spec_hash` is it possible in the current (or intended future) workflow that the .spec remains the same, but the .conf or .sol files changed? In such situations, a rerun makes sense.

composer/spec/source/prover.py
```diff
@@ -276,22 +309,101 @@ async def verify_spec(
...
+            if not stat:
+                all_verified = False
+                break
+
+        spec_hash = string_hash(
+            state["curr_spec"]
+        )
+
+        stuck_rules = {
+            k: v for (k,v) in result.raw_rule_status.items() if v in ("TIMEOUT", "ERROR", "SANITY_FAILED")
```
jar-ben: `"SANITY_FAILED"` is sometimes the expected and "right" result, especially for parametric rules/invariants if there is a method that is intended to always revert (in such case, the whole rule/invariant is also flagged with `SANITY_FAILED`. I am not sure autoprover can currently fix this, and nagging with "You may need to significantly change your approach, or skip the property if this is a persistent issue" might just drop the rule completely.

composer/spec/source/author.py
```diff
@@ -331,6 +335,41 @@ async def run(self) -> Command | str:
...
+- Then publish what remains via the `result` tool.
+</system-alert>
+"""
+
+
+def _author_monitor() -> "Callable[[SourceCVLGenerationState], MonitorReturn]":
+    """The author's monitor: budget wrap-up takes precedence; otherwise the
+    usual reminders-channel drain. On the (single) turn the budget warning
+    fires any pending reminders are dropped — moot, since the warning tells
+    the agent to ignore prover/feedback outcomes anyway."""
```
jar-ben, line 639: It might be good to propagate to the reporting part (or at least log) that we dropped validations.

jar-ben (commented):
Left some comments. The high-level approaches here look good; I think both the budget constraints and the check for "looping over the same prover input" can be helpful (and I think it would solve the real issue we have seen).

composer/spec/source/prover.py
jtoman: that's a great point actually.

jtoman: yeah, this assumes the only thing the agent can change is the spec. That's true currently, but will change. we'll have to digest the conf and/or VFS once those changes land.

composer/spec/source/prover.py
jtoman: I'm writing this comment from memory, but I thought there was something here that didn't nag about rules that are expected to fail. If a rule's "correct" outcome is a failure (whether sanity fail or whatever), the agent should mark those rules as expected to fail, which will suppress this nag.

If we don't skip "expected to fail" rules in the nagging, that's something I need to fix.

composer/spec/source/author.py
jtoman: :facepalm: of course, why didn't I think of this.

composer/spec/source/prover.py
jtoman: I just double checked, and we ignore the "expected to fail" state when considering nags; that's a bug

composer/diagnostics/budget.py
```diff
@@ -0,0 +1,214 @@
...
+    caps: Mapping[str, float]
+) -> Iterator[None]:
+    """Install the run's budget: ``total`` is the pool (the real bound on
+    spend) and ``caps`` are per-phase ceilings. Caps need not sum to the
+    pool — they only bound how much a single phase may hog, so each can be
+    generous; whatever a phase doesn't spend simply remains in the pool for
+    later phases."""
+    curr = _cost_centers.get()
+    if curr is not None:
+        raise RuntimeError("Not good")
```
ericeil: Should we use a more descriptive message here?

composer/diagnostics/budget.py
```diff
@@ -0,0 +1,214 @@
...
+    pool = BudgetCounter(total_budget=total, curr_cost=0.0)
+    prev = _cost_centers.set({
+        k: BudgetCounter(total_budget=v, curr_cost=0.0, parent=pool) for (k, v) in caps.items()
+    })
+    # Work running outside any named center (e.g. the report phase) accrues
+    # to — and feels pressure from — the pool directly.
+    prev_accum = _budget_accumulator.set(pool)
+    try:
+        yield None
+    finally:
```
ericeil, line 178: (I'm just learning about `@contextmanager` now - what an odd convention! - so forgive me if this question is nonsense): isn't the `finally` implied here?  Is this the idiomatic way to write these?

composer/diagnostics/budget.py
```diff
@@ -0,0 +1,214 @@
...
+        raise RuntimeError("Not good")
+    pool = BudgetCounter(total_budget=total, curr_cost=0.0)
+    prev = _cost_centers.set({
+        k: BudgetCounter(total_budget=v, curr_cost=0.0, parent=pool) for (k, v) in caps.items()
+    })
+    # Work running outside any named center (e.g. the report phase) accrues
+    # to — and feels pressure from — the pool directly.
+    prev_accum = _budget_accumulator.set(pool)
+    try:
+        yield None
```
ericeil: Is there a difference between `yield None` and the bare `yield` you use elsewhere?

composer/llm/pricing.py
```diff
@@ -0,0 +1,127 @@
...
+# OpenAI's published >272K input-token threshold for long-context
+# pricing. Once an individual call's input crosses this, the long
+# tier applies *for the full session* per OpenAI's terms; we
+# approximate that by switching on a per-message basis (a session
+# that drifts above 272K will mostly stay there).
+_OPENAI_LONG_CONTEXT_THRESHOLD = 272_000
+
+
+# Pricing tables transcribed from Anthropic + OpenAI rate cards.
+# Sources should be re-checked when new model families ship.
```
ericeil: Is there no way to query for the current rates programmatically?

ericeil: (Oh I see this is existing code copied from elsewhere)

ericeil (commented):
I like the use of context vars for this; it feels like a good fix for "ambient" data that would be a pain to pass around explicitly everywhere.  I left some naive comments, but overall the mechanism looks good to me.

composer/diagnostics/budget.py
jtoman: I forgot about this (obvious placeholder) message lol, I'll fix it

composer/diagnostics/budget.py
jtoman: Get ready for a crash course on context managers!

A context manager is just an object which defines two special functions `__enter__` and `__exit__`.

When you write:
```
with some_context_manager as x:
```

python calls `__enter__` on `some_context_manager`; the value returned by that function is used to bind to `x` (roughly, I'm glossing over/partially misremembering some details)

When the context block is exited (normally, or via exception) it calls the special `__exit__` method. If there is an exception being propagated, those are passed into the `__exit__` function.

So, what does `@contextmanager` do?

Well, now we need to understand generators. If the keyword `yield` appears anywhere in a function, python transforms it into a generator object. A generator object exposes three methods, two of which are important to us, `send` and `throw`. On the first call to `send`, the body of the function begins executing up until the first `yield` statement. When it reaches said `yield`, execution of the generator suspends, and the yielded value is returned from `send()`. If you call `send(something)` on the generator again, `something` becomes the value returned from the yield expression, and execution of the generator resumes from the `yield` until the function exits, or another yield is hit.

The eagle eyed reader (which you are) will notice these are just kotlin coroutines; it is left as an exercise to the reader how you could implement generators on top of kotlin coroutines. Fun bonus fact: before `async` became first class, it was all bolted on top of this generator api.

The dual of `send` is `throw`, instead of having the most recent `yield` return a value, you have it throw the passed in exception.

You can now probably guess how `@contextmanager` works. It roughly looks like this:

```python
class ContextManagerImpl:
    wrapped: Generator
   def __enter__(self):
         return self.wrapped.send(None)

    def __exit__(self, exception):
          if exception:
             self.wrapped.throw(exception)
          else:
             self.wrapped.send(None) 
```

where `wrapped` is the generator returned by calling the function decorated with `@contextmanager`. I'm glossing over some details (you can suppress exceptions during `__exit__` for example) but this is the basic shape of it.

Now with all that background we can understand why we need `finally`. Without it, if the `with` block is exited with an exception, the `yield` statement in `named_budget` would throw, and that would immediately terminate any further execution in the body which would in turn skip the cleanup statements. By wrapping the `yield`  in `try: ... finally: ...` we can ensure that any cleanup we need to do happens even if the with block terminates exceptionally.

TL;DR - isn't the finally implied here? -> No.

ericeil: Thanks for the explanation!

composer/diagnostics/budget.py
jtoman: no. can fix to be consistent

composer/llm/pricing.py
jtoman: It's a good question, and the answer is: not really. There are 3rd party efforts to catalog this stuff into some standardized format. Might be worth looking into.

jtoman:
image

example from a smoketest run

composer/diagnostics/cost_callback.py
```diff
@@ -0,0 +1,84 @@
...
+        cache_read = usage["cache_read_tokens"]
+        cache_write = usage["cache_write_tokens"]
+        # Fresh input is the total minus the two cache buckets, which the tier
+        # prices separately. Clamp against provider rounding wobble.
+        fresh_input = max(0, usage["total_input_tokens"] - cache_read - cache_write)
+        cache_write_rate = tier.cache_write_1h if self._long_cache else tier.cache_write
+
+        # thinking_tokens are a subset of total_output_tokens and bill at the
+        # output rate, so they need no separate term here.
+        per_mtok = (
```
shellygr: confusing name. it's the raw cost before normalizing for the "per mtok"

composer/llm/pricing.py
shellygr: I'm also not sure this updates only when new models ship. they can update pricing independently of releasing a new model

composer/llm/pricing.py
```diff
@@ -0,0 +1,136 @@
...
+    ("claude-opus-5",   ModelPricing(short=PriceTier(5.00, 25.00, 0.50, 6.25, 10.0))),
+
+    # ---- Anthropic ----
+    # claude-opus-4.5 / 4.6 / 4.7 share a rate card; older 4 / 4.1
+    # are pricier. Matching by prefix-of-prefix so "claude-opus-4-7"
+    # and "claude-opus-4-7-20260301" both hit the right entry.
+    ("claude-opus-4-7", ModelPricing(short=PriceTier(5.00, 25.00, 0.50, 6.25, 10.00))),
+    ("claude-opus-4-6", ModelPricing(short=PriceTier(5.00, 25.00, 0.50, 6.25, 10.00))),
+    ("claude-opus-4-5", ModelPricing(short=PriceTier(5.00, 25.00, 0.50, 6.25, 10.00))),
+    ("claude-opus-4-1", ModelPricing(short=PriceTier(15.00, 75.00, 1.50, 18.75, 30.00))),
```
shellygr: 4-1 and earlier were discontinued already iirc

composer/llm/pricing.py
```diff
@@ -0,0 +1,136 @@
...
+
+    ("claude-haiku-4-5", ModelPricing(short=PriceTier(1.00, 5.00, 0.10, 1.25, 2.00))),
+
+    # ---- OpenAI ----
+    # gpt-5.5 / 5.4 publish short (≤272K input) and long (>272K) tiers.
+    # Pro variants don't publish a cached-in discount (cache_read =
+    # base input). Mini/nano don't publish a long tier; we use short
+    # for everything on those. OpenAI has no separate cache-write rate,
+    # so cache_write_1h mirrors cache_write on every OpenAI entry.
+    ("gpt-5.5-pro", ModelPricing(
```
shellygr, line 97: please add 5.6 variants too.

composer/pipeline/cli.py
```diff
@@ -62,6 +68,58 @@ def user_ns(*parts: str | tuple[str, ...]) -> tuple[str, ...]:
     return user_data_ns() + tuple(out)
 
 
+BUDGET_PHASES: tuple[str, ...] = tuple(PhaseBudget.__annotations__)
+
+
+class BudgetFile(BaseModel):
```
shellygr, line 84: @beiraoCertora for https://github.com/Certora/zeus-saas/issues/42

composer/spec/source/report/schema.py
```diff
@@ -171,4 +209,7 @@ class AutoProverReport(BaseModel):
     #: Formalization gaps — properties that exist but no rule formalizes (see the two gap types).
     skipped: list[SkippedClaim] = Field(default_factory=list)
     gave_up_components: list[GaveUpComponent] = Field(default_factory=list)
+    #: Components the run budget cut short — excluded from every table above, rendered as an
+    #: appendix.
+    curtailed_components: list[CurtailedComponent] = Field(default_factory=list)
```
shellygr, line 294: @beiraoCertora relevant for reporting in https://github.com/Certora/zeus-saas/issues/42

composer/spec/source/author.py
```diff
@@ -331,6 +339,41 @@ async def run(self) -> Command | str:
...
+_BUDGET_WRAPUP_MESSAGE = """
+<system-alert>
+You have almost exceeded the token cost budget for this task. Wrap up IMMEDIATELY;
+a partial spec is better than going over budget. Concretely:
+
+- The feedback and prover validation requirements on publishing have been lifted. You no
+  longer need approval from the feedback judge — ignore any pending or future feedback,
+  including a judge response saying it was terminated.
+- Do NOT start new prover runs or research.
+- Delete any rules/invariants that do not currently work.
```
shellygr, line 628: wdym by "work" - compiling? running? producing _any_ output verified/violated?

composer/spec/artifacts.py
```diff
@@ -25,6 +25,12 @@
 
 _log = logging.getLogger(__name__)
 
+#: Terminal suffix for artifacts persisted from a budget-curtailed generation. The
+#: suffix keeps the file inert — no conf can reference a ``.spec.unverified`` and
```
shellygr, line 29: is it synced over in the saas app?

jtoman: no, see integration-notes.md, and your comments to beirao above.

composer/spec/source/author.py
jtoman: left vague on purpose. We provide no guarantees about the output if we hit "wrap it up", so what the agent delivers as "works" is likewise left ambiguous.

shellygr (approved):

PR #93, opened 2026-07-21
LLM backend plugins

jtoman:
Instead of hardcoding "anthropic" and "openai", we generalize ProviderKind into an abstract class, one with two key dynamic dispatches that encapsulate the behavior we would case split over the two constants for: selecting the correct memory tool implementation and text/file document rendering. This new ABC is called ProviderServices, and is threaded through everywhere the old ProviderKind was.

Further, the actual providers are now auto-discovered and loaded via importlib.metadata entrypoints under `certora.autoprove.llm_provider`. These entrypoints are expected to resolve to an instance of ProviderSpec, which describes how to select which provider to use for a model name, and the provider itself. The provider is responsible for taking the model name and configuration options and producing a modelprovider; itself encapsulates the model "builder" (itself parameterized over caching/thinking behavior) and the "provider services", which is the ABC described above.

As a validation of this approach, the existing OpenAI and anthropic providers have been moved to be "built in plugins". In principle adding Kimi, Qwen, Deepseek, etc. should simply be a matter of implementing the various Provider apis. Common base classes/abstractions are provided for the anticipated common cases to cut down on boilerplate.

While doing this refactoring, took the opportunity to clean up some apis; including removing the (deprecated and very broken) get_checkpointer, the composer.workflow.providers shim, and other now unused and orphan llm builders.


composer/llm/anthropic.py
```diff
@@ -77,14 +79,43 @@ def matches(model: str) -> bool:
 
 # --- Files API uploader ----------------------------------------------------
 
+class AnthropicRenderer:
+    def text_block(self, text, *, with_cache: bool) -> dict:
+        to_ret : dict[str, Any] = {"type": "text", "text": text}
+        if with_cache:
+            to_ret["cache_control"] = {
+                "type": "ephemeral",
+                "ttl": "5m"
```
naftali-g: how does this work in the 1hour ttl cases?

composer/llm/api.py
naftali-g, line 1: ?

composer/llm/list_iter.py
```diff
@@ -0,0 +1,25 @@
+from dataclasses import dataclass, field
+
+class NoSuchElementError(RuntimeError):
+    pass
+
+
+@dataclass
+class ListIter[T]:
```
naftali-g, line 8: I think this should be somewhere else in the codebase, and the filename should for sure be changed

composer/llm/openai.py
```diff
@@ -121,18 +125,38 @@ def _reasoning_effort(thinking_tokens: int) -> Literal["low", "medium", "high"]:
...
-class OpenAIFileUploader(_UploaderBase):
+class OpenAIFileUploader(UploaderBase):
     """``FileUploader`` impl backed by OpenAI's Files API
     (``purpose="user_data"``)."""
 
     client: openai.AsyncOpenAI
     uploaded: dict[str, str] | None = None
     _seed_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
-    provider: ProviderKind = "openai"
+    provider: ContentRenderer = field(default_factory=OpenAIRenderer)
```
naftali-g: > 
```suggestion
    renderer: ContentRenderer = field(default_factory=OpenAIRenderer)
```

composer/llm/api.py
jtoman: what the.

"?" indeed. I'll delete it haha

composer/llm/anthropic.py
jtoman: it, uses a shorter cache. I don't really know how to plumb through the cache configuration. it's ... difficult.

naftali-g (approved):
Approving to unblock you

composer/llm/openai.py
jtoman: thanks. I hate how hard it is to have abstract properties...

PR #94, opened 2026-07-21
unbreak messages

jtoman:
For mysterious reasons, we don't allow # in the prover call message. We have added a "#{n}" to the messages. This didn't work and breaks everything.

Unbreaking the tests!


ericeil (approved):
Sure

PR #95, opened 2026-07-22
update anthropic models

naftali-g:


composer/input/types.py
```diff
@@ -130,15 +130,15 @@ class _ModelOptionsCommon(Protocol):
 
 class ModelOptions(_ModelOptionsCommon, Protocol):
     model: Annotated[str, Arg(
-        help="Model to use for code generation (default: {default})", default="claude-opus-4-6"
+        help="Model to use for code generation (default: {default})", default="claude-opus-4-8"
```
jtoman: we need to add a feature flag to the anthropic handler; after 4.6 (i.e. > 4.7) thinking summary blocks aren't included by default. They are handy for debugging at least, so its worth updating the builders to include said thinking blocks.

jtoman: Actually, we'll need to update the context token counts in graphcore too. But, come to think of it, we've hard-coded the limits of like ... 2 models (both anthropic). Given our shift towards more provider portability, it might make sense to push this backup to the builder level rather than trying to auto configure in graphcore.

composer/llm/provider.py
```diff
@@ -35,11 +35,29 @@ class ModelProvider(Protocol):
...
 
+# Compaction has to leave room for everything that rides on top of the prompt just measured: the
+# response, the thinking budget, and the next batch of tool results all have to fit the same
+# window. Half of it is a working figure, not a derived one.
+_PROMPT_TOKEN_SHARE = 0.5
+
+
+def compaction_threshold(context_window: int) -> int:
+    """The prompt-token budget to allow a model with a ``context_window``-token window."""
+    return int(context_window * _PROMPT_TOKEN_SHARE)
```
jtoman, line 92: wait, this is a huge reduction? we compact after reaching 50% of the context limit? am I reading this right? That seems *really* aggressive as a threshold.

composer/ui/message_renderer.py
```diff
@@ -89,15 +89,21 @@ class _ModelPricing:
 # Sources should be re-checked when new model families ship.
 _PRICING: list[tuple[str, _ModelPricing]] = [
     # ---- Anthropic ----
-    # claude-opus-4.5 / 4.6 / 4.7 share a rate card; older 4 / 4.1
-    # are pricier. Matching by prefix-of-prefix so "claude-opus-4-7"
-    # and "claude-opus-4-7-20260301" both hit the right entry.
+    # Fable is its own rate card, a tier above Opus.
+    ("claude-fable-5", _ModelPricing(short=_PriceTier(10.00, 50.00, 1.00, 12.50))),
```
jtoman, line 93: my eyes pop whenever I see these prices, good gravy

pyproject.toml
```diff
@@ -15,7 +15,7 @@ authors = [
 ]
 
 dependencies = [
-    "graphcore @ git+ssh://git@github.com/Certora/graphcore.git@8b9d320316a2cfbfa775da0a961501539a4124f0",
+    "graphcore @ git+ssh://git@github.com/Certora/graphcore.git@e9e47c7d9ab82455f4959921845c6f80fe06419e",
```
jtoman: you'll want this to track master on graphcore once your pr lands, right?

jtoman (approved):
approving with one substantive nit/objection which shouldn't block merging.

pyproject.toml
naftali-g: yup

composer/llm/provider.py
naftali-g: This is the same as we had before - 500,000 for the 1m context models.
I agree it's too conservative, but let's change it in a different PR at some point if needed.

PR #96, opened 2026-07-23
Ecosystem abstraction (for Solana support)

ericeil:
## PR 1 of 3 — Ecosystem abstraction (EVM + Solana front-half)

Part of the stacked split of the Crucible work (and preparatory for full Solana Prover support).
**Stack:** `master` → **`eric/ecosystem`** → `eric/rust` → `eric/crucible-app`.

Introduces a runtime **ecosystem** seam so the shared pipeline driver generalizes over the
analyzed domain instead of being hard-wired to Solidity — and lands the Solana *front half*
(analysis + property extraction) against it. **EVM behavior is unchanged** (the driver defaults
to `ecosystem=EVM`); Solana is added as a second ecosystem, exercised end-to-end by a null
(no-verifier) backend. The Solana verification backend (Crucible) is PR 3.

### The seam
- **`composer/pipeline/ecosystem.py`** (new) — `Ecosystem[App, Main, Unit]` + `Language` frozen
  dataclasses, the `EVM` / `SOLANA` bindings, and the `ECOSYSTEMS` registry. An ecosystem factors
  into a *language* facet (how the analyzed source is read — Solidity vs Rust) and a *chain* facet
  (model + prompts + unit split).
- **`composer/pipeline/core.py`** — `run_pipeline` takes an `ecosystem` (default `EVM`) and drives
  the front half through it (analyzed-model type, prompts, validation, `locate_main`, `units`).
  The per-unit loop iterates `ecosystem.units(main)` over any `FeatureUnit`.
- **`composer/spec/system_model.py`** — the ecosystem-agnostic `FeatureUnit` protocol; EVM stays
  bound to `ContractInstance` / `ContractComponentInstance` with byte-identical cache keys.
- **`prop_inference.py` / `system_analysis.py` / `cli.py` / `ptypes.py`** — threaded through to
  accept the ecosystem's prompts/units/`Main` generics.

### Typing
- `PreparedSystem.main` and the `PipelineBackend[..., U, Main]` generics carry `Main`/`FeatureUnit`
  instead of `Any`; `_batch_cache_key` is typed via a result-type witness.
- The one deliberately-erased boundary (`run_pipeline`'s `ecosystem: Ecosystem[Any, Any, Any]`) is
  documented with the invariance reasoning.
- `FeatureUnit.context_tag` / `feature_json` return `dict[str, object]`; dropped a no-op
  `@runtime_checkable`.

### Solana front half
- **`composer/spec/solana/model.py`** (new) — `SolanaApplication` (the standalone analog of
  `SourceApplication`): programs, instructions, account constraints, CPIs, signers; plus the
  `SolanaProgramInstance` / `SolanaInstructionInstance` index wrappers. Whole-program extraction
  (`units` → a singleton `[program]`).
- **`composer/spec/solana/null_backend.py`** (new) — a report-only backend that records extracted
  properties without verifying, so the front half can be gated without a prover.
- **`composer/templates/solana/*`** — Solana analysis + property prompts, and the `RUST` language's
  `rust/_failure_modes.j2` fragment they compose in.

### Shared prompt partials (EVM + Solana dedup)
The EVM and Solana prompts duplicated a lot of mechanical boilerplate (the iterative prior-rounds
block, quality-over-quantity / adaptive-thinking guidance, the architect Behavior/Tools block, the
analysis memory paragraph) — and the copies had drifted. Factored into
**`composer/templates/shared/*`** carrying the EVM canonical wording, with the one domain noun
(`component`/`program`) parameterized. Mechanical boilerplate unified; domain prose (backgrounds,
examples, failure modes) left per-file. **EVM prompts render byte-identical**; Solana re-converges
to canonical wording.

### Docs & tests
- **`ARCHITECTURE.md`** (new) — high-level system map, written around the ecosystem seam.
- **`docs/ecosystem-abstraction.md`** — describes the implemented seam (present tense).
- **`tests/test_null_solana_backend.py`** (new) — deterministic unit test of the null backend
  (no LLM / Postgres / prover).
- **`CLAUDE.md`** — repo rule against `from __future__ import annotations`.

_(The Rust application framework doc `application-abstraction.md` lives with PR 3, where that code
lands.)_


composer/templates/solana/analysis_prompt.j2
```diff
@@ -0,0 +1,70 @@
+# Background
+
+{% if sort == "greenfield" %}
+You have been provided a system/design document describing a Solana application. No
```
julian-certora: Perhaps I am wrong, but it seems like a variable capturing the language, rather than the word Solana, might make prompts like this reusable for Soroban too...

julian-certora:
similarly, files like `.../composer/templates/analyzer_system_prompt.j2` that hardwire Solidity might be amenable to addiing a language variable and then perhaps be reusable?

composer/spec/system_model.py
```diff
@@ -1,10 +1,48 @@
...
+        """Stable per-run index used in task ids."""
+        ...
+
+    def cache_material(self) -> str:
+        """Stable string identifying this unit, hashed into its per-unit cache key."""
+        ...
+
+    def context_tag(self) -> dict[str, object]:
+        """The tag persisted alongside this unit's workflow context."""
+        ...
```
jtoman, line 40: you almost certainly want `str, Any` to mean, well, "any type".

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+class Language:
+    """The language of the **code being analyzed** — a facet of the ecosystem, shared by every
+    chain whose programs are written in it (e.g. the ``rust`` facet is shared by Solana and
+    Soroban). It drives how the shared front half *reads* the target's source (fs-exclusion
+    pattern, code-explorer prompt, failure modes).
+
+    This is emphatically **not** the language the AutoProver backend is *implemented* in: a
+    backend implemented as a Rust wheel (:mod:`composer.rustapp`) may analyze Solidity
+    (``echoprover`` → EVM) or Rust (Crucible → Solana). The implementation language is not
+    associated with the ecosystem; only the analyzed-source language is.
```
jtoman: uh.... duh? idk if we need this.

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+    #: Typed over ``BaseApplication`` (not ``App``): the validator receives the produced model
+    #: and narrows internally (as ``_validate_connectivity`` does), and this keeps it assignable
+    #: to ``run_component_analysis``'s ``validate`` parameter without a contravariance clash.
+    validate_analysis: Callable[[BaseApplication, SolidityIdentifier | None], str | None]
+    #: Locate the target unit (the "main contract"/program) in the analyzed model.
+    locate_main: Callable[[App, SourceCode], Main]
+    #: Enumerate the units the extraction phase infers properties for — one batch per unit. EVM
+    #: returns one per component; a whole-program ecosystem (Solana) returns a singleton ``[main]``,
+    #: so all its invariants are inferred + formalized in a single harness + run
+    #: (docs/crucible-unit-granularity.md §3).
```
jtoman: ref not found

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+    system_model: type[App]
+    #: Prompts for the system-analysis agent.
+    analysis_prompts: PromptPair
+    #: Prompts for the per-component property-inference agent.
+    property_prompts: PromptPair
+    #: Connectivity/shape validation of the analyzed model (retry feedback on failure).
+    #: Typed over ``BaseApplication`` (not ``App``): the validator receives the produced model
+    #: and narrows internally (as ``_validate_connectivity`` does), and this keeps it assignable
+    #: to ``run_component_analysis``'s ``validate`` parameter without a contravariance clash.
+    validate_analysis: Callable[[BaseApplication, SolidityIdentifier | None], str | None]
```
jtoman: "SolidityIdentifier"

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+
+LanguageTag = Literal["solidity", "rust"]
+ChainTag = Literal["evm", "solana", "soroban"]
+
+
+@dataclass(frozen=True)
+class PromptPair:
+    """A (system prompt, initial prompt) template-name pair for one agent."""
+
+    system: str
```
jtoman: are these the prompt wholesale? or the `j2` name? If the latter, I'd much prefer these were typedtemplate instances.

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+EVM: Ecosystem[SourceApplication, ContractInstance, ContractComponentInstance] = Ecosystem(
+    name="evm",
+    language=SOLIDITY,
+    system_model=SourceApplication,
+    analysis_prompts=PromptPair(
+        "application_analysis_system.j2", "application_analysis_prompt.j2"
+    ),
+    property_prompts=PromptPair(
+        "property_analysis_system_prompt.j2", "property_analysis_prompt.j2"
+    ),
```
jtoman: oh dear. typedtemplates imo

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+# The Solana chain (RUST ⊕ solana)
+# ---------------------------------------------------------------------------
+
+
+def _solana_validate(app: BaseApplication, expected_main: SolidityIdentifier | None) -> str | None:
+    """Connectivity/shape validation for a ``SolanaApplication`` (retry feedback on failure).
+    Mirrors the EVM ``_validate_connectivity`` structure: unique program identifiers, unique
+    instruction slugs within a program, the expected main program present, CPI targets known."""
+    if not isinstance(app, SolanaApplication):
+        return None
```
jtoman: ??? In a validation function, saying "you handed me the wrong type is not an error" is a *wild* decision lol.

composer/pipeline/ecosystem.py
```diff
@@ -0,0 +1,292 @@
...
+            slug = slugify_filename(ins.name)
+            if slug in slug_origin:
+                errors.append(
+                    f"Instructions {slug_origin[slug]!r} and {ins.name!r} in {prog.name} "
+                    f"reduce to the same filename slug {slug!r}; give them more-distinct names."
+                )
+            slug_origin[slug] = ins.name
+            # CPI targets may be well-known external programs (SPL Token, System, …)
+            # that are not declared in the model; we do not flag those. A future
+            # policy can require known_programs | known_authorities | an allowlist.
```
jtoman: I'm probably not familiar enough with rust to understand what this is talking about and how it fits into the surrounding code?

composer/pipeline/core.py
```diff
@@ -40,6 +39,10 @@
 from composer.spec.source.report.schema import RuleName, ReportBackend
 from composer.spec.source.report import build as report_build
 from composer.spec.source.task_ids import SYSTEM_ANALYSIS_TASK_ID, REPORT_TASK_ID
+# The ecosystem seam supplies the domain-specific front half (analyzed model type, prompts,
+# analysis validation, unit enumeration). ``main_instance`` moved here too and is re-exported
+# so existing EVM backends keep doing ``from composer.pipeline.core import main_instance``.
```
jtoman: fix the backends, this just leads to import chasing and buys us nothing but technical debt

composer/pipeline/core.py
```diff
@@ -106,36 +119,35 @@ def artifact_store(self) -> ArtifactStore[A, FormT]: ...
...
+    return CacheKey(string_hash(c.cache_material()))
+
+
+def _batch_cache_key[FormT: BackendResult](
+    props: list[PropertyFormulation], result_type: type[FormT]
+) -> CacheKey[ComponentGroup, FormT]:
+    # The cache VALUE type can't be inferred from ``props`` (they don't carry the backend's
+    # result type), and ``CacheKey``/``WorkflowContext`` are invariant — so a bare bound
+    # (``CacheKey[ComponentGroup, BackendResult]``) won't assign to the caller's
+    # ``WorkflowContext[FormT]``, and a return-only TypeVar trips ``reportInvalidTypeVarUse``.
```
jtoman: you can just ignore that warning. one of my open prs does. If the only reason to pass through the result type is to get pyright to shut up about "invalid type var use", not worth it imo.

composer/pipeline/core.py
```diff
@@ -147,28 +159,45 @@ def formalize_task_id(idx: int) -> str:
...
-async def run_pipeline[P: enum.Enum, FormT: BackendResult, H, A: ArtifactIdentifier](
-    backend: PipelineBackend[P, FormT, H, A],
+async def run_pipeline[P: enum.Enum, FormT: BackendResult, H, A: ArtifactIdentifier, U: FeatureUnit, Main](
+    backend: PipelineBackend[P, FormT, H, A, U, Main],
     run: PipelineRun[P, H],
     *,
     interactive: bool = False,
     threat_model: Document | None = None,
     max_bug_rounds: int = 3,
+    ecosystem: Ecosystem[Any, Any, Any] = EVM,
```
jtoman: what the, `Any`? shouldn't these match the type parameters for `PipelineBackend`? Yeah you'll have to downcast at callers, but that's fine.

But with all that said, given the ecosystems is closed, you don't have to type it hetereogeneously; declare it as a typeddict:
```
class Ecosystems(TypedDict):
    evm: Ecosystem[...]
    ...
```

side steps all of this.

Also, having `EVM` be the default so the typechecker doesn't scream bloody murder if you pass in a Solana verification job but don't pass `Ecosystems[solana]` is a foot gun waiting to happen. Fix the upstream callers, preserving existing calling conventions by trading off type safety is the wrong choice.

composer/pipeline/core.py
```diff
@@ -147,28 +159,45 @@ def formalize_task_id(idx: int) -> str:
...
-                "Your output MUST contain an explicit contract instance with this solidity identifier.",
-                *spec.extra_input
-            ],
+            ty=ecosystem.system_model, child_ctxt=run.ctx.child(CacheKey(spec.analysis_key)),
+            input=source, env=run.env,
+            extra_input=[*ecosystem.analysis_extra_input(source), *spec.extra_input],
             expected_main_id=source.contract_name,
+            system_template=ecosystem.analysis_prompts.system,
+            initial_template=ecosystem.analysis_prompts.initial,
+            validate=ecosystem.validate_analysis,
```
jtoman, line 245: I still think this could be typed better. You're passing in the model type, so you should be able to ensure the types for validate are correct.

composer/pipeline/core.py
```diff
@@ -181,30 +210,36 @@ async def run_pipeline[P: enum.Enum, FormT: BackendResult, H, A: ArtifactIdentif
     #    this preserves the prover's autosetup ∥ bug-analysis overlap, generically.
     formalizer_task = asyncio.create_task(prepared.prepare_formalization(run))
 
-    batches = await _extract_all(
-        backend.analysis_spec.properties_key, 
-        prepared.main, backend.backend_guidance, run,
-        phases["extraction"], interactive, threat_model, max_bug_rounds
+    # Extraction yields ``FeatureUnit`` batches (the ecosystem is invariant in its unit type and
```
jtoman: I don't buy this even a little bit.

composer/pipeline/core.py
```diff
@@ -256,32 +291,41 @@ async def _run(batch: _Batch) -> ComponentOutcome[FormT]:
 
 async def _extract_all[P: enum.Enum, H](
     prop_key: str,
-    main: ContractInstance, backend_guidance: str, run: PipelineRun[P, H],
+    # ``main`` stays untyped here: this internal helper drives the type-erased
+    # ``Ecosystem[Any, Any, Any]`` (see run_pipeline), so there is nothing to tie it to.
```
jtoman: hogwash imo. fix `run_pipeline`, this `Any` unsafety isn't worth it

composer/spec/prop_inference.py
```diff
@@ -101,13 +101,14 @@ class RefinementState(MessagesState):
 
 
 def _get_initial_prompt(
-    context: ContractComponentInstance,
+    context: FeatureUnit,
     sort: Sort,
     prev_results: list[_AgentRoundResult],
     backend_guidance: str,
+    template: str = "property_analysis_prompt.j2",
```
jtoman: no. no no no.

composer/spec/prop_inference.py
```diff
@@ -101,13 +101,14 @@ class RefinementState(MessagesState):
...
 ) -> str:
     return load_jinja_template(
-        "property_analysis_prompt.j2",
+        template,
```
jtoman: this is about to be a typed template anyway. and a `ITypedTemplate[...]` as a parameter here instead of a bare name makes it *way* less likely someone is confused and writes `foo_property_analysis.j2` that references `ctxt` instead of `context.`

composer/spec/prop_inference.py
```diff
@@ -279,12 +280,14 @@ def validate(_state: Any, result: _AgentRoundResult) -> str | None:
...
     env: ServiceHost,
-    component: ContractComponentInstance,
+    component: FeatureUnit,
     front_matter_items: Sequence[str | dict],
     ctx: WorkflowContext[_AgentResult],
     round: int,
     prev: list[_AgentRoundResult],
     backend_guidance: str,
+    system_template: str = "property_analysis_system_prompt.j2",
+    initial_template: str = "property_analysis_prompt.j2",
```
jtoman: see advice elsewhere about hardcoding these names. I do not like it sam-i-am.

jtoman: I'm going to stop flagging them, I assume you can carry the lesson forward.

composer/spec/system_analysis.py
```diff
@@ -114,7 +122,7 @@ class AnalysisInput(RoughDraftState, FlowInput):
     def _validation_wrapper(
         _: Any, app: BaseApplication
```
jtoman: there is *zero* `app` couldn't be typed as `T`; I don't know why it isn't. this would let you strengthen the type of `validate` which should propagate nicely upward.

composer/templates/rust/_failure_modes.j2
```diff
@@ -0,0 +1,15 @@
+Rust language-level failure modes (shared by every Rust on-chain ecosystem — Solana, Soroban, …):
```
jtoman: I guess this is fine, but this really seems like something that should be in a threat model? Its ubiquity says probably not, but I'd call this "ecosystem_threat_model" to emphasize its role alongside the user provided threat_model.

composer/templates/solana/analysis_prompt.j2
```diff
@@ -0,0 +1,70 @@
+# Background
+
+{% if sort == "greenfield" %}
```
jtoman: I strongly suspect you can drop all the `greenfield` variants. I don't think we're ever planning on supporting natspec for Solana/soroban.

composer/templates/solana/analysis_prompt.j2
```diff
@@ -0,0 +1,70 @@
...
+You have been provided a system/design document describing a Solana application. No
+implementation exists yet, and the document may vary in technical rigor.
+{% else %}
+You have been provided a system/design document describing a Solana application, plus access to
+the Rust source (native and/or Anchor) that implements it.
+{% endif %}
```
jtoman: I'm fairly certain I wrote a wrapper that gracefully handles the case where the system doc is `None`. I'd urge you to adapt and then use that.

composer/templates/solana/analysis_prompt.j2
```diff
@@ -0,0 +1,70 @@
...
+- `accounts`: every account it takes in its accounts context. For each account give its `name`,
+  its `account_type` (e.g. `Signer`, `Account<Vault>`, `Program`, `SystemAccount`,
+  `UncheckedAccount`, a PDA of specific seeds), its `roles` (signer / writable / readonly / pda /
+  program / sysvar), and the `constraints` the program is responsible for enforcing on it
+  (signer/owner checks, `has_one`, `seeds`+`bump`, `address`, or documented invariants). If an
+  account has no enforced constraints, record an empty list — that is itself important signal.
+- `signers`: which accounts must sign (the authorities the instruction authenticates).
+- `cpis`: cross-program invocations it performs (target program + what it does / signer used).
+- `requirements`: the instruction's behavioral specification, stated as requirements ("The
+  instruction must ...").
```
jtoman, line 64: wait hang on. Are you SURE you want this level of granularity?

The Solidity pipeline pretty intentionally chose the (rather vague) notion of "Component" to sit between "a single entry point" and "the whole ass contract". If there are, say, 20 entry points that means we'll get *20* property inference passes, *20* CVLR authors etc. Further I think (but can't prove) that thinking about functional units, not individual entry points is the more useful framing for properties.

I'm curious why you chose Instructions (~ entry points) specifically over the same abstract notion of "component" from the solidity world. I'm perfectly open to being told the abstraction doesn't transplant cleanly to Solana smart contracts.

composer/templates/solana/instruction_context.j2
```diff
@@ -0,0 +1,52 @@
...
+{% endfor %}
+{% endif %}
+
+The `{{ context.program.name }}` program is part of a `{{ context.app.application_type }}` application.
+{% if context.program.account_types %}
+It owns / derives these account (state) types: {{ context.program.account_types | join(", ") }}.
+{% endif %}
+
+{% set siblings = [] %}
+{% for p in context.app.programs %}{% if p.program_identifier != context.program.program_identifier %}{% set _ = siblings.append(p) %}{% endif %}{% endfor %}
```
jtoman: what the. Just make this a computed property of whatever `p` is?

composer/templates/solana/program_context.j2
```diff
@@ -0,0 +1,21 @@
+This task targets whole-program security invariants of the program `{{ context.program.name }}`
+(`{{ context.program.program_identifier }}`), part of a `{{ context.app.application_type }}` application.
+
+{% if context.program.account_types %}
+The program owns / derives these account (state) types: {{ context.program.account_types | join(", ") }}.
+{% endif %}
+
+The program exposes these instructions (the actions a fuzzer will drive in random sequences):
```
jtoman: strange choice to "hardcode" fuzzer here. There must be some backend agnostic verbiage we can use.

composer/templates/solana/property_prompt.j2
```diff
@@ -0,0 +1,92 @@
...
+<task>
+Input: {% if sort != "greenfield" %}the Rust/Anchor implementation of a Solana program{% else %}a design document for a Solana program{% endif %}, and a description of its instructions.
+Output: a list of whole-program "security properties" a fuzzer can check after every action.
+
+Follow these steps exactly:
+
+## Step 1
+
+Analyze the provided {% if sort != "greenfield" %}implementation{% else %}design{% endif %} and formulate a set of security
+properties that are PLAUSIBLE for this program. Prefer **cross-instruction** properties — ones a
```
jtoman: but isn't this whole pass explicitly scoped to a single instruction? that's what seems weird to me.

jtoman (changes requested):

composer/pipeline/ecosystem.py
ericeil: This was a source of confusion early on in my interactions with Claude.  I'm still not sure which of us was actually confused. :)  I'll remove this part of the comment.

composer/pipeline/ecosystem.py
ericeil: Will use TypedTemplate

composer/pipeline/ecosystem.py
ericeil: Yeah, that needs to be renamed, but I didn't want to clutter up this PR with that.

composer/pipeline/ecosystem.py
ericeil: Apparently this is just here to narrow the type?  Thanks for pointing it out; will fix.

composer/pipeline/ecosystem.py
ericeil: Will rewrite the comment

composer/spec/system_model.py
ericeil: FWIW, Claude has an opinion on this 😄 

> Yes, mechanically it can be dict[str, Any] — every concrete value type is assignable into either, so nothing breaks. But they mean different things to the type checker, and object is the better choice here:
> 
> dict[str, object]: values can be anything, but once you have one, pyright only lets you do what's valid for any object — str(), ==, pass it along, put it in another container. Try tag[k] + 1 or tag[k].upper() and pyright flags it, because it doesn't know the value supports that.
> dict[str, Any]: values can be anything, and type checking is switched off for them. tag[k] + 1 or tag[k].upper() type-checks fine even though it might blow up at runtime, because Any is compatible with everything in both directions.
> Given context_tag()'s only consumer ([context.py:208](vscode-webview://0mneq2jo099915ebfimbo0t512qjnvvgtsgktlokgdldqdl1hesp/composer/spec/context.py#L208)) treats the dict as an opaque blob to hand to a store — never inspects or operates on individual values — object is the more accurate signal: "this is a heterogeneous bag of data, don't assume anything about its contents." Any would be strictly weaker typing for no benefit, and it's the kind of blanket type-checking bypass the project's conventions (favor precise types, avoid Any-shaped escape hatches) push against. I'd leave it as object.

composer/templates/rust/_failure_modes.j2
ericeil: "Threat model" isn't quite right either.  Maybe "vulnerability patterns?"

composer/templates/solana/analysis_prompt.j2
ericeil: Maybe.  I'm open to generalizing this as part of the Soroban work.  But let's make sure this is really what we want there, first.

composer/templates/solana/analysis_prompt.j2
ericeil: Makes sense

composer/templates/solana/property_prompt.j2
ericeil: Cleaning this up along with the rest of the "unit" confusion

composer/templates/solana/analysis_prompt.j2
ericeil: Let me know if the new template is not what you had in mind

composer/templates/solana/analysis_prompt.j2
ericeil: Switched to a component abstraction

composer/pipeline/ecosystem.py
ericeil: Actually it might not be too bad.  Let's see...

composer/templates/solana/analysis_prompt.j2
chandrakananandi: i am working off of this branch to create some templates for soroban to get started. I would say for now let's keep solana and soroban separate. If we want better abstractions later we can do that but since we are actively developing both Solana and soroban at the same time, these generalizations may be hard to do right away.

composer/pipeline/cli.py
```diff
@@ -274,7 +275,12 @@ async def cont[FormT: BackendResult, A: ArtifactIdentifier, U: FeatureUnit, Main
                     run=run,
                     interactive=args.interactive,
                     max_bug_rounds=args.max_bug_rounds,
-                    threat_model=threat_model
+                    threat_model=threat_model,
+                    # cli_pipeline is the EVM CLI entry point (Solidity identifiers throughout
+                    # above); App/Main/U are only generic here to admit FoundryBackend and
+                    # ProverBackend (same EVM triple, different FormT/A). Downcast to the
+                    # generic ecosystem parameter rather than widening run_pipeline back to Any.
+                    ecosystem=cast(Ecosystem[App, Main, U], EVM),
```
jtoman: this could be polymorphic before because we didn't need the U/App/Main types. Now that we do, fix them. I think both downstream users use this. Or, better yet, make this passed in/generated in a way that doesn't require this cast.

composer/spec/system_analysis.py
```diff
@@ -94,11 +108,11 @@ async def run_component_analysis[T: BaseApplication](
...
     extra_input: list[str | dict],
-    expected_main_id: SolidityIdentifier | None = None,
+    expected_main_id: SourceIdentifier | None = None,
     *,
-    system_template: str = "application_analysis_system.j2",
-    initial_template: str = "application_analysis_prompt.j2",
-    validate: Callable[[BaseApplication, SolidityIdentifier | None], str | None] = _validate_connectivity,
+    system_template: TypedTemplate[AnalysisPromptParams] = ANALYSIS_SYSTEM_TEMPLATE,
+    initial_template: TypedTemplate[AnalysisPromptParams] = ANALYSIS_INITIAL_TEMPLATE,
+    validate: Callable[[T, SourceIdentifier | None], str | None] = _validate_connectivity,
```
jtoman: I'm still confused why you would EVER want to allow the default prompts here or elsewhere. Ditto validate connectivity I guess.

composer/spec/system_model.py
```diff
@@ -1,11 +1,12 @@
 from dataclasses import dataclass
-from typing import Literal, Protocol
+from typing import Literal, Protocol, runtime_checkable
 from pydantic import BaseModel, Field
 from functools import cached_property
 from composer.spec.util import slugify_filename
 from .types import ComponentName, SolidityIdentifier, ContractName
 
 
+@runtime_checkable
```
jtoman: I don't actually see where you use this, necessary?

composer/pipeline/core.py
```diff
@@ -63,6 +61,23 @@ class Formalizer[FormT: BackendResult, U: FeatureUnit](ABC):
     formalized_type: type[FormT]
     backend_tag: ReportBackend
 
+    async def begin(self, jobs: list[BackendJob[U]], run: PipelineRun) -> None:
```
jtoman: make this a `Sequence` which has two nice effects: 1) the caller is discouraged by the type system from mutating it, and 2) it makes it covariant, which means you don't need the `cast` below.

composer/pipeline/core.py
```diff
@@ -210,26 +208,24 @@ async def run_pipeline[P: enum.Enum, FormT: BackendResult, H, A: ArtifactIdentif
...
+        prepared.main, backend.backend_guidance, run,
+        phases["extraction"], interactive, threat_model, max_bug_rounds, ecosystem)
     formalizer = await formalizer_task
     if not batches:
         raise ValueError("No properties extracted from any component.")
 
-    # 4. Per-component formalization. Caching is core-owned, keyed by the backend's result type.
+    # 4. Any shared artifact the units build on is authored HERE — once, from every unit's
+    #    properties — not lazily by whichever unit formalizes first (see ``Formalizer.begin``).
+    await formalizer.begin(cast("list[BackendJob[U]]", batches), run)
```
jtoman: super confused why your claude thinks you need to quote types here, I'm 99% certain you don't? But in any event, you shouldn't need this cast once you take my advice about Sequence.

composer/pipeline/core.py
```diff
@@ -63,6 +61,23 @@ class Formalizer[FormT: BackendResult, U: FeatureUnit](ABC):
...
+        from the union of every unit's properties (it is what makes them checkable), and it cannot
+        be authored in ``prepare_formalization`` because that runs concurrently with extraction, so
+        no properties exist yet. This is the only point where both are true: extraction is done, and
+        no unit has been formalized. Doing it inside ``formalize`` instead means the first unit to
+        arrive decides the shared artifact for all of them — harmless at one unit, silently wrong at
+        several.
+
+        The prover's peer (``invariants.spec``) is staged in ``prepare_formalization`` and needs
+        nothing here."""
+        return None
```
jtoman: I would much prefer this sequencing becomes explicit in the API. a type like `StagedFormalizer` that is returned from `prepare_formalization` which you can call `begin` on and have it return the actual `Formalizer`. If you wanted, you could have the return type of `formalize` below be `StagedFormalizer[FormT, U] | Formalizer[FormT, U]` and have the pipeline driver introspect types to determine whether to `begin` or go straight to `formalizing`.

jtoman (changes requested):
getting close

composer/spec/system_model.py
ericeil: At one point a test needed this, but it doesn't seem to be used now.  Will drop it.

composer/pipeline/cli.py
ericeil: Fixed

composer/spec/system_analysis.py
ericeil: It turned out there was a little more plumbing needed in the natspec code.  Done.

composer/pipeline/core.py
ericeil: Good idea

composer/pipeline/core.py
ericeil: Done

composer/pipeline/core.py
```diff
@@ -104,6 +93,36 @@ async def finalize(self, outcomes: list[ComponentOutcome[FormT, U]], run: Pipeli
...
+    then told to work within — harmless at one unit, silently wrong at several. ``begin`` sits
+    between the two, where extraction is done and no unit has been formalized.
+
+    A separate type rather than a hook on ``Formalizer`` so the types carry the ordering instead of
+    the driver's call sequence: ``begin`` *returns* the formalizer, so the shared artifact arrives as
+    a constructor argument and there is no window in which a formalizer exists without it — the same
+    property the rest of the phase chain has.
+
+    Most backends need none of this and return a ``Formalizer`` directly; the prover's shared peer
+    (``invariants.spec``) is staged in ``prepare_formalization``."""
```
jtoman, line 112: SHUT UP CLAUDE. Good *GRAVY* it loves to yap about why stuff exists.

jtoman (approved):

composer/pipeline/ecosystem.py
```diff
@@ -130,13 +130,12 @@ class Ecosystem[App: BaseApplication, Main, Unit: FeatureUnit]:
     #: backend-neutral grounds, and let a backend that wants coarser work aggregate in its
     #: ``Formalizer`` instead.
     units: Callable[[Main], list[Unit]]
-    #: The EVM view of a unit, for the pipeline plugins (``composer.pipeline.plugin_api``) —
-    #: whose hooks are typed on ``ContractComponentInstance``, so today they only have something
-    #: to say about EVM. EVM supplies the identity; an ecosystem with no EVM view supplies
-    #: ``None``, and ``run_pipeline`` skips its plugin phases (and leaves its unit cache keys
-    #: free of the plugin digest) rather than handing a hook a unit it cannot read. Widening the
-    #: plugin API to ``FeatureUnit`` is what retires this field.
-    plugin_unit: Callable[[Unit], ContractComponentInstance] | None
```
ericeil: This was a temporary workaround while doing the initial merge of the plugin change from master; this code is not in master.

composer/pipeline/core.py
```diff
@@ -362,19 +364,12 @@ async def _extract_all[P: enum.Enum, H, Main, U: FeatureUnit](
...
+    plugins: PluginPhaseManager[P, U],
 ) -> list[_Batch[U]]:
     prop_ctx = run.ctx.child(PROPERTIES_KEY(prop_key))
 
-    # Plugin hooks are typed on the EVM unit — see ``Ecosystem.plugin_unit``. An ecosystem with no
-    # EVM view of its units has nothing to hand them, so for it neither the plugin phases run nor
-    # the plugin digest applies (it would otherwise perturb every unit's cache key for plugins
-    # that never touched the unit).
-    to_plugin_unit = ecosystem.plugin_unit
-    plugin_digest = plugins.plugin_digest if to_plugin_unit is not None else None
```
ericeil: This was a temporary workaround while doing the initial merge of the plugin change from master; this code is not in master.

composer/pipeline/plugin_api.py
```diff
@@ -100,23 +113,75 @@ def _load_jinja_template(template_name: str, **kwargs: Any) -> str:
...
+class PipelinePluginLoader[U: FeatureUnit](ABC):
+    """Declares a plugin's :data:`PluginScope` and builds it.
+
+    Scope lives here rather than on the plugin so the host can skip a plugin that doesn't apply
+    *without* running ``initialize`` — which is where a plugin is documented to acquire its DB
+    pools and subprocesses.
+
+    Because ``scope`` and ``initialize`` share ``U``, a mismatch between the two is a type error at
+    the declaration site: a loader that says ``ForEcosystem(SolanaComponentInstance)`` cannot yield
+    a ``PipelinePlugin[ContractComponentInstance]``."""
```
jtoman: thanks for telling us how type systems work claude 🙄

composer/pipeline/plugins.py
```diff
@@ -79,12 +82,46 @@ async def run[T](
...
+        if ep.name in seen:
+            raise RuntimeError(f"Multiple plugins with name: {ep.name}, failing")
+        seen.add(ep.name)
+        loader = ep.load()
+        if not isinstance(loader, type) or not issubclass(loader, PipelinePluginLoader):
+            raise RuntimeError(f"Bad plugin declaration: {ep.name}: {ep.module}.{ep.attr} is not a PipelinePluginLoader")
+        yield ep.name, loader()
+
+
+def applicable_plugin_manifest(unit_type: type) -> list[str]:
```
jtoman: you could make this `[U: FeatureUnit](unit_type: type[U])` I think? serves nothing but documentation but worth it perhaps.

jtoman (commented):
honestly, way less painful than I thought it would be!

composer/pipeline/plugins.py
ericeil: Good idea

PR #101, opened 2026-07-24
autosetup: disable rule_sanity for the test run

jar-ben:
Our `sanity` rule we run in the _test run_ does the sanity check; applying the rule_sanity checks hence lead to work duplication. 

Instead of changing the .conf file, we pass the `--rule_sanity none` as an extra arg (i.e. on the command line) because the test .conf file is later adopted for building actual rules (which should be running the rule sanity checks).


certora_autosetup/autosetup/autosetup.py
```diff
@@ -800,11 +800,14 @@ async def prepare_contract_warmup() -> Optional[ProverJobSpec[Any]]:
                     "_test_run",
                     target_dir=internal_confs_dir,
                 )
+                # The test run exercises the generated sanity rule; rule_sanity's own
+                # checks would duplicate it, so they are turned off for this invocation
+                # only (via extra_args, leaving the conf's rule_sanity untouched).
```
naftali-g, line 805: I believe basic sanity is enabled by default in the prover anyway, so maybe we don't need anything in the conf?

jar-ben: Yes, that's how it is. We don't have it in the .conf because it is on by default. This PR turns it off for the "test .conf file run", i.e. for following rule:
```
rule sanity {
    env e;
    calldataarg args;
    method certoraF;
    certoraF(e, args);
    satisfy true, "sanity check failed";
}
```

It makes no sense to "double" check the sanity here.

naftali-g: ah ok, the comment seems to imply there's a `rule_sanity` field in the conf already, that's all. if you feel like it, maybe the comment can be reworded

naftali-g (approved):

PR #102, opened 2026-07-24
autosetup: don't inherit via_ir from the build system

jar-ben:
AutoProver previously copied the project's `via_ir` into the generated conf —
both the profile-level `via_ir = true` (`base.py`) and per-path
`compilation_restrictions` (`FoundryRestrictionManager`). That is faithful to
the project's deployment build but harmful to verification: **viaIR inlines
internal functions, so the autofinder can't detect them and CVL internal
summaries fail to attach.

viaIR is not dropped for good — the compilation workarounds already re-enable
it **per contract** when a base compile genuinely needs it
(`stack_too_deep_via_ir`, `via_ir_required_feature`) and disable it when the
solc version can't support it (`unsupported_solc_via_ir`). So viaIR is now
added strictly out of necessity, exactly as the optimizer setting already is.

### Changes
- `base.py`: `_apply_common_solc_settings` no longer emits `solc_via_ir`; logs that the setting is ignored.
- `foundry.py`: `apply_restrictions` no longer applies per-path via_ir (still applies `optimizer_runs`); removed the now-unused `_via_ir_map` / `_merge_via_ir_maps` helpers and the `default_via_ir` constructor parameter; docstring updated.

### Testing
- AutoProver-local unit tests green (`test_compilation_workarounds`, `test_compiler_flag_reconciliation`, `test_foundry_runner`, `test_foundry_entry_args`, `test_hardhat_config` — 43 passed).


naftali-g (changes requested):
Did you check this actually works when a via-ir is needed for a single contract? because from what I see, the `_apply_via_ir_workaround` just adds a single contract to the map, but iirc if you have the map field in the conf it _must_ contain all contracts of the scene

jar-ben:
> Did you check this actually works when a via-ir is needed for a single contract? because from what I see, the `_apply_via_ir_workaround` just adds a single contract to the map, but iirc if you have the map field in the conf it _must_ contain all contracts of the scene

We seed the map before the workarounds are applied, see https://github.com/Certora/AutoProver/blob/2b17d28298838a55c6acbadde997363b55802f6e/certora_autosetup/utils/compilation_workarounds.py#L440

naftali-g:
> > Did you check this actually works when a via-ir is needed for a single contract? because from what I see, the `_apply_via_ir_workaround` just adds a single contract to the map, but iirc if you have the map field in the conf it _must_ contain all contracts of the scene
> 
> We seed the map before the workarounds are applied, see
> 
> https://github.com/Certora/AutoProver/blob/2b17d28298838a55c6acbadde997363b55802f6e/certora_autosetup/utils/compilation_workarounds.py#L440

got it

naftali-g (approved):

PR #107, opened 2026-07-27
compilation workarounds: detect stack-too-deep in whole-project solc errors

jar-ben:
`_detect_stack_too_deep_errors` only recognized the per-file autofinder format — a
`Compiling <path>...` progress line followed by the error. A whole-project solc error
(`solcX had an error:` / `CompilerError: Stack too deep` / `   --> <path>:line:col`),
which has no `Compiling` line, was never mapped to a contract, so the
`stack_too_deep_via_ir` workaround didn't fire and compilation aborted with "no
applicable workaround".

This matters now that #102 stopped inheriting `via_ir` from the build system: a
project that genuinely needs `via_ir` to compile must have it re-added on demand by
this workaround rather than getting it from the build config.

## Changes
- Add `_path_from_source_location_line` (+ `_SOURCE_LOCATION_RE`) to recover the file
  from a solc `--> <path>:line:col` source-location line.
- Add Pattern 3 to `_detect_stack_too_deep_errors`: when no `Compiling` line is
  present, map the error to a contract via the `-->` line. Runs after Patterns 1/2 so
  the per-contract format keeps precedence.

## Testing
- 3 new unit tests: bulk-format detection, unmapped path → `None`, and
  `Compiling`-line precedence over the `-->` fallback.
- Full `tests/test_compilation_workarounds.py` green (31 passed).
- Validated end-to-end: a project that previously failed compilation analysis with
  "no applicable workaround" now detects stack-too-deep, re-enables `via_ir`, and
  proceeds past the compilation phase.


naftali-g (approved):
lgtm

PR #110, opened 2026-07-28
autosetup: match curated library summaries across the whole scene

jar-ben:
The curated `CTHelpers.getCollectionId` summary was registered but never attached:
`match_summaries_from_all_methods` scoped candidates to
`get_methods_by_originating_contract(main_contract)`, and the library's single
`originatingContract` need not be any summarized contract — so the entry was
invisible and the scene-optimization phase kept OOM-ing on the un-summarized
arithmetic.

## Fix

For `library_names` entries, add every `isLibrary` method in the scene to the
candidate set (unioned with the originating-scoped methods):

```python
library_scene_methods = [m for m in mp.get_all_methods() if m.get("isLibrary")]
candidate_methods = (
    contract_methods + library_scene_methods if func_info.get("library_names") else contract_methods
)
```

## Tests

`tests/test_curated_library_scene_matching.py`: a foreign-originating library still
matches; a non-library base from a foreign unit does not match scene-wide but does
when it originates in scope (originating-scoped behavior preserved); and the shipped
`ctHelpers_getCollectionId` entry stays library-scoped.

### DIFF
--- a/certora_autosetup/setup/setup_summaries.py
+++ b/certora_autosetup/setup/setup_summaries.py
@@ -2383,15 +2383,25 @@ Method signature: {method_signature}
         self.log(f"Matching summaries for {main_contract}...")
         # Filter to only methods that originate from this contract (compilation unit)
         contract_methods = mp.get_methods_by_originating_contract(main_contract)
+        # For a `library_names` entry, also consider every library method in the scene. A library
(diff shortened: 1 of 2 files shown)


shellygr (approved):
seems fine to me

PR #111, opened 2026-07-29
autosetup: record all originating contracts per method (originatingContracts)

jar-ben:
## Summary

Two related fixes to per-compilation-unit handling.

**1. Record all originating contracts per method.** `all_methods.json` kept a single,
first-seen `originatingContract` per method (the field was excluded from the dedup key, so
repeats were dropped). A method inlined into several units — e.g. a shared library — kept only
one arbitrary unit, so any consumer scoping to a contract's compilation unit missed it for the
others. Replace it with **`originatingContracts`** (every unit that inlines the method) and match
by **membership**; drop the singular field.

Applied to all four readers:
1. `setup_prover.py` — dedup keyed into a dict; repeats union their unit into `originatingContracts`.
2. `method_parser.get_methods_by_originating_contract` — membership test.
3. `setup_summaries.py` — matcher + `analyze_with_llm` scoping use membership (removing the
   scene-wide `isLibrary` union).
4. `solidity_utils.py` — used-libraries scan uses membership.

For curated matching this both attaches a library summary for a contract whose unit inlines the
library (previously missed) and avoids matching every scene library onto every contract.

**2. Reconcile compiler maps for warmup libraries.** The warmup path wrote the resolved library
files into `files=` with a plain `update_config_with_properties`, which trims
`solc_via_ir_map` / `compiler_map` / `solc_optimize_map` to the contracts left in `files=` —
leaving the just-added libraries in `files=` but absent from the maps, so certoraRun rejected them
as *not matched in `solc_via_ir_map`*. Route the extra library handles through
`add_files_to_config` so their map entries are filled. (Main + `--additional-contracts` already get
theirs from `create_base_config`, so they stay on the plain write.)

## Tests

- `tests/test_curated_library_scene_matching.py` — drives `match_summaries_from_all_methods` over an
  in-memory `all_methods.json`: library in the unit → matched; not in the unit → not matched;
  per-unit isolation across siblings; shipped registry entry stays library-scoped. Non-vacuous by
  mutation (old singular logic fails the inclusion cases).
- `tests/test_compiler_flag_reconciliation.py` — adds `update_via_ir_map_for_contract` coverage
  mirroring the existing `solc_optimize_map` tests.

Reproduced the warmup fix end to end on a real project (pinned commit): the map is trimmed to 1
entry, the two library files are re-added to `solc_via_ir_map`, and `certoraRun
--compilation_steps_only` (the failing step) passes.

### DIFF
--- a/certora_autosetup/parsers/method_parser.py
+++ b/certora_autosetup/parsers/method_parser.py
@@ -40,7 +40,7 @@ class MethodParser:
 
     def get_methods_by_originating_contract(self, contract_name: str) -> List[Dict]:
         """Get all methods whose compilation unit includes ``contract_name``."""
-        return [m for m in self.methods if contract_name in m.get('originatingContracts', [])]
+        return [m for m in self.methods if contract_name in m['originatingContracts']]
 
     def get_methods_by_name(self, method_name: str) -> List[Dict]:
         """Get all methods with a specific name across all contracts."""
--- a/certora_autosetup/parsers/method_parser.py
+++ b/certora_autosetup/parsers/method_parser.py
@@ -39,8 +39,8 @@ class MethodParser:
         return [m for m in self.methods if m['contractName'] == contract_name]
 
     def get_methods_by_originating_contract(self, contract_name: str) -> List[Dict]:
-        """Get all methods that originate from a specific contract (compilation unit)."""
-        return [m for m in self.methods if m.get('originatingContract') == contract_name]
+        """Get all methods whose compilation unit includes ``contract_name``."""
+        return [m for m in self.methods if contract_name in m.get('originatingContracts', [])]
 
     def get_methods_by_name(self, method_name: str) -> List[Dict]:
         """Get all methods with a specific name across all contracts."""
--- a/certora_autosetup/setup/setup_summaries.py
+++ b/certora_autosetup/setup/setup_summaries.py
@@ -1700,7 +1700,7 @@ Method signature: {method_signature}
             contract_files: Set of Solidity files to analyze
             methods_to_skip: Set of (contract, method) pairs to skip (already marked for
                 summarization by non-LLM step or matched by previous LLM recipes)
-            main_contract: Optional main contract name to filter methods by originatingContract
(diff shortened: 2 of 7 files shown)


certora_autosetup/parsers/method_parser.py
```diff
@@ -39,8 +39,8 @@ def get_methods_by_contract(self, contract_name: str) -> List[Dict]:
         return [m for m in self.methods if m['contractName'] == contract_name]
 
     def get_methods_by_originating_contract(self, contract_name: str) -> List[Dict]:
-        """Get all methods that originate from a specific contract (compilation unit)."""
-        return [m for m in self.methods if m.get('originatingContract') == contract_name]
+        """Get all methods whose compilation unit includes ``contract_name``."""
+        return [m for m in self.methods if contract_name in m.get('originatingContracts', [])]
```
naftali-g: We know `m` has the `originatingContracts` key. Just use square brackets instead of the `get`.

certora_autosetup/setup/setup_summaries.py
```diff
@@ -1719,13 +1719,13 @@ async def analyze_with_llm(
...
-        # Filter methods by properties and originatingContract
+        # Filter methods by properties and compilation-unit membership
         all_methods = mp.get_all_methods()
         filtered_methods = []
 
         for method in all_methods:
-            # Filter by originating contract if main_contract is specified
-            if main_contract and method.get("originatingContract") != main_contract:
+            # Filter to the main contract's compilation unit if specified
+            if main_contract and main_contract not in method.get("originatingContracts", []):
```
naftali-g: ditto

certora_autosetup/setup/solidity_utils.py
```diff
@@ -392,7 +392,7 @@ def find_libraries_used_by(
     """
     used: set[str] = set()
     for method in methods:
-        if method.get("originatingContract") != contract_name:
+        if contract_name not in method.get("originatingContracts", []):
```
naftali-g: Here too...

naftali-g (approved):
Just a nit on Claude's overly defensive programming

PR #114, opened 2026-07-31
Anchor build-system detection on the contract, not the run root

shellygr:
## Problem

Autosetup resolved the build system from the process CWD, and
`BuildSystemManager.find_config_file` only walks *upward* from there. A repository whose
Foundry/Hardhat project lives in a subdirectory (`<pkg>/foundry.toml`) is therefore invisible
when the run root is the repo root:

- `BuildSystemDetector` reports `unknown` → *"No build system detected, continuing without build config"*
- the generated conf carries neither the project's remappings nor its pinned solc/evm_version
- solc fails with a bare `ParserError: Source "@pkg/Foo.sol" not found. Searched the following locations: "".`

The reactive `source_not_found_packages` workaround could not rescue it either:
`_build_packages_from_remapping_sources` hardcoded `base_dir=Path(".")`, so it read the same
root that had nothing to offer, recomputed an **identical** packages list, and the retry guard
correctly gave up with *"Workarounds applied but the conf and command are unchanged"*. The run
died with `CompilationAnalysisError` before ever reaching the Prover.

## Fix

Anchor on the contract under analysis instead of on the run root. `find_build_config_dir`
walks up from the main contract's own path to the nearest ancestor holding a `foundry.toml` /
`hardhat.config.{js,ts}`, bounded by the run root, and that directory is used for:

- `BuildSystemDetector.resolve(...)`
- the `BuildSystemManager` instance (so `find_config_file` starts in the right place)
- `base_dir` for remapping sources in the reactive workaround

No new plumbing or arguments are needed — the main contract handle is already set before
`setup_build_system_config()` runs.

### Why this is safe for existing projects

- A root-level project resolves to the run root, so behaviour is unchanged.
- `build_packages_from_remapping_sources` already resolves relative targets **absolute**
  against `base_dir`, so the emitted packages remain valid from the run CWD.
- `CompilationWorkaroundManager.project_root` is untouched and keeps anchoring the conf and
  generated harnesses at the run root; the new `build_config_dir` is a separate concept and
  defaults to `project_root`, so the `fixconf` call site needs no change.

## Tests

- `tests/test_project_dir.py` — nested project, root-relative contract path, root-level
  project, no config anywhere, nearest-ancestor-wins over an outer config, hardhat anchor,
  walk stops at the run root, contract outside the root.
- `tests/test_compilation_workarounds.py` — the reactive workaround now picks up a nested
  `foundry.toml` remapping (the exact case that produced the identical-conf give-up), plus a
  default-to-`project_root` case.

Full suite: 271 passed, 10 pre-existing failures identical on `master` (missing `tabulate` /
`certora_cli` in the test env), and 6 pre-existing collection errors from the same missing
dependency.

### DIFF
--- a/certora_autosetup/utils/project_dir.py
+++ b/certora_autosetup/utils/project_dir.py
@@ -12,6 +12,7 @@ Anchoring on the main contract's own path fixes this without any new plumbing: t
 that owns a contract is the nearest ancestor of that contract holding a build config.
 """
 
+import os
 from pathlib import Path
 from typing import Optional
 
@@ -52,6 +53,22 @@ def find_build_config_dir(contract_path: Path, root: Path) -> Path:
         if current == root:
             return root
         current = current.parent
+
+
+def rebase(rel_path: str, from_dir: Path, to_dir: Path) -> str:
+    """Reinterpret *rel_path* — given relative to *from_dir* — as relative to *to_dir*.
+
+    Pure path arithmetic; nothing is required to exist on disk. Absolute inputs are
+    returned unchanged, since they need no anchor.
+
+    Build systems record source paths relative to their own project dir (Foundry's
+    ``compilationTarget``, for one), so a nested project hands back ``src/Foo.sol``
+    while the process CWD is the repo root and needs ``pkg/portal/src/Foo.sol``.
+    """
+    if Path(rel_path).is_absolute():
+        return rel_path
+    absolute = (from_dir / rel_path).resolve()
+    return os.path.relpath(absolute, to_dir.resolve())
 
 
 def describe_build_config_dir(build_config_dir: Path, root: Path) -> Optional[str]:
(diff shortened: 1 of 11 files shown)


certora_autosetup/utils/project_dir.py
```diff
@@ -0,0 +1,85 @@
+"""Locate the build-system project directory that owns the contract under analysis.
+
+A repository's Foundry/Hardhat project is not always at the repo root: monorepos keep the
+build config next to the sources it governs (``<repo>/<package>/foundry.toml``). Autosetup
+runs with CWD at the repo root, and ``BuildSystemManager.find_config_file`` only walks
+*upward* from there, so a nested project is invisible to it — detection reports ``unknown``,
```
jar-ben: The comment is describing issues we are fixing. Especially I don't like:
> ``ParserError: Source "@pkg/Foo.sol" not found ... Searched the following locations: ""``

Just describe what the code does and perhaps why, but don't reference issues of the past codebase.

jar-ben (approved):

certora_autosetup/utils/project_dir.py
shellygr: I can't get rid of that pesky behavior

PR #115, opened 2026-07-31
Extended RAG, CVL context

jtoman:
Our attempts to make all CVL knowledge RAG driven has been yielding mixed results. This attempts to condense some of the most common and important information about CVL into "context documents" which are inlined into every CVL agent's task. Explicit cache markers are added to make this avoid prompt blow ups.

The context documents incorporate (and moot) several existing artifacts:
1. The store-based KB. All of the advice in the KB has now been migrated to the `recipes` index (with some extra information as well beyond what was in kb_populate)
2. `cvl_guidance.j2` has been expanded and updated and converted to `cvl_baseline.md`.
3. "summarization" rag and "invariants & quantifiers" rag attempt to distill the core knowledge behind these key techniques.

In addition, we've tried our hand at CoNsTiTuTiOnAl Ai: we have pretty consistently viewed agents treating "get the spec all verified" as the end goal. The "verification_charter.j2" is our first attempt at making the goal of the agent explicit: verified software is the goal, not a verified spec. The feedback judge's criteria are now AlIgNeD with this constitutional principle; most of the evaluations we ask of it are grounded in this principle.

The feedback prompt has also been expanded to close one common failure mode; a property which ends up passing but skips all of the actually important methods.


composer/kb/resources/recipes/mocks-and-edits.md
```diff
@@ -0,0 +1,26 @@
+> Where this recipe and the CVL manual or other Certora materials diverge, this recipe governs.
+
+# 6. Mocks and source edits `[EDIT]`
+
+The governing axis: **CVL-expressibility.** Behavior expressible as a CVL
+model — summaries, ghosts, hooks, axioms — is modeled in CVL. A mock or
+source edit is reserved for behavior that must exist at EVM level for the
+Prover to use it:
+
+- **Storage layout declarations** — the R2 harness (declarative fields only).
```
shellygr, line 10: ~R2?~ okay this is parseable after seeing the whole thing

composer/kb/resources/recipes/r06-immutables.md
```diff
@@ -0,0 +1,22 @@
...
+rule ownerNeverChanges(env e, method f, calldataarg args) {
+    address before = currentContract.OWNER;
+    f(e, args);
+    assert before == currentContract.OWNER;
+}
+```
+
+---
+
+# 2. Delegation patterns
```
shellygr: ??

composer/kb/resources/recipes/r07-contract-extensions.md
```diff
@@ -0,0 +1,35 @@
...
+        { "extension": "Extension2", "exclude": ["someFunction"] }
+    ]
+}
+```
+
+`exclude` lists extension functions to leave out of the routing.
+
+---
+
+# 3. EVM-level facts
```
shellygr: ??

composer/kb/resources/recipes/r10-nativecodesize.md
```diff
@@ -0,0 +1,8 @@
+> Where this recipe and the CVL manual or other Certora materials diverge, this recipe governs.
+
+### R10. EOA vs contract `[CVL]`
+
+**Trigger:** a property distinguishing externally-owned accounts from
+contracts.
+
+**Formula:** `nativeCodesize[addr] == 0` characterizes an EOA.
```
shellygr, line 8: I am not sure what happens if you use `nativeCodesize` within a `preserved constructor` on the constructed address. technically it's still 0...
wondering if there are other gotchas here, e.g. dynamic contracts

composer/kb/resources/recipes/r11-ecrecover-axioms.md
```diff
@@ -0,0 +1,37 @@
...
+    ...
+}
+```
+
+Each axiom is an assumption about the cryptography; the pack is the standard,
+stated form of that assumption.
+
+---
+
+# 4. Quantified invariants
```
shellygr: ???

composer/kb/resources/recipes/r12-ghost-mirror-quantified-invariants.md
```diff
@@ -0,0 +1,62 @@
+> Where this recipe and the CVL manual or other Certora materials diverge, this recipe governs.
+
+### R12. The mirror pattern `[CVL]`
+
+**Trigger:** an invariant needs to quantify over the contents of an array,
+mapping, or set — and quantifier bodies can contain no storage accesses, no
+contract calls, and no `require_*`/`assert_*` casts.
+
+For *unquantified* reads of contract state, direct storage access is preferred
+and no mirroring is needed; this pattern earns its cost exactly because
```
shellygr: "this pattern earns its cost" very llm-ish . good that it isn't load bearing. can we humanize a bit: "this pattern is relevant because direct storage access cannot be used within a quantified expression"

composer/kb/resources/recipes/r12-ghost-mirror-quantified-invariants.md
```diff
@@ -0,0 +1,62 @@
...
+instantiations — when the invariant body must call view functions or access
+storage (both barred under quantifiers), or when grounding fails on the
+quantified form. The fallback's discipline, on this example: two invariants
+`A(mathint i)` / `B(bytes32 v)`; only `remove` needs `preserved` blocks,
+carrying `requireInvariant A(ghostLength - 1)` (the swap-and-pop relocation
+witness) and `requireInvariant B(w)`. Declare the index parameter `mathint`:
+at length zero the instantiation `A(-1)` is vacuously true, where a
+`require_uint256(ghostLength - 1)` cast would silently prune the empty-set
+executions instead. The fallback's cost is exactly that every needed
+instantiation must be identified by hand and maintained across refactors.
```
shellygr: uhhh can't we spell it out more clearly with an example?

composer/kb/resources/recipes/r15-preserved-blocks.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+The generic (unnamed) `preserved` block applies only to the induction step; an
+invariant failing on the base case takes its assumptions from a dedicated
+`preserved constructor()` block.
+
+A `preserved` block adds prestate requirements *on top of* the invariant being
+proved: the check is `assume I; m(); assert I`, with the block's contents
+inserted after the `assume I`. The invariant itself is always assumed —
+requiring `I` manually in its own preserved block is redundant.
```
shellygr: is there a place where we give guidance on what kinds of requires are okay within preserved? I imagine something like `e.msg.sender != currentContract` is a reasonable pattern (if checked otherwise to be unrealistic)

composer/kb/resources/recipes/r15-preserved-blocks.md
```diff
@@ -0,0 +1,28 @@
...
+    preserved constructor() { ... }                     // constructor base case
+}
+```
+
+The generic (unnamed) `preserved` block applies only to the induction step; an
+invariant failing on the base case takes its assumptions from a dedicated
+`preserved constructor()` block.
+
+A `preserved` block adds prestate requirements *on top of* the invariant being
+proved: the check is `assume I; m(); assert I`, with the block's contents
```
shellygr: assume? not require?

composer/kb/resources/recipes/r18-reentrancy-recipes.md
```diff
@@ -0,0 +1,73 @@
...
+       env e;
+       if (g_sighash == sig:withdrawAll().selector) {
+           withdrawAll@withrevert(e);
+           g_reverted = lastReverted;
+       } else if (g_sighash == sig:withdraw(uint256).selector) {
+           calldataarg args;
+           withdraw@withrevert(e, args);
+           g_reverted = lastReverted;
+       } else {
+           g_reverted = true;
```
shellygr, line 35: kinda confusing. g was never called....

composer/kb/resources/recipes/r18-reentrancy-recipes.md
```diff
@@ -0,0 +1,73 @@
...
+       assert called_extcall => g_reverted;
+   }
+   ```
+3. **"Storage accesses don't straddle external calls"** (guard-free safety) —
+   classify every storage access as before/after the first external call and
+   assert the combination never occurs:
+   ```cvl
+   persistent ghost bool called_extcall;
+   persistent ghost bool access_before_call;
+   persistent ghost bool access_after_call;
```
shellygr, line 54: you don't want to use a mapping from loc to value? this is quite coarse grained (very safe)

composer/kb/resources/recipes/r19-address-dispatch.md
```diff
@@ -0,0 +1,19 @@
...
+different dispatch:
+
+- With a `calldataarg`: dispatches over every provided contract that has a
+  function named `foo`.
+- With typed arguments: dispatches over every method in any provided contract
+  whose name and declared parameter types match.
+
+The dispatch is a chain over `x == <ProvidedContract>.address` whose final
+branch is `assume false`: **the address is assumed to be one of the provided
+contracts.** This is a closed-world assumption; any rule using address-calls
```
shellygr, line 18: if there are no `foo`s to match it fails typechecker. worth noting here because we were already bit by it in the SafeERC20 summaries

composer/kb/resources/cvl_baseline_facts.md
```diff
@@ -0,0 +1,76 @@
...
+  reverted; it is rarely used.
+- In CVL, `^` is exponentiation; bitwise exclusive-or is the keyword `xor`.
+
+## Hooks
+
+- Hooks fire only on contract-code storage accesses; reads and writes made
+  from CVL, including direct storage access, do not trigger hooks.
+- Hooks do not fire recursively: a storage access performed inside a hook
+  body does not trigger the hook again.
+- Hooks do not fire during a `reset_storage` command; ghost mirrors are stale
```
shellygr, line 69: they also do not fire upon reverts

composer/kb/resources/cvl_invariants_quantifiers.md
```diff
@@ -0,0 +1,284 @@
...
+> this document governs.
+
+Owns all invariant and quantifier knowledge. The summarization knowledge base,
+the baseline facts, and the recipes index defer to this document on these
+topics; recipes-index entries marked `see` point at sections here.
+
+## 1. The proof shape: induction
+
+Proving invariant `I` checks, for every method `m` of every provided contract:
+`assume I; m(); assert I` — plus a base case in which the constructor runs
```
shellygr: require?

composer/kb/resources/cvl_invariants_quantifiers.md
```diff
@@ -0,0 +1,284 @@
...
+
+The routes prove the same thing: importing `Q` into `I`'s preserved blocks
+and `I` into `Q`'s is exactly proving `I && Q` inductive, so mutual imports
+among proved invariants are sound. The recurring special case: an invariant
+violated only from a start state the contract can never reach is not
+inductive, and the strengthening conjoins the reachable-configuration facts —
+often a disjunction of the legal states — that exclude it.
+
+**`preserved` blocks.** A `preserved` block inserts *additional* prestate
+requirements after the `assume I`. The invariant itself is always assumed;
```
shellygr: assume is standard, but I think it's better to stick to our require

composer/kb/resources/cvl_invariants_quantifiers.md
```diff
@@ -0,0 +1,284 @@
...
+```cvl
+forall uint i. forall uint j. i <= j => a[i] <= a[j]
+```
+
+Sortedness is the canonical case; the general form of the rule: if the
+property's use sites need `P(x, z)` while the natural statement gives
+`P(x, y)` and `P(y, z)`, quantify the closed relation, not the generating
+step.
+
+## 7. Quantifier bodies, and the mirror pattern
```
shellygr, line 183: are the duplications with the R* docs on purpose?

composer/kb/resources/cvl_invariants_quantifiers.md
```diff
@@ -0,0 +1,284 @@
...
+Both quantified conjuncts follow §6's rules: variables index directly
+(`ghostValues[i]`, `ghostIndexes[v]`), and each conjunct is stated so its use
+sites need one instance. (The `ghostIndexes[v] - 1` inside the *second*
+conjunct sits in an access position — tolerated here because the matching
+term for that conjunct is `ghostIndexes[v]` itself; when such a formula
+resists grounding, restating or the §5 fallback applies.) Ghosts mirroring
+storage-like state stay regular, not `persistent`, so they havoc and revert
+when the storage they mirror does.
+
+## 8. Properties beyond first-order: closure relations
```
shellygr, line 230: Nice 🔥

composer/kb/resources/cvl_summarization_rag_draft.md
shellygr, line 1: do we not want to include https://docs.certora.com/en/latest/docs/prover/approx/unresolved-harness.html anywhere? it's the last-resort option for unresolvable calls, including e.g. precompiled contracts with non-ABI convention

composer/kb/resources/cvl_summarization_rag_draft.md
```diff
@@ -0,0 +1,925 @@
...
+
+## B2. When to summarize — and when not to
+
+### B2.1 The governing question
+
+A summary trades fidelity for tractability, so the decision is always relative
+to a property: *is the behavior about to be erased load-bearing for this
+property, and does the replacement preserve it in the sound direction?* There
+is no property-independent answer to "should `f` be summarized" — and the
+methods block is **spec-global**: a summary that is benign for rule A silently
```
shellygr:
```suggestion
methods block is **spec-global**: a summary that is benign for rule A 
```

composer/kb/resources/cvl_summarization_rag_draft.md
```diff
@@ -0,0 +1,925 @@
...
+   callee (A3) if it is unresolved, or leave it unsummarized. Every summary is
+   an assumption someone must maintain, and it rots: the contract evolves, the
+   summary doesn't, and nothing fails when they diverge.
+3. *A faithful summary would replicate the callee.* If soundness for the
+   property requires most of the callee's complexity, no tractability was
+   gained and a divergence channel was added — verify the real thing, or do
+   decomposition properly (B2.2 case 3, with the callee-side proof).
+4. *The justification comment can't be written.* The comment requirement (B1)
+   is the diagnostic: inability to articulate what the summary assumes means it
+   is doing load-bearing work nobody understands.
```
shellygr, line 740: 🥂

composer/kb/resources/cvl_summarization_rag_draft.md
```diff
@@ -0,0 +1,925 @@
...
+
+### B2.4 Summaries of the verified contract's own surface rewrite the theorem
+
+`DELETE` and aggressive summarization of `currentContract`'s own methods change
+what "for all methods `f`" ranges over in parametric rules and what an
```
shellygr, line 745: hmm? if we summarize currentContract's calls externally this does not affect CVL calls via a parametric f

composer/kb/resources/cvl_summarization_rag_draft.md
```diff
@@ -0,0 +1,925 @@
...
+Selector-unresolved raw calls take `unresolved external ... DISPATCH [...]
+default <summary>` (A4.4), with the `default` choice made deliberately —
+`ASSERT_FALSE` where "everything is dispatched" should be a checked obligation
+rather than an assumption. Receiver-unresolved calls with a known selector
+take the hybrid CVL dispatch (B4), or `DISPATCHER` when its scene-wide
+signature scoping is genuinely intended.
+
+### B3.4 Rungs 3–4 — model: spec-backed, then axiomatized
+
+If resolution is impossible or defeats the purpose (the code is the timeout),
```
shellygr: what does the "the code is the timeout" even mean

composer/kb/resources/cvl_summarization_rag_draft.md
```diff
@@ -0,0 +1,925 @@
...
+
+## B5. Diagnosing summaries that did not apply
+
+When a rule failure shows a havoc at a call site an entry was expected to
+cover, the common causes: wrong visibility face (public duality, A2.2);
+default `UNRESOLVED` policy on a wildcard while the call is resolved (A2.3); a
+catch-all entry whose receiver the Prover could not pin to the named contract;
+entry pattern mismatch (locations on reference-type parameters, exact
+signature); an entry naming the inheriting contract instead of the defining
+contract for an inherited method (A2.1); the call originating from CVL (A2.5).
```
shellygr, line 906: I think we cannot just assume sighash is conveniently always resolved and not to put specific instructions in such cases

composer/templates/property_judge_prompt.j2
```diff
@@ -142,9 +196,44 @@ As an exception to the above, a requirement like "msg.sender is never contract C
...
+   function as `NONDET` makes transfers no-ops, trivializing every property over token balances. (Whether
+   the weakening was *licensed* is Criteria 2; here the question is what it does to each property.)
+4. Pinned values: `ALWAYS(v)` or `CONSTANT` on a return value the property then checks — the
+   assertion tests the pinned value, not the implementation.
+5. Adversary removal: a property whose point is robustness against arbitrary callee behavior (attack
+   vectors above all) verified under a summary or closed-world construct that models only benign
+   behavior — the modeled adversary can no longer attempt the attack the property exists to rule out.
+
+Two calibrations. First, modeling contracts *outside* the provided set — the ghost-mapping pattern,
+hybrid dispatch, &c — is what summaries are for; the defect in shapes 1-2 is summarizing the
```
shellygr: &c?

composer/kb/resources/recipes/r11-ecrecover-axioms.md
jtoman: I am confused what happened here.

composer/kb/resources/recipes/r12-ghost-mirror-quantified-invariants.md
jtoman: > "load-bearing"

Claude? is that you?

jtoman: but yeah this the text being cute. I edited it heavily and we still got this, sigh

composer/kb/resources/recipes/r15-preserved-blocks.md
jtoman: yeah I mean, I would love to get an inventory of the set of "reasonable assumptions" the author is allowed to make "for free". I see, for example, `require totalSupply() < 2^128` pretty frequently in our hand authored rules, so there's at least 2...

composer/kb/resources/cvl_invariants_quantifiers.md
jtoman: yeah, this is based off some text I wrote, and I'm so used to seeing `AssumeCmd` in tac I forgot we use `require` for the surface syntax...

composer/kb/resources/cvl_summarization_rag_draft.md
jtoman: the cvl spec isn't the problem, the code is. this is awful wording lol

composer/kb/resources/cvl_summarization_rag_draft.md
jtoman: ? I'm not sure what in the text you're referring to here.

composer/templates/property_judge_prompt.j2
jtoman: etc. I'll just write etc, _I_ was being cute this time.

composer/kb/resources/cvl_invariants_quantifiers.md
jtoman: yeah a little. if we think we should follow DRY, I don't mind trimming the fat here.

composer/kb/resources/recipes/r10-nativecodesize.md
jtoman: I'm going to double check this, because I don't know if we assume codesize == 0, or codesize != 0 in preserved constructor...

jtoman: image

HMMM

Worth noting here, and worth noting this is (technically) wrong in our semantics I guess.

composer/kb/resources/recipes/r18-reentrancy-recipes.md
jtoman: It's pulled word for word from here:

https://github.com/Certora/Examples/blob/master/CVLByExample/Reentrancy/certora/spec/Reentrancy.spec

A comment on the original suggests this is simulating "calling" the fallback which just reverts (which of course this isn't sound with a non-trivial fallback, will annotate).

composer/kb/resources/recipes/r18-reentrancy-recipes.md
jtoman: you're not gonna believe where we pulled these examples lol: https://github.com/Certora/Examples/blob/master/CVLByExample/Reentrancy/certora/spec/NoGuardSafety.spec

Whatever design went into that approach applies here (I don't know I'm afraid)

composer/kb/resources/cvl_baseline_facts.md
jtoman: good catch

composer/kb/resources/cvl_summarization_rag_draft.md
jtoman: clarified this applies to "internal" methods; where if you aggressively summarize the internal behavior of a contract you end up not proving much.

composer/kb/resources/cvl_summarization_rag_draft.md
jtoman: my co-author asked about this, and I said "haha no, the agent will abuse it". I stand by that gut feeling, but we can expose it later if we like?

composer/kb/resources/recipes/r12-ghost-mirror-quantified-invariants.md
jtoman: yeah this sucks.

jtoman: it's also better explained in the dedicated invariants/quantifier document, to your comment about repeating...

shellygr (approved):

PR #116, opened 2026-08-01
Add Truffle build-system support

shellygr:
Stacked on #114 — that PR anchors build-system detection on the contract instead of the run root; this one teaches detection a third build system to anchor on.

## Why

Truffle predates Foundry and Hardhat and is still what many 0.4.x/0.5.x projects ship, but detection knew only `foundry.toml` and `hardhat.config.*`. Those repos resolved to "unknown build system", which costs two things: the project's pinned solc (and optimizer settings) is lost, and — the expensive one — no `packages` list is emitted, so every bare-scope `@scope/pkg/...` import fails to resolve.

## What

- **`TruffleManager` / `TruffleConfig`** — evaluates `truffle-config.js` (or the v4-era `truffle.js`) with node, the same way `HardhatManager` already reads its config, and takes `compilers.solc.version` plus the v4 top-level `solc` settings block. Unlike the Hardhat extractor it never requires the build system itself: a Truffle config is a plain CommonJS module, so a project with no `truffle` in `node_modules` still yields its pin.
- **Non-pin versions are not guessed.** Truffle also accepts `"pragma"`, `"native"` and path/docker specs in `compilers.solc.version`; none names a version a conf can carry, so they behave like an absent version and the compilation phase resolves per contract from source pragmas.
- **`packages` come from the shared remapping-source builder**, so a Truffle project gets the same `node_modules`-derived resolver list a Foundry project gets. This is emitted even when the config itself cannot be evaluated (typically a `require` of a dependency that was never installed) — that case logs the reason and falls back to Truffle's defaults instead of aborting the run.
- **`TruffleContractExtractor`** — reads Truffle's flat `build/contracts/*.json`, dropping bytecode-less artifacts (interfaces/abstract) and anything whose recorded `sourcePath` lies outside the project root.
- **Detection order**: `foundry.toml` > `hardhat.config.*` > truffle config > `package.json` dependency > artifact layout. Truffle stays last throughout, so a repo that migrated off it keeps using the config that now drives it.
- **`truffle-config.js` / `truffle.js` also anchor `find_build_config_dir`**, so a monorepo keeping one config per app is detected per contract (this is what composes with #114).
- **Packaging fix**: `**/*.js` was missing from `[tool.setuptools.package-data]`. The node extractors run out of the installed package, so neither this one nor the existing hardhat one was shipped in the wheel.

## Validation

Unit tests (21 new) cover config parsing for both Truffle generations, version-pin rejection, detection precedence, nested-config anchoring, artifact extraction, and two node-backed cases (a config that computes its own settings; a config whose `require` fails).

Beyond the unit tests, run against real Truffle projects from the sweep corpus:

- a project whose config sits one directory below the run root: anchors to that directory, detects truffle, recovers its `solc5.17` pin — previously lost entirely
- a project whose config requires an uninstalled dev dependency: logs `Could not evaluate truffle-config.js (Cannot find module '...'); using Truffle defaults` and still emits 29 packages
- a project carrying both a truffle and a hardhat config: resolves to hardhat, unchanged
- a Foundry project: unchanged, still `solc8.23` with its full remappings

`354 passed, 9 skipped` on the CI pytest command; pyright clean.

## Not covered

The build step itself — nothing here runs `truffle compile`. This fixes what the conf carries, which is what the prover's own compilation consumes. Contract auto-discovery for a Truffle project consequently still needs a `build/contracts` directory to exist (before this change that path hard-failed with "no build system detected", so it is not a regression).

### DIFF
--- a/certora_autosetup/build_systems/truffle.py
+++ b/certora_autosetup/build_systems/truffle.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+"""
+Truffle Manager - Manages Truffle project configuration and artifacts.
+
+Parallel to FoundryManager/HardhatManager. Truffle predates both and is still what many
+0.4.x/0.5.x projects ship (`truffle-config.js`, or `truffle.js` on Truffle v4), so without
+it those repos detect as "unknown build system": their pinned solc is lost and their
+`@scope/pkg/...` imports resolve against nothing.
+"""
+
+import json
+import subprocess
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+from packaging.version import InvalidVersion, Version
+
+from certora_autosetup.build_systems.base import BuildSystemConfig
+from certora_autosetup.build_systems.manager import BuildSystemManager
+from certora_autosetup.utils.remappings import build_packages_from_remapping_sources
+
+
+@dataclass
+class TruffleConfig(BuildSystemConfig):
+    """Parsed Truffle configuration with resolved settings."""
+
+    # Truffle-specific fields (common fields inherited from BuildSystemConfig)
+    contracts_build_directory: Optional[str] = None
+
+    # node_modules-derived remappings; see to_certora_dict()
+    packages: Optional[List[str]] = None
+
+    def __post_init__(self):
+        """Initialize default values for mutable fields."""
+        super().__post_init__()
+
+        if self.src is None:
+            self.src = "contracts"
+        if self.contracts_build_directory is None:
+            self.contracts_build_directory = "build/contracts"
+        if self.packages is None:
+            self.packages = []
+
+    def to_certora_dict(
+        self,
+        convert_solc_to_certora_format: bool = True,
+        include_packages: bool = True
+    ) -> Dict[str, Any]:
+        """
+        Convert Truffle config to Certora format.
+
+        Args:
+            convert_solc_to_certora_format: Whether to convert "0.8.19" to "solc8.19" format
+            include_packages: Whether to include packages/remappings
+
+        Returns:
+            Dictionary with Certora config format
+        """
+        result = self._apply_common_solc_settings(convert_solc_to_certora_format)
+
+        # Truffle resolves bare imports (`@scope/pkg/contracts/Foo.sol`) through node_modules
+        # with no remapping file anywhere in the project, so the packages list built from
+        # package.json is the only thing that lets solc find those sources.
+        if include_packages and self.packages:
+            result["packages"] = self._relativize_packages(self.packages)
+
+        return result
+
+    def get_artifact_directory(self) -> str:
+        """Return Truffle artifact directory."""
+        return self.contracts_build_directory or "build/contracts"
+
+
+class TruffleManager(BuildSystemManager):
+    """Truffle project manager: config parsing and artifact discovery."""
+
+    def __init__(self, project_root: Path, scope):
+        """
+        Initialize Truffle manager.
+
+        Args:
+            project_root: Root directory of the project
+            scope: Centralized scope for consistent filtering
+        """
+        super().__init__(project_root, scope, "TruffleManager")
+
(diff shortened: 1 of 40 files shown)


certora_autosetup/build_systems/truffle.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python3
+"""
+Truffle Manager - Manages Truffle project configuration and artifacts.
+
+Parallel to FoundryManager/HardhatManager. Truffle predates both and is still what many
+0.4.x/0.5.x projects ship (`truffle-config.js`, or `truffle.js` on Truffle v4), so without
+it those repos detect as "unknown build system": their pinned solc is lost and their
```
jar-ben: the part
>  so without
it those repos detect as "unknown build system": their pinned solc is lost and their
`@scope/pkg/...` imports resolve against nothing

really does not need to be here.

jar-ben (approved):

certora_autosetup/build_systems/truffle.py
shellygr: Claude answers: Removed. The module docstring now says what the manager recognises and reads (both config names, `truffle-config.js` and Truffle v4's `truffle.js`; compiler settings and `@scope/pkg/...` package roots from whichever is present) with no account of what happened before it existed.

Applied the same to `tests/test_truffle.py`, whose docstring had the same "Regression origin: …" narrative. Same cleanup went into #114 where you raised this first.

---

Unrelated to the comment, but found while re-reading this: **this PR's `project_dir.py` change becomes a no-op once #114 lands.** It adds the two Truffle names to `BUILD_CONFIG_FILENAMES`, but #114 has since reworked `find_build_config_dir` to go through a new `_artifact_dir_of()` helper — anchoring now requires the candidate to hold a build config *and* have its artifact directory on disk. `BUILD_CONFIG_FILENAMES` is no longer read by that path, and `_artifact_dir_of` only knows foundry (`out`, from foundry.toml) and hardhat (`artifacts`).

So after rebasing this onto #114, a Truffle monorepo would silently stop anchoring on `apps/<app>/truffle-config.js`. The fix is a third branch in `_artifact_dir_of` returning Truffle's artifact dir — `build/contracts` by default, or `contracts_build_directory` when the config sets it, which `TruffleConfig.get_artifact_directory()` already resolves.

I've left it alone rather than guess at the rebase order. Happy to add it here once #114 merges, or fold it into #114 instead — whichever you prefer.

shellygr:
Merged `master` in (`5f7d2de`) now that #114 has landed, and closed the gap I flagged above.

**The gap was real.** #114 reworked `find_build_config_dir` to anchor only on a directory that holds a build config *and* has its artifact directory on disk, routed through the new `_artifact_dir_of`. `BUILD_CONFIG_FILENAMES` is no longer consulted by that path — so this branch's addition of the two Truffle names to it had quietly become inert. A Truffle monorepo would have resolved back to the run root with nothing failing loudly.

Resolution: took master's `project_dir.py` and added the third branch to `_artifact_dir_of` — `build/contracts`, Truffle's default. The `contracts_build_directory` override is deliberately not read there: that needs node against the config, and this only has to *recognise* a built project. `TruffleConfig.get_artifact_directory()` still resolves the real path when the manager needs it.

`tests/test_project_dir.py` conflicted add/add purely as an artifact of the squash merge — this branch never touched it, so master's copy stands.

**Tests.** Three new cases in `test_project_dir.py`: `truffle-config.js` anchors, the v4 `truffle.js` name anchors, and an unbuilt Truffle config correctly yields to a built root. `test_truffle.py`'s `test_nested_truffle_config_anchors_detection` needed its fixture updated to create `build/contracts` — under the new rule a config without artifacts is intentionally not enough to anchor on, which is exactly the property that keeps maple-core-v2 and autonolas-v1 from regressing.

90 pass across the touched modules; full suite 308 pass with the same 10 pre-existing env failures (missing `tabulate` / `certora_cli`) that master has. All four CI checks green.

Merged rather than rebased to keep the branch push-only — no force-push.

PR #117, opened 2026-08-01
Fix Field() description/default confusion in LLM-facing schemas

shellygr:
## The bug

`pydantic.Field`'s only positional parameter is `default`. So

```python
path: str = Field("The relative path to the file which defines the contract type this represents")
```

does not describe the field — it makes `path` **optional** and sets its default value to that sentence. Two things go wrong at once:

1. The model is never told what `path` means (no `description` reaches the JSON schema).
2. `path` is no longer required, so the model may legitimately omit it — and when it does, we silently get the sentence back as data.

Six fields were declared this way:

| field | what the bogus default becomes |
| --- | --- |
| `SourceExplicitContract.path` | a contract source path |
| `Invariant.name` | a CVL invariant name |
| `MethodWithoutReturn.host_contract` | a contract type |
| `ImportSpec.spec_file` | a spec path |
| `ContractImport.contract_name` / `.as_name` | a contract name / CVL binding |

## Why it only shows up sometimes

Nothing is broken at construction time — the model validates fine, because the sentence *is* a valid `str`. The defect only materialises when the LLM happens to omit the field on that particular call, which is a per-call sampling accident: the same repo at the same commit will produce a correct `path` on most contracts and a defaulted one occasionally. That is why this sat latent since #71 (April) and why, in the run that surfaced it, six of seven per-contract jobs on the *same commit* succeeded and one died.

There is also no type error to catch it downstream. For `SourceExplicitContract.path` the value flows into

```python
extra_files = [c.path for c in sys_desc.transitive_closure if c.solidity_identifier != source.contract_name]
```

and on into the autosetup argv, so the failure surfaces far from its origin, as

```
ValueError: Error: The relative path to the file which defines the contract type this represents is not a .sol file
```

and the run exits 1 ~25 minutes later when `prepare_formalization`'s `gather` re-raises. The other four fields fail more quietly still — a sentence as an invariant name or a contract binding just produces bad CVL.

## The fix

Spell all six `description=`. The fields become required again and the prose reaches the model as documentation, which is what it was written for.

`tests/test_pydantic_field_descriptions.py` adds an AST lint over `composer/`: a string literal in `Field`'s positional slot is always this confusion (a real string default is `Field(default=...)`), while `Field(...)` stays legal.

## Testing

- new lint passes; verified it flags `Field("prose")` and accepts `Field(...)` / `Field(description=...)`
- full suite: `327 passed, 9 skipped`, 5 failures — all 5 reproduce identically on unmodified `master` (`invariant-cvl` tape lane exhausted in the autoprove/codegen end-to-end tests; two `test_indexed_tool` cache-count assertions)
- no in-tree constructor relied on any of the six defaults, so nothing internal starts failing validation

### DIFF
--- a/composer/cvl/schema.py
+++ b/composer/cvl/schema.py
@@ -2,11 +2,11 @@ from typing import Annotated, Literal, Protocol
 from pydantic import BaseModel, Field, Discriminator
 
 class ImportSpec(BaseModel):
-    spec_file: str = Field("The (relative) path to the spec file to import")
+    spec_file: str = Field(description="The (relative) path to the spec file to import")
 
 class ContractImport(BaseModel):
-    contract_name: str = Field("The name of the contract to import (e.g. ERCTokenA)")
-    as_name: str = Field("The CVL identifier to bind to, e.g. 'tokenA'")
+    contract_name: str = Field(description="The name of the contract to import (e.g. ERCTokenA)")
+    as_name: str = Field(description="The CVL identifier to bind to, e.g. 'tokenA'")
 
 class MappingType(BaseModel):
     """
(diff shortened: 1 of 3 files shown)


jtoman (approved):
*SIGH*

PR #118, opened 2026-08-02
Don't crash when there's no design doc

naftali-g:


composer/pipeline/cli.py
```diff
@@ -291,7 +301,7 @@ async def cont[FormT: BackendResult, A: ArtifactIdentifier, U: FeatureUnit, Main
                 embed_model=model,
                 root_key=root_key,
                 source=SourceCode(
-                    content=system_doc_doc,
```
jtoman, line 294: "doc_doc" good stuff

jtoman (approved):
seems to do what it says on the tin

composer/spec/source/design_doc_finder.py
```diff
@@ -282,28 +292,24 @@ async def resolve_design_doc[P: HasName](
...
-    supplied doc produce an identical key.
+) -> pathlib.Path | None:
+    """Resolve the design document to a path, or ``None`` for a source-only run.
+
+    Run the finder under a ``run_task`` discovery phase. If it finds a document,
+    return its path (which feeds the byte-hash root cache key, so a discovered doc
+    and a supplied doc produce an identical key). If it finds nothing, return
+    ``None`` instead of failing: the pipeline degrades to source-only, deriving the
+    component breakdown from the source code alone. ``_discover`` has already
+    surfaced the source-only downgrade to the user via ``_emit_choice``.
```
jar-ben: I don't like when function's docstring describes what the callers of this function do, i.e. this part is extra:
> the pipeline degrades to source-only, deriving the
    component breakdown from the source code alone. ``_discover`` has already
    surfaced the source-only downgrade to the user via ``_emit_choice``

But perhaps that's the style used in this part of the codebase, i.e. tioli.

PR #122, opened 2026-08-04
Unblock the containerized local run: urllib3, RAG rebuild, results download

shellygr:
Four independent failures found while running the pipeline locally in docker, each
stopping it at a different point. None of them are related to each other; they are
together because they are all in the way of the same thing.

## 1. `import urllib3` is broken in the image

`certora-cloud` → `niquests` → `urllib3-future`, and `urllib3-future` ships a
`urllib3_future.pth`. That hook runs on **every interpreter start** and, whenever it finds
both distributions installed and the directory writable, deletes `site-packages/urllib3`
and copies `urllib3_future` over it. What lands is a hybrid carrying urllib3-future's
`backend/` but no `urllib3/exceptions.py`, so every plain `import urllib3` fails:

```
ImportError: cannot import name 'exceptions' from partially initialized module 'urllib3'
             (most likely due to a circular import)
```

The message points at a circular import, which is misleading — the module is simply
incomplete. `setup-db` hits it through `spacy → requests → urllib3` and never runs.

Fixed in `scripts/Dockerfile`: drop the hook, remove the hybrid tree, reinstall urllib3.
Two details matter. Removing the tree is **required** — a reinstall only replaces files
listed in urllib3's own `RECORD`, so `backend/` survives and the hook reads it as "already
substituted" and clobbers again on the next start. And the version comes from
`importlib.metadata`, i.e. whatever the lockfile already resolved, rather than a second
place to keep a pin in sync.

The step is guarded on the `.pth` existing, so it no-ops if this is fixed upstream, and it
asserts `import urllib3, requests, spacy, niquests` in the same layer — a dependency bump
that reintroduces the problem now fails the build instead of surfacing as a runtime
`ImportError` deep inside a pipeline phase. `niquests` imports `urllib3_future` directly
and is unaffected.

## 2. `setup-db` cannot be re-run

`ragbuild` appends, and `manual_sections` is `UNIQUE (h1..h6, part)`, so a second
`setup-db` over a populated schema dies:

```
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "parts_unique"
```

No new mechanism added — `scripts/wipe_rag.py` and the wipe-then-rebuild ordering in
`refresh_rag.sh` already exist. The entrypoint now empties the tables before populating,
with the same `--skip-confirmation` rationale `refresh_rag.sh` documents (the wipe is
immediately followed by a rebuild, not a standalone destructive op). The RAG content
derives entirely from the CVL HTML baked into the image, so there is nothing to preserve.

`wipe_rag.py` lives in `scripts/` rather than as `composer` package data, so the image
needs it copied in — that `COPY` is why `setup-db` could not already reuse it.

## 3. `wipe_rag.py` fails for every caller

Reusing it surfaced that it was already broken. Its hardcoded list names a `code_refs`
table, and no DDL in the repo creates one — the RAG schema is `documents`,
`manual_sections`, `manual_section_code_refs`. So it raises
`UndefinedTable: relation "code_refs" does not exist`, which also means **`refresh_rag.sh`
does not work today**.

Corrected the list, and resolved it against `pg_catalog` before naming anything: `TRUNCATE`
has no `IF EXISTS`, so one absent name fails the whole statement, and this list has drifted
from the schema once already. The hardcoded allowlist still bounds what can ever be
truncated — the catalog query only narrows it. Table names are now composed with
`sql.Identifier` rather than interpolated.

## 4. The prover results download is bounded on the wrong thing

`composer/prover/cloud.py` bounded the whole request at `ClientTimeout(total=300)`. Results
archives reach hundreds of megabytes, and elapsed time cannot distinguish a large transfer
from a stalled one, so a healthy download is aborted mid-stream:

```
File "composer/prover/cloud.py", line 216, in cloud_results
  async for chunk in resp.content.iter_chunked(8192):
...
TimeoutError
```

That killed CVL_GEN after 19 minutes on a real run. Bound inactivity instead —
`total=None, connect=60, sock_read=180`. `sock_read` is rearmed on every chunk, so a slow
transfer runs to completion while a dead connection still fails in bounded time.
`tests/test_cloud_results_timeout.py` pins that shape, since "add a `total` back" is an
easy and silent regression.

## Validation

- **urllib3** — exercised by the real image build: the step detected the `.pth`, read the
  pinned version, repaired the tree, and its `import urllib3, requests, spacy, niquests`
  assertion passed.
- **RAG rebuild** — ran `setup-db` against a schema already holding 231 documents / 198
  sections / 442 code refs, i.e. the exact state that used to raise `parts_unique`. It
  emptied the tables, rebuilt, reported `setup-db done.`, and landed on **identical**
  counts (231 / 198 / 442). Same input, same resulting state, no duplicates.
- **wipe_rag.py** — `TRUNCATE TABLE manual_section_code_refs, manual_sections, documents
  RESTART IDENTITY CASCADE` now succeeds where it previously raised `UndefinedTable`.
- **Download timeout** — unit tested, and asserted on the built image. Not yet exercised
  end-to-end; that needs a run reaching CVL_GEN.
- `pyright composer/ analyzer sanity_analyzer certora_autosetup` clean (0 errors; the one
  `FormT` warning in `composer/pipeline/core.py` is pre-existing and untouched).
  `scripts/` is outside CI's pyright scope; the one error reported there for
  `wipe_rag.py` is a pre-existing psycopg symbol-resolution gap that also reports on
  `master`.

## Note on the local build

All three `scripts/Dockerfile` stages are pinned `--platform=linux/amd64`, so on Apple
Silicon the whole build runs under QEMU. The embedding-model download step alone exceeds
30 minutes there. That is pre-existing and not addressed here.


shellygr:
Replaced the urllib3 approach after digging into why urllib3-future does this.

**The earlier fix treated the symptom.** I had been deleting `urllib3_future.pth` and
reinstalling urllib3 after the fact. Reading `hatch_build.py` in urllib3-future's sdist shows
the hook is only half of the mechanism:

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/urllib3", "src/urllib3_future"]
```

The default wheel **contains its own top-level `urllib3` package**, so it overwrites the real
urllib3's files at install time; the `.pth` merely re-applies that on every interpreter start.
Deleting the hook left the wheel's duplicate files in place and left the outcome dependent on
install ordering.

**It is also worse than a hard failure.** In a clean venv the substitution completes and
nothing raises — `import urllib3` reports **2.24.900**, i.e. the fork, with urllib3 2.6.3
installed. Our image happened to catch it mid-substitution, which is the only reason it
showed up as an error rather than as silently running on different HTTP code.

**The fix upstream sanctions.** The same build hook reads `URLLIB3_NO_OVERRIDE`, and when set
it `rmtree`s `./src/urllib3` and skips the `.pth`. Their README's packaging notes say so
directly: *"When packaging for a distribution registry, set `URLLIB3_NO_OVERRIDE=1` during the
build ... This prevents the in-place upgrade behavior"*. Reaching the hook requires building
from the sdist, hence `[tool.uv] no-binary-package = ["urllib3-future"]` plus the variable on
the installing stage. urllib3-future is pure-Python/hatchling, so the sdist build is cheap and
its binary deps (`jh2`, `qh3`) still install as wheels.

Measured in clean venvs:

| | default wheel | no-binary + URLLIB3_NO_OVERRIDE |
|---|---|---|
| `urllib3_future.pth` | present | absent |
| `urllib3/` file collision | yes | none |
| `import urllib3` reports | **2.24.900** (fork) | **2.6.3** (real) |
| installed dirs owned | overlapping | disjoint: `urllib3` / `urllib3_future` |
| `niquests` / `requests` | ok | ok |

`scripts/check_urllib3_identity.py` gates the build, verified in both directions — it passes on
a correct env and fails on the default one with:

```
- urllib3_future.pth is installed; it substitutes urllib3 on interpreter start
- urllib3/backend/ exists; the urllib3 package directory is urllib3-future's
- the imported urllib3 reports version 2.24.900 but the installed urllib3 distribution is 2.6.3
```

No file deletion or post-install repair remains.

shellygr:
Claude answers: the urllib3 change is split out into #126 and removed from here. This PR is now just the three RAG-rebuild / results-download fixes:

- `setup-db` empties the RAG tables before rebuilding, so it can be re-run
- `wipe_rag.py` no longer names a `code_refs` table that no schema creates (it raised `UndefinedTable` for every caller, `refresh_rag.sh` included)
- the prover results download bounds inactivity rather than elapsed time

`462 passed, 9 skipped, 9 deselected`, pyright clean.

PR #123, opened 2026-08-04
Don't download whole zipOutput

naftali-g:


composer/prover/cloud.py
```diff
@@ -197,28 +194,15 @@ async def on_status(status: str) -> None:
...
-        finally:
-            tmp_path.unlink(missing_ok=True)
-
-        yield (find_results_root(dest), runtime_ms)
+        # Fetch just the tree view and the compiled sources, which is everything
+        # the results parse and the CEX analyzer read. The job's zipOutput archive
+        # also carries ``outputs/`` and ``debugs/`` — the split dumps — which reach
+        # tens of gigabytes on real jobs and used to exhaust the disk; across the
+        # jobs measured here these two subtrees are ~3% of the archive. POU writes
+        # them in the same layout the archive had, so the parse is unchanged.
```
jar-ben, line 204: "Fetch just the tree view and the compiled sources, **which is everything the results parse and the CEX analyzer read**" - I will trust you with this one. 

But I think this whole comment can be removed; it's not describing what the code does but rather which issue it fixed. `_results_api().fetch_sources_and_treeview_files` says everything the reader wants to know.

jar-ben (approved):

PR #124, opened 2026-08-05
Provide system doc also to property inference (when available)

naftali-g:


composer/pipeline/core.py
```diff
@@ -406,6 +409,22 @@ async def _post_plugin_props(
...
+        # Mirrors the natspec path's `system_doc` injection (composer/spec/natspec/pipeline.py):
+        # the design doc, when supplied on the source/prover path too, should inform property
+        # inference the same way it informs component analysis, not just the latter.
+        design_doc = run.source.content
+        if design_doc is not None:
+            pre_input = [
+                *pre_input,
+                CacheablePropertyGenerationInput(
+                    "certora:system-doc", "generic", "always",
+                    lambda cache, doc=design_doc: [
```
jtoman, line 421: claude loves to think you need this weird capture trick. I'm fairly certain you don't. It's harmless I guess, but whatever.

jtoman (approved):

PR #130, opened 2026-08-07
Tell the harness agent not to extend a non-deployable contract

shellygr:
## Problem

The harness generation prompt says the harness should be "a minimal contract which simply extends the target contract", adding "only code necessary to satisfy the Solidity compiler/typechecker". That holds only while the target is deployable.

When the target is `abstract`, extending it directly produces a contract solc rejects — and even if it compiled, the Prover would have no instance for it. Judging deployability from the target's own file is not enough either: a function can be declared in an interface the target inherits and implemented by a sibling contract mixed in further down the hierarchy, so the target's file shows no sign it is missing.

That is exactly how a dev run failed. The agent harnessed an abstract base, implemented the unimplemented functions it could see in that base's file, and missed five more declared only in an inherited interface. The run died in AutoSetup.

## Change

Prompt-only. Directs the agent to treat deployability as a precondition: when the target is abstract (or otherwise leaves an inherited function unimplemented), harness the concrete contract the protocol actually deploys, choosing by the deployment scripts/configuration where several exist. Implementing the missing functions in the harness itself stays available, but as the last option, for a target with no concrete descendant at all.

## Validation

Template renders. This is the weakest of the three fixes on its own — it asks the model to get something right rather than making it impossible to get wrong — so it is worth landing behind the compile gate, which turns a wrong choice into a repair loop instead of a dead run.

## Follow-up worth considering

Whether a target is abstract is a fact we can compute rather than ask for: a parse-only `solc --standard-json` (`stopAfter: "parsing"`) yields `ContractDefinition.abstract` in milliseconds with no dependency resolution. Passing that per target into the prompt — or, better, resolving abstract components to their concrete descendants upstream in the component analysis, where `num_instances` is assigned to a contract that can never be a Prover instance — would remove the judgment call entirely. Left out here to keep this a prompt change.

## Companions

Two independent PRs address the same failure: the harness compile gate, and a terminal AutoSetup detection for the resulting compiler error.

Companion PRs: #129 #131. They are independent — any order, any subset.


shellygr:
as far as prompting can go...

jtoman (approved):

PR #131, opened 2026-08-07
Treat an unimplemented contract as a terminal compilation failure

shellygr:
## Problem

When a contract in the compilation input inherits functions it never implements, solc rejects it outright (`Contract "X" should be marked as abstract.`). Nothing in the conf can change that — it is a defect in the contract's source.

The retry loop cannot tell, so it treats the failure like any other: the specific detectors miss, the `use_relpaths_for_solc_json` catch-all fires, and the run spends further `certoraRun` passes plus the import-patch pass before reporting a generic "compilation analysis failed" that names neither the contract nor the reason. In a dev run this cost four compilations and the import patch, all reaching the identical error, and the log's only actionable content was the raw solc output buried in a warning dump.

The contracts most exposed to this are the generated harnesses: they are written against their target's own file, so a function the target inherits from an interface and leaves to a sibling mixin is invisible at the point the harness is written.

## Change

Detect the diagnostic in the retry loop, immediately after the existing `_detect_abstract_main_contract` check, and raise `UnimplementedContractError` naming the contract and the file that declares it. Like that check it runs before any workaround is applied, so the loop stops at the compilation that reported the error rather than after exhausting the catch-all.

Detection normalizes whitespace first (`_normalize_ws`), as the other multi-line detectors in this module do — solc hard-wraps its diagnostics, and the wrap can fall inside this sentence. The contract name is quoted by solc so it survives the fold; the source location that follows is optional, since only the diagnostic itself is guaranteed to be present.

Consistent with `AbstractMainContractError`, the raise unwinds to `run_compilation_analysis`'s handler, which logs the message and fails the phase.

## Validation

- `uv run --no-sync python -m pytest tests/ -m "not expensive" -q` → 507 passed; `pyright` → 0 errors
- new tests: detection with and without a source location, a negative case, and a loop-level test asserting exactly one `certoraRun` before the raise
- ran the detector over the verbatim output of the real run that motivated this: it returns the contract and its file

## Note on approach

The detector is a regex over compiler text, which is the established mechanism in this module (every existing detector works that way) — the structured `--json` diagnostics are not available on this path, since what is being parsed is `certoraRun`'s output rather than solc's.

## Possible extension

The message currently reaches the log but not the run's user-facing failure, which stays the generic "Compilation analysis failed" — same as for `AbstractMainContractError` today. Propagating either one as the reported reason is a separate change.

## Companions

Two independent PRs address the same failure earlier in the pipeline: a compile gate on generated harnesses, and prompt guidance against extending a non-deployable contract.

Companion PRs: #129 #130. They are independent — any order, any subset.


certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -264,6 +278,21 @@ def _detect_abstract_main_contract(self, output: str, compilation_config: Dict)
...
+    def _detect_unimplemented_contract(self, output: str) -> Optional[Tuple[str, Optional[str]]]:
+        """Return the (contract name, declaring file) a contract in the input was
+        rejected for: it inherits functions it does not implement, and is not declared
+        ``abstract``.
+
+        The file is None when the diagnostic carries no source location. This is a
+        source-level defect — most often in a generated harness, which is written
+        against the target's own file and misses the functions the target inherits
+        from elsewhere — so the compilation settings have no bearing on it.
+        """
```
shellygr, line 288:
```suggestion
        The file is None when the diagnostic carries no source location.
```

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -264,6 +278,21 @@ def _detect_abstract_main_contract(self, output: str, compilation_config: Dict)
         no_bytecode = {m.group(1) for m in re.finditer(r"Contract (\S+) has no bytecode", output)}
         return main_contract if main_contract in no_bytecode else None
 
+    def _detect_unimplemented_contract(self, output: str) -> Optional[Tuple[str, Optional[str]]]:
+        """Return the (contract name, declaring file) a contract in the input was
+        rejected for: it inherits functions it does not implement, and is not declared
+        ``abstract``.
+
+        The file is None when the diagnostic carries no source location. This is a
+        source-level defect — most often in a generated harness, which is written
```
jar-ben: I would drop this part: 
_"most often in a generated harness, which is written against the target's own file and misses the functions the target inherits from elsewhere"_

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -545,6 +574,22 @@ def run_compilation_with_workarounds(
                     f"verified. Re-run with a concrete implementation as the main contract."
                 )
 
+            # Terminal for the same reason, one step earlier: a contract in the input
```
jar-ben: "Terminal for the same reason" - the "for the same reason" depends on the current codebase ordering. Just drop that part of the comment or rephrase.

jar-ben (approved):
I agree it's an improvement and I am ok with merging it (just stated some comment improvements). 

However, I have to say the obvious - this just allows us to crash faster, but it's not fixing the issue. I thought we have a last resort compilation workaround - dropping non-main contract from the scene if no compilation workaround helps (perhaps it was not merged?). Does it make sense to do it here? I guess it depends on if it can be fixed by the user or if it's auto-generated by us. 

And ofc, if it's a harness we auto-generate, I hope we are fixing the generation.

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: dropped — the docstring now says only that it is a source-level defect.

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: rephrased to "Also terminal:" — no longer leans on the ordering relative to the abstract-main-contract check.

shellygr:
Claude answers: on the review body.

- **"crash faster, not fixing the issue"** — agreed, that is exactly the scope. The value is that the run stops at the compilation that reported the error and names the contract and its file, instead of burning 4 `certoraRun` passes plus the import patch and then reporting a generic "compilation analysis failed".

- **last-resort workaround dropping non-main contracts from the scene** — not merged, and not present on master. The only `last_resort=True` workaround in `compilation_workarounds.py` is `use_relpaths_for_solc_json`; there is no drop-contract-from-scene workaround anywhere in the repo. So there is nothing here to reuse.

  Whether to add one is a real question but a separate one: dropping a contract silently shrinks the scene, which changes what the run verifies. If the offending contract is a generated harness, dropping it also drops the very thing the run was built around, so failing loudly is arguably the more honest outcome. If it is user source, dropping it is a scene change we would want to report, not do quietly. Happy to open a follow-up if you want it explored.

- **"if it's a harness we auto-generate, I hope we are fixing the generation"** — yes, both companions are merged: #129 compile-checks generated harnesses before accepting them, and #130 tells the harness agent not to extend a non-deployable contract. This PR is the backstop for the cases those two miss (and for the non-harness case, where the defect is in user source).

Both inline comments are addressed in f742e60.

PR #133, opened 2026-08-08
Answer every tool call of the turn that invoked verify_spec

shellygr:
## Problem

Counter-example analysis (`analyze_cex_raw`) is a one-shot LLM call made from *inside* the `verify_spec` tool, and it reuses the agent's live message history. That history is snapshotted while the tools node is still executing, so its trailing `AIMessage` is the turn that opened the currently-running tool calls — none of which has a result yet.

The function appended a `ToolMessage` for `verify_spec`'s own `tool_call_id` only. When the model batches another tool call into the same turn (e.g. `feedback_tool` alongside `verify_spec`, which it does), that sibling `tool_use` block reaches the API with no `tool_result` after it and the request is rejected:

```
400 messages.N: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_…
```

The failure is deterministic for that turn shape as soon as one rule comes back `VIOLATED`, and it lands *after* the prover has run — so a CVL-generation phase can burn the better part of an hour, plus cloud prover time, and then die with nothing to show. It has hit real runs.

## Fix

Fill in a stand-in `ToolMessage` for every tool call of the trailing assistant turn that nothing has answered yet (`verify_spec`'s own result still carries the counter-example). Already-answered ids are skipped, so a sibling that finished first keeps its real result and never gets a duplicate.

## Tests

`tests/test_cex_analysis_messages.py` — batched sibling gets a result, the lone `verify_spec` case still gets exactly one, an already-answered sibling is not answered twice, and the caller's history is not mutated. The first of those fails on master.

Local: `508 passed, 9 skipped, 9 deselected`; `pyright` 0 errors.

## Validating by eye

The end-to-end path is covered by the replay harness rather than a live run:
`uv run --no-sync python -m pytest tests/ -m "not expensive" -k autoprove` exercises the `verify_spec` → counter-example-analysis round trip against the taped LLM.

### DIFF
--- a/composer/prover/analysis.py
+++ b/composer/prover/analysis.py
@@ -1,3 +1,5 @@
+from typing import Sequence
+
 from langchain_core.messages import ToolMessage, HumanMessage, AIMessage, BaseMessage, AnyMessage
 from langchain_core.runnables import Runnable
 from langchain_core.language_models.base import LanguageModelInput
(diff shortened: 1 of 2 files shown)


shellygr:
Closing this in favour of #90, which removes the trigger at source rather than downstream of it.

That PR adds a guard at the top of `verify_spec` (`composer/spec/source/prover.py`): if the assistant turn that invoked it contains any other tool call, the tool returns an instruction to retry alone and never reaches the prover. The turn shape this PR was written for — a judge call batched with `verify_spec` — therefore can't produce a history with an unanswered `tool_use` in the first place.

That covers the whole reachable surface, not just the observed case: reusing the agent's live history is specific to `TrivialFanoutCexHandler` (`DefaultCexHandler`), and `verify_spec` is the only tool that binds it. The codegen path uses `AgenticCexHandler`, which spawns sub-agents instead. The guard's assumption also holds — monitors are applied in the tool_result node, after the tools node, so nothing can be appended after the assistant message while its tool calls are still executing.

And to be fair to what this PR actually does: it is not a general repair of the tool_use/tool_result pairing. It only fills in unanswered calls belonging to the *trailing* assistant turn — the same single shape #90's guard rejects, patched at a different site. With that guard in place it is redundant rather than defence in depth, so there is no reason to carry both.

One thing worth keeping in view: the failure remains reachable until #90 merges and is deployed, and it surfaces only *after* a prover run has completed — so an occurrence costs the whole CVL-generation phase plus the prover time that preceded it. Worth weighing when scheduling that rollout.

The analysis lives on `shelly/cex-analysis-sibling-tool-results` if anyone wants the test cases later.

PR #152, opened 2026-08-12
serialize the edit-cvl tool

naftali-g:


graphcore
naftali-g, line 1: Once the graphcore PR is done and merged I'll update this to pint at the "real" master.

PR #153, opened 2026-08-12
Fix the stuck-rule nag detector crashing every stuck prover run

naftali-g:
The loop in `verify_spec` that decides whether to warn the author "this rule has failed identically 3 runs in a row" carried three defects since it landed:

1. it iterated `stuck_count.keys()` while deleting from it, raising `RuntimeError: dictionary changed size during iteration`;
2. it never decremented `history_ind` on the `run` branch, so it re-counted the *same* prover run instead of walking back through history;
3. `del stuck_count[r]` for previously-nagged rules `KeyError`s when a rule that was nagged earlier is not stuck now.

1 and 2 compound. The tally starts at 1 for the current run, the same run is counted repeatedly, and on the third pass the rule trips the threshold and is deleted mid-iteration — so *any* rule coming back TIMEOUT/ERROR/SANITY_FAILED crashed the whole CVL-generation task on its first occurrence. The detector could never have fired as intended.

Lift the loop out of the closure into a module-level `stuck_rule_warnings` (it was already a pure function of the history) and fix all three: iterate a key snapshot, advance the index once per item, `pop(r, None)` for nagged rules. Behaviour is otherwise as originally intended — 3 consecutive identical failures, nag markers restart the streak, rule-scoped re-runs stay transparent to rules they did not exercise.

### DIFF
--- a/composer/spec/source/prover.py
+++ b/composer/spec/source/prover.py
@@ -15,7 +15,8 @@ import time
 from contextlib import contextmanager, asynccontextmanager, ExitStack, nullcontext
 from pathlib import Path
 from typing import (
-    Annotated, AsyncIterator, Callable, Iterator, override, AsyncContextManager, Sequence, Literal
+    Annotated, AsyncIterator, Callable, Container, Iterator, Mapping, override,
+    AsyncContextManager, Sequence, Literal
 )
 from typing_extensions import TypedDict, NotRequired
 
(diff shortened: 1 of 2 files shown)


composer/spec/source/prover.py
```diff
@@ -111,6 +112,69 @@ class NagMarker(TypedDict):
...
+            for r in it["nagged_rules"]:
+                # A previously-nagged rule need not be stuck now — drop it if present.
+                stuck_count.pop(r, None)
+            continue
+        assert it["sort"] == "run"
+        if it["tool_call_id"] not in known_tool_call_ids:
+            seen_post_compaction_history = True
+        target_rules = set(it["rules"]) if it["rules"] else None
+        # Snapshot the keys: the body deletes from ``stuck_count`` as streaks end.
+        for k in list(stuck_count.keys()):
```
jtoman, line 163: I thought this was just a Java whoopsie! When writing this I literally thought "do I need to worry about concurrent modification exceptions? Nah that's a java thing". *sigh*

composer/spec/source/prover.py
```diff
@@ -429,44 +493,15 @@ async def run_in(run_root: str) -> str | Command:
...
 
-            seen_post_compaction_history = False
-
-            history_ind = len(state["prover_history"]) - 1
-            while history_ind >= 0 and len(stuck_count) > 0:
-                it = state["prover_history"][history_ind]
-                if it["sort"] == "nag":
-                    for r in it["nagged_rules"]:
-                        del stuck_count[r]
-                    history_ind -= 1
```
jtoman, line 452: oh ... fuck me

tests/test_stuck_rule_warnings.py
```diff
@@ -0,0 +1,126 @@
...
+
+1. it iterated ``stuck_count.keys()`` while deleting from it, so any run that broke a
+   rule's streak crashed the whole CVL-generation task with ``RuntimeError: dictionary
+   changed size during iteration``;
+2. it never advanced the history index on the ``run`` branch, so it re-counted the *same*
+   prover run until the threshold tripped — every first stuck run nagged as though it had
+   already failed identically three times;
+3. it did ``del stuck_count[r]`` for previously-nagged rules, which ``KeyError``s whenever
+   a rule that was nagged before isn't stuck now.
+"""
```
jtoman, line 14: OKAY WE DON'T NEED TO SHAME ME IN COMMENTS OKAY

jtoman (approved):
sorry...

PR #154, opened 2026-08-12
add `--extra-context`

naftali-g:
Added in a similar manner to the `--threat-model` flag. It supports being used multiple times, and one can provide either a path to a file or to a directory (in which case all relevant files in that directory will be collected).

Also fixed a bunch of missing whitespace/newlines in some prompt snippets.

### DIFF
--- a/composer/foundry/entry.py
+++ b/composer/foundry/entry.py
@@ -78,6 +78,10 @@ class FoundryArgs(ExtendedModelOptions, FoundryRAGDBOptions, Protocol):
     def threat_model(self) -> None:
          ...
 
+    @property
+    def extra_context(self) -> None:
+         ...
+
 
 # ---------------------------------------------------------------------------
 # Entry point
@@ -135,7 +139,7 @@ def _build_parser() -> argparse.ArgumentParser:
     parser.add_argument("--forge-timeout-s", type=int, default=600, help="Per-`forge test` invocation timeout in seconds (default: 600)")
     parser.add_argument("--budget", default=None, help="Path to a run-budget file (JSON or YAML): {total: USD, caps: {phase: USD, ...}}. Omit to run unbudgeted.")
     parser.add_argument("--time-budget", default=None, type=float, help="Total wall time to run the entire execution. Omit to run without in process limit")
-    parser.set_defaults(threat_model=None)
+    parser.set_defaults(threat_model=None, extra_context=None)
     return parser
 
 
--- a/composer/foundry/entry.py
+++ b/composer/foundry/entry.py
@@ -23,7 +23,7 @@ from typing import Annotated, AsyncIterator, Awaitable, Callable, Protocol, cast
 
 from composer.core.user import get_uid
 from composer.diagnostics.timing import RunSummary
-from composer.input.parsing import Arg, add_protocol_args
+from composer.input.parsing import Arg, add_extra_context_args, add_protocol_args
 from composer.input.types import DEFAULT_RECURSION_LIMIT, RAGDBOptions, ExtendedModelOptions
 from composer.io.multi_job import HandlerFactory
 from composer.io.thread_logging import RunDataLogger
@@ -73,13 +73,10 @@ class FoundryArgs(ExtendedModelOptions, FoundryRAGDBOptions, Protocol):
     max_forge_runners: int
     budget: str | None
     time_budget: float | None
+    extra_context: list[str] | None
 
     @property
     def threat_model(self) -> None:
-         ...
-
-    @property
-    def extra_context(self) -> None:
          ...
 
 
@@ -139,7 +136,8 @@ def _build_parser() -> argparse.ArgumentParser:
     parser.add_argument("--forge-timeout-s", type=int, default=600, help="Per-`forge test` invocation timeout in seconds (default: 600)")
     parser.add_argument("--budget", default=None, help="Path to a run-budget file (JSON or YAML): {total: USD, caps: {phase: USD, ...}}. Omit to run unbudgeted.")
     parser.add_argument("--time-budget", default=None, type=float, help="Total wall time to run the entire execution. Omit to run without in process limit")
-    parser.set_defaults(threat_model=None, extra_context=None)
+    add_extra_context_args(parser)
+    parser.set_defaults(threat_model=None)
     return parser
 
 
--- a/composer/input/files.py
+++ b/composer/input/files.py
@@ -28,7 +28,7 @@ import pathlib
 import zlib
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Protocol, overload
+from typing import Protocol, Sequence, overload
 
 from composer.llm.types import CacheLevel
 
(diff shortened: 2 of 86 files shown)


composer/pipeline/cli.py
```diff
@@ -147,6 +147,12 @@ def max_concurrent(self) -> int:
     def threat_model(self) -> str | None:
         ...
 
+    @property
+    def extra_context(self) -> str | None:
+        """Path to a free-form document (text or PDF) of user-supplied background about
+        the application, fed to property inference alongside the threat model."""
+        ...
```
jtoman: I'm confused; what would be in here that isn't in the system doc?

composer/foundry/entry.py
```diff
@@ -135,7 +139,7 @@ def _build_parser() -> argparse.ArgumentParser:
     parser.add_argument("--forge-timeout-s", type=int, default=600, help="Per-`forge test` invocation timeout in seconds (default: 600)")
     parser.add_argument("--budget", default=None, help="Path to a run-budget file (JSON or YAML): {total: USD, caps: {phase: USD, ...}}. Omit to run unbudgeted.")
     parser.add_argument("--time-budget", default=None, type=float, help="Total wall time to run the entire execution. Omit to run without in process limit")
-    parser.set_defaults(threat_model=None)
+    parser.set_defaults(threat_model=None, extra_context=None)
```
jtoman: unclear why we'd not include this with foundry. For that matter, I'm not entirely sure why we excluded the threat model (that's past me's puzzling decision)

composer/pipeline/core.py
```diff
@@ -498,7 +503,7 @@ async def _one(feat: U) -> _Batch[U] | None:
                 CacheablePropertyGenerationInput(
                     "certora:system-doc", "generic", "always",
                     lambda cache, doc=design_doc: [
-                        "For reference, the system document describing the entire application is as follows.",
+                        "For reference, the system document describing the entire application is as follows.\n\n",
```
jtoman, line 502: so yeah, I noticed this. It's probably harmless; but I am genuinely uncertain if the "missing whitespace" is just langgraph/langchain being stupid about its `text` function or whether `["hello", "world"]` sent to claude actually does show up as "helloworld" to the tokenizer.

Like I said, adding the newlines just for a nicer UI experience is perfectly fine, but I am curious whether we've been sending weirdly garbled text this whole time...

composer/spec/source/autoprove_common.py
```diff
@@ -81,7 +82,8 @@ async def _entry_point(summary: RunSummary) -> AsyncIterator[Executor]:
     parser.add_argument("--memory-ns", default=None, help="Memory namespace (default: thread id)")
     parser.add_argument("--cloud", action="store_true", help="Run prover jobs in the cloud")
     parser.add_argument("--interactive", action="store_true", help="Interactively refine the security properties after extraction")
-    parser.add_argument("--threat-model", type=str, default=None, help="Path to a 'thread' model (text or pdf) with which to seed the property extraction process")
+    parser.add_argument("--threat-model", type=str, default=None, help="Path to a 'threat' model (text or pdf) with which to seed the property extraction process")
+    parser.add_argument("--extra-context", type=str, default=None, help="Path to a document (text or pdf) of any extra information about the application — protocol notes, deployment assumptions, audit scope — to hand to the property extraction process as background")
```
jtoman: see previous comments about why we think this is a distinct document from the system document/threat model.

I guess if we're going to go from 1 context document (the system document) to 2 (system document + extra context), we might as well go whole hog and make this an unbounded list of inputs right? For when we inevitably find a protocol which has *three* documents (protocol_design.md, deployment.md, algorithm_notes.md or whatever)

composer/spec/prop_inference.py
```diff
@@ -409,14 +428,28 @@ def to_cache_level(s: bool) -> CacheLevel:
...
         ))
+    if extra_context is not None:
+        actual_extra_input.append(CacheablePropertyGenerationInput(
+            "certora:extra_context", "generic", "always",
+            provide=lambda cache: [
+                "In addition, the user requesting this analysis has provided the following extra context about "
+                "the application: notes, assumptions, deployment details, or anything else they considered "
+                "relevant. It was written for the entire application (not just the component you are analyzing), "
+                "so parts of it may not bear on your analysis. Treat it as authoritative information about how "
+                "the system is intended to work and be deployed, and let it inform which behaviours count as "
```
jtoman: commonwealth spelling jumpscare

Excuse me, I mean "jumpscaure"

jtoman (commented):
code seems fine, basic design tho

composer/pipeline/cli.py
naftali-g: I _think_ this will be used to allow the user to provide feedback on a run, then rerun with that feedback. right @shellygr ?

composer/pipeline/core.py
naftali-g: yeah I don't know what langchain does, but for ap-trail this is nice to have.

composer/spec/source/autoprove_common.py
naftali-g: Good idea, will do.

composer/input/files.py
```diff
@@ -95,6 +95,59 @@ def file_digest(path: pathlib.Path) -> str:
...
+def resolve_document_paths(paths: Sequence[str] | None) -> list[pathlib.Path]:
+    """Resolve extra-context selections into an ordered document list. Each entry is
+    either a document, taken as-is, or a directory, swept by :func:`discover_documents`
+    and spliced in place.
+
+    Order is fully determined — entries in the order given, each sweep sorted by name —
+    because the result feeds both the prompt and a cache key. Raises ``ValueError`` for a
+    path that does not exist, or a swept directory with no supported documents."""
+    out: list[pathlib.Path] = []
+    for raw in paths or []:
```
jtoman: return return [] `if not paths`. what a weird way to accomplish that lol

composer/input/files.py
```diff
@@ -95,6 +95,59 @@ def file_digest(path: pathlib.Path) -> str:
...
+        if p.is_dir():
+            found = discover_documents(p)
+            if not found:
+                raise ValueError(
+                    f"no supported documents directly under {p} (looked for "
+                    f"{', '.join(sorted(DOCUMENT_SUFFIXES))}; the sweep is not recursive)"
+                )
+            out.extend(found)
+        elif p.is_file():
+            out.append(p)
```
jtoman, line 153: probably worth doing the suffix sniff on this too right?

composer/input/files.py
```diff
@@ -95,6 +95,59 @@ def file_digest(path: pathlib.Path) -> str:
...
+            p for p in root.iterdir()
+            if p.suffix.lower() in DOCUMENT_SUFFIXES
+            and not p.name.startswith(".")
+            and p.is_file()
+        ),
+        key=lambda p: p.name,
+    )
+
+
+def resolve_document_paths(paths: Sequence[str] | None) -> list[pathlib.Path]:
```
jtoman, line 125: worth documenting whether the paths are relative, or simply "whatever the input is wrt to" (looks like the latter).

jtoman (approved):
one nit which doesn't block you landing this (I trust u)

composer/input/files.py
naftali-g: I think if the user explicitly supplied this filename, we shouldn't block the file because they used an unexpected suffix

PR #157, opened 2026-08-12
Key families

jtoman:
Until now, Cache keys had to be built with loosely defined conventions; this was usually some key construction function (defined as a "private" `_` prefixed module function) that you just kinda had to know was used to define a `CacheKey[Parent, Child]`.

This PR addresses that issue by introducing `KeyFamilies`, which declaratively tie the key generation to the expected types. This is done by declaring `KeyFamily(ParentType, ChildType, key_derivation)`.

`KeyFamily` defines `__call__(**P)`, where `P` is the signature of `key_derivation`. Thus, you are enforced (by the type system) to pass exactly the parameters needed for each key to each derivation step. This should cut down on the proliferation of imported helper functions all over the place.


ericeil (approved):
Looks good, definitely an improvement!

PR #162, opened 2026-08-14
Isolate a failing CEX analysis from the run that produced it

shellygr:
## What happened

A long dev run died at its very last stage. The prover had finished; CEX analysis
called the LLM for one counterexample, got a `500`, and the exception propagated
out of `analyze` → `run_prover` → the pipeline. Hours of verification work were
discarded for one transient server error.

```
composer/prover/analysis.py:37   analyze_cex_raw → ainvoke
anthropic.InternalServerError: Error code: 500
```

## Why one bad response is fatal

Both handlers fan their per-rule analyses out under a bare `asyncio.gather`:

- `composer/prover/core.py:313` — `TrivialFanoutCexHandler`
- `composer/prover/agentic_analyzer.py:493` — `AgenticCexHandler`

Without `return_exceptions`, the first exception aborts the gather and orphans
its siblings, and nothing above catches it. Every rule's result is lost because
one rule's *explanation* could not be produced.

## The fix

Gather with `return_exceptions=True` in both handlers and drop only the failed
rule's contribution. It keeps its prover-assigned status and loses its
explanation (trivial handler) or its root causes (agentic handler); every other
rule reports as before, and the caller still gets a rendered report.
`CancelledError` is re-raised rather than logged, so cancellation still unwinds.

One adjacent correction: the trivial handler's `failed_count` was summed over
the analyzed results, so dropping a failed analysis would quietly lower it and
change the summarization-threshold decision. It now counts violated rules.

## On retrying the 5xx

Deliberately not added. `composer/llm/anthropic.py:284` already builds
`ChatAnthropic(..., max_retries=8)`, and the SDK retries `>=500` by default — so
that `500` had already been retried eight times before it surfaced. A second
retry layer would be redundant, and would not have saved this run. The blast
radius was the problem, not the frequency.

## Testing

`tests/test_cex_analysis_failure_isolation.py` — one rule's analysis raising
still yields the others' explanations; all of them raising still renders a
report; `CancelledError` still propagates. The first two fail against `master`
and pass here.

Full pass: `693 passed, 9 skipped, 9 deselected`, `pyright` 0 errors.

### DIFF
--- a/composer/prover/agentic_analyzer.py
+++ b/composer/prover/agentic_analyzer.py
@@ -29,6 +29,7 @@ and the actual source compiled into the verification problem.
 
 import asyncio
 from dataclasses import dataclass
+import logging
 from pathlib import Path
 from typing import Annotated, Literal, NotRequired, Union, override
 import uuid
@@ -56,6 +57,8 @@ from composer.templates.loader import load_jinja_template
 from composer.diagnostics.timing import set_current_task_id
 from composer.prover.cex_task_ids import cex_rule_task_id, cex_aggregator_task_id
 from composer.tools.thinking import RoughDraftState, get_rough_draft_tools
+
+_logger = logging.getLogger(__name__)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jtoman (approved):
FWIW, a rerun with the same parameters should pick up basically where it left off.

PR #164, opened 2026-08-14
autosetup: express remapping contexts against the run root

shellygr:
## The bug

A remapping's two halves are matched against different things:

- the **target** is a filesystem path — absolute or cwd-relative, both resolve;
- the **context** is a prefix solc matches against the *source unit name* of the importing
  file, i.e. the path as it appears in the conf's `files`, relative to the directory
  certoraRun runs from.

`build_packages_from_remapping_sources` resolved the target against `base_dir` but copied the
key verbatim, and `forge remappings` reports contexts relative to the Foundry project dir. When
the project is a subdirectory of the run root (`<repo>/<chain-dir>/foundry.toml`, one bundle per
subdirectory under it), a reported context `src/Widget_1234/` never prefixes the source unit name
`<chain-dir>/src/Widget_1234/…`, so **no context ever matches and every package for that
sub-project is inert**. The failure is confusing because the conf plainly contains the right
mapping and the target directory really exists:

```
ParserError: Source "@openzeppelin/contracts/utils/ReentrancyGuard.sol" not found:
File not found. Searched the following locations: "".
```

The empty search-location list is the tell — solc had no applicable remapping at all.

`source_not_found_packages` cannot rescue it either: since the builder was unified it recomputes
the identical list and stops with *"Workarounds applied … but the conf and command are unchanged
— retrying would fail identically, giving up"*.

## The fix

`_rebase_context` re-expresses the context against the run root, which the three call sites
(`foundry.py`, `truffle.py`, `compilation_workarounds.py`) already hold as `self.project_root`.

A context is rebased only when `base_dir/context` is a real directory — that is what separates a
project-relative context from one already written against the run root, and it avoids inventing a
doubled prefix. A context resolving outside the run root cannot be named by any source unit name,
so it is kept as authored with a WARNING.

**When the build config sits at the run root — every flat project — the rebasing is the
identity**, so nothing changes for the vast majority of runs.

## Tests

Five new cases in `tests/test_remappings.py`: nested project rebases, flat project is
byte-identical with and without `run_root`, a context naming no directory is left alone, a
context outside the run root warns and is left alone, and `parse_config` end-to-end.

`pytest -m "not expensive"`: 700 passed, 9 skipped. `pyright`: 0 errors.

Fleet: a new corpus entry reproducing this layout is queued through the mass-test rig — a
before/after on it plus a full-corpus regression run. Results will be posted here before merge.

### DIFF
--- a/certora_autosetup/utils/remappings.py
+++ b/certora_autosetup/utils/remappings.py
@@ -182,8 +182,9 @@ def _rebase_context(context: str, base_dir: Path, run_root: Optional[Path], log_
 
     A context is rebased only when ``base_dir/context`` is a real directory, which is what tells
     a project-relative context apart from one that is already run-root-relative. Anything else
-    (no ``run_root``, a context that resolves outside ``run_root``, a context naming no directory)
-    is returned untouched: those are shapes this cannot improve on, so leave them alone.
+    (no ``run_root``, a context naming no directory, a context resolving to the run root itself
+    or outside it) is returned untouched: those are shapes this cannot improve on, so leave them
+    alone. Only the outside-the-run-root case warns — it is the one that cannot work at all.
     """
     if run_root is None or not context:
         return context
@@ -193,7 +194,13 @@ def _rebase_context(context: str, base_dir: Path, run_root: Optional[Path], log_
         return context
 
     rebased = os.path.relpath(resolved, run_root)
-    if rebased == os.curdir or rebased.startswith(os.pardir):
+    if rebased == os.curdir:
+        # The context IS the run root, so it already covers every source unit name. Its faithful
+        # translation is a global, context-free remapping — but promoting a scoped remapping to
+        # global would let it shadow correct global mappings (solc ranks longest matching context
+        # first), so keep it as authored. Nothing is wrong here, hence no warning.
+        return context
+    if rebased.startswith(os.pardir):
         log_fn(
             f"Remapping context '{context}' resolves outside the run root {run_root} "
             f"— leaving it as is",
--- a/certora_autosetup/utils/remappings.py
+++ b/certora_autosetup/utils/remappings.py
@@ -11,7 +11,7 @@ import json
 import os
 import subprocess
 from pathlib import Path
-from typing import Callable, Dict, List
+from typing import Callable, Dict, List, Optional
 
 import tomllib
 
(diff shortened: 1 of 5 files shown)


shellygr:
### Verification

Two independent checks, both against a clean `origin/master` build of the same image.

**1. Mass-test rig, whole-scene path.** A repo whose Foundry project is a subdirectory, with one
context-scoped remapping set per bundle:

| contract | `origin/master` | this branch |
|---|---|---|
| a bundle contract reachable by detection | failed — `ParserError … Searched the following locations: ""` | **ok** |

Its recovery path is worth recording: `compiler_version_mismatch` → `source_not_found_packages` →
`use_relpaths_for_solc_json` → compiled. On master the second of those matched but produced an
*unchanged* conf, so the loop stopped with "retrying would fail identically, giving up" and never
reached the last-resort. With contexts rebased the same workaround now genuinely changes the conf,
so the escalation continues.

**2. The main contract, cloud-shaped conf.** Reproducing the invocation the cloud run died on
(`files` = main contract + the mock autosetup copies into `certora/mocks/`, run from the repo root):

| conf | `origin/master` | this branch |
|---|---|---|
| as-is | `ParserError` | `ParserError` |
| `+ use_relpaths_for_solc_json` | `ParserError` | `CompilerError: Stack too deep …` (imports resolved) |
| `+ viaIR + optimizer` | `ParserError` | **rc=0, autofinders generated** |

The middle row is the mechanism, and it is worth knowing: solc matches a context against the
**source unit name**, and certoraRun names sources absolutely unless `use_relpaths_for_solc_json`
is set. So context-scoped remappings — correct or not — are inert until that workaround fires.
The stack-too-deep in the third row is a property of that contract's deployment settings (viaIR +
optimizer), which autosetup reaches through `stack_too_deep_via_ir` / `yul_exception_add_optimizer`.

A full-corpus regression run on this branch is in flight; I'll post the diff when it finishes.

certora_autosetup/utils/remappings.py
```diff
@@ -148,12 +161,51 @@ def build_packages_from_remapping_sources(base_dir: Path, log_fn: LogFn, profile
...
+    """
+    if run_root is None or not context:
+        return context
+
+    resolved = Path(context) if Path(context).is_absolute() else base_dir / context
+    if not resolved.is_dir():
+        return context
+
+    rebased = os.path.relpath(resolved, run_root)
+    if rebased == os.curdir or rebased.startswith(os.pardir):
```
jar-ben: -Just based on reading the code - when `resolved` equals `run_root`, `os.path.relpath` returns `"."`, i.e. `os.curdir`. The code logs a "resolves outside the run root" warning — but that case is the run root itself, not something outside it. The message is misleading there. Or am I reading it wrong?

jar-ben (approved):

certora_autosetup/utils/remappings.py
shellygr: Claude answers: You read it right — `"."` means the context *is* the run root, not outside it. Fixed in ba7fb1c: that case now returns the context untouched with no warning (it is left as authored only so a scoped remapping is not promoted to a global one, which could shadow correct global mappings); the warning is kept for `..`, which genuinely cannot match. Test added.

PR #169, opened 2026-08-16
autosetup: pick the build system and the artifact directory from artifacts on disk

shellygr:
**Draft** — the tie-break half needs a paired builder change before it earns its keep; see *What this
actually fixes today*.

## Two defects, both surfacing as "no contracts"

**A. The Foundry/Hardhat tie is broken by config order, not by evidence.** When a directory holds both
a `foundry.toml` and a Hardhat config, detection always answered Foundry and logged
`Both Foundry and Hardhat detected, defaulting to Foundry`. If Hardhat is what built the tree, its
output is under `artifacts/` while `out/` is empty or absent, so the extractor reads nothing and the
run dies with `No project-local logic contracts found in FoundryManager build output` — 12 contract
rows across 6 projects in the last corpus sweep. Detection now prefers the build system whose artifact
directory actually holds artifacts. Foundry keeps the tie when it has artifacts of its own and when
neither side has any (an unbuilt tree offers nothing to rank by), and `--build-system` still wins
outright.

**B. A configured artifact directory was ignored when the default name existed as a parent.** The
extractor tested `out/` for existence, so a project with `out = "out/foundry"` found the (empty)
parent and reported that the build produced nothing. It now takes the default only while it *holds*
artifacts, else the directory the config names.

Evidence is one predicate per build system — `holds_artifacts` — and all of it requires a *file*, not
a directory: a `*.sol/` dir for Foundry, a `*.json` under `contracts/`or `build-info/` for Hardhat, a
flat `*.json` for Truffle. A configured-but-never-run build leaves the directories behind, so
directory existence is not evidence that a build happened.

## What this actually fixes today

Measured in the fleet image against the real clones:

- **B fixes 2 of the 12 rows** (`reconcopy-mitosis`: `rc=1 No project-local logic contracts` →
  `rc=0`, 39 contracts found in `out/foundry`).
- **A fixes 0 of them by itself** — in all six projects Foundry *does* have artifacts, because the
  builder only ever runs `forge build` there, so Hardhat never gets a chance to leave any. A is the
  precondition for fixing that on the builder side; on its own it changes no corpus project's answer
  (measured across all 314 clones: 87 hold both configs, 0 flip).
- One project (`sales-compound-multiplier`) fails for an unrelated reason — its `node_modules/` is
  empty, so `forge build` produced a single interface with no bytecode. No detection change can
  recover that.

## Known limits, deliberately not fixed here

- The tie-break reads Foundry's `out` from `[profile.default]`, so a tree built under a non-default
  `FOUNDRY_PROFILE` with a different `out` reads as having no artifacts. Documented in the docstring.
- A Hardhat project whose `foundry.toml` governs a forge *test* harness looks like Foundry the moment
  that harness is compiled. Also documented.
- `find_build_config_dir` still ranks Foundry first for a directory holding both configs, so it can
  disagree with `detect()` on a monorepo package that only Hardhat built. Worth closing, separately.

## Tests

`tests/test_build_system_choice.py` — 15 cases: every single-build-system path unchanged (built and
unbuilt), both-configs-with-Foundry-artifacts unchanged, both-unbuilt unchanged, the tie-break itself,
empty artifact directories rejected on both sides, evidence read from a configured `out`,
`--build-system` still authoritative, plus the extractor's resolution and error-message cases.

`pytest -m "not expensive"`: 759 passed, 9 skipped. `pyright`: 0 errors.


certora_autosetup/parsers/base.py
```diff
@@ -80,24 +80,46 @@ def extract_logic_contracts_impl(self, artifacts_dir: Path) -> List[ContractHand
...
 
-    def extract_logic_contracts(self) -> List[ContractHandle]:
-        """Extract logic contracts from build artifacts with config fallback."""
-        # Get default artifact directory from manager
+    def resolve_artifacts_dir(self) -> Optional[Path]:
+        """Return the directory holding this project's build artifacts, or None if there is none.
+
+        The default directory name is only a guess. A project can configure its output
+        elsewhere (Foundry's ``out``, Hardhat's ``paths.artifacts``) and still have a
+        directory by the default name, most often as the *parent* of the configured one:
```
shellygr:
```suggestion
        directory by the default name, for example as the *parent* of the configured one:
```

certora_autosetup/utils/project_dir.py
```diff
@@ -27,37 +27,64 @@
...
 
-# Truffle writes artifacts here unless the config sets `contracts_build_directory`. Reading
-# that override means running node against the config, which is more than this needs: it
-# only has to recognise a built project, and the manager resolves the real path later.
+# Hardhat writes artifacts here unless the config sets `paths.artifacts`, and Truffle
+# unless it sets `contracts_build_directory`. Reading either override means running node
+# against the project's own config (see `HardhatManager._extract_config_via_node`), which
+# is more than these helpers need: they only have to recognise a built project, and the
+# managers resolve the real path later. Foundry's `out` is the one override read here,
+# because a TOML parse costs nothing.
```
shellygr:
```suggestion
# managers resolve the real path later. Foundry's `out` is the one override read here,
# because a TOML parse is easy.
```

certora_autosetup/parsers/base.py
```diff
@@ -80,24 +80,46 @@ def extract_logic_contracts_impl(self, artifacts_dir: Path) -> List[ContractHand
...
+        extractor can read. So the default only counts while it actually holds artifacts,
+        and the config decides otherwise. The config is the second question because it is
+        the expensive one — Hardhat and Truffle answer it by running node.
+
+        A default directory that exists but holds no artifacts is still the right answer
+        when the config offers nothing better: it means the build produced nothing, which
+        is what the caller should report on.
+        """
         default_artifact_dir = self.project_root / self.manager.get_default_artifact_dir()
+        if self.manager.holds_artifacts(default_artifact_dir):
```
jar-ben, line 113: Why do we prioritise the default dir over the dir specified in the config?

certora_autosetup/parsers/base.py
```diff
@@ -80,24 +80,46 @@ def extract_logic_contracts_impl(self, artifacts_dir: Path) -> List[ContractHand
...
-        if not artifacts_dir.is_dir():
-            raise Exception(f"'{artifacts_dir}' exists but is not a directory")
+    def extract_logic_contracts(self) -> List[ContractHandle]:
+        """Extract logic contracts from build artifacts with config fallback."""
+        artifacts_dir = self.resolve_artifacts_dir()
+        if artifacts_dir is None:
+            # Name the directory the caller should go look at — the one the config declares
+            # when it declares one — and say which way it is unusable, since a path that is
+            # there but is a file needs a different remedy than one that is absent.
+            missing = self._try_read_artifact_dir_from_config()
```
jar-ben: Can you rename/better-specify-name `missing`? The code is not easy to read now - the logic of `extract_logic_contracts` is based on internal logic of `resolve_artifacts_dir` (which is opaque here)

jar-ben (approved):
just minor requests

certora_autosetup/parsers/base.py
shellygr: Claude answers: Two reasons, both now in the docstring since the question will come up again.

The first is cost. Reading the config means running a subprocess: Foundry's `parse_config` ends up shelling out to `forge remappings` (`utils/remappings.py:63`, 30s timeout), and Hardhat and Truffle load the project's config through node (`hardhat.py:152`, `truffle.py:167`, 10s each). None of it is memoized, and `resolve_artifacts_dir` is reached from three places (`contract_utils.py:111` and `:172`, `setup_prover.py:249`), so each one pays again. On Hardhat and Truffle a failed read is worse than useless: `parse_config` swallows the error and hands back the default config, whose `get_artifact_directory()` returns the default name, so we'd have spent 10s to land where the cheap branch already was. The image we run in has no node installed at all.

The second is staying consistent with detection. When both configs sit side by side, `_pick_foundry_or_hardhat` ranks the Hardhat side by whether `artifacts/` holds anything, and never reads a Hardhat config (`project_dir.py:60`). If the extractor then prefers a configured directory, it can end up reading somewhere the detector never looked while the populated directory that decided the pick goes unread. That shows up as `No project-local logic contracts found`.

The order isn't free though, and you're pointing at a real gap: a project that builds into a configured `out` while a stale `out/` survives, or one whose profile sets its own `out`, gets the stale one read. I think the fix there is to rank both sides by artifacts, so a configured directory that holds artifacts beats a populated default, rather than flipping the order. That needs the config read memoized per `(project_root, profile)` and a containment guard, since `find_config_file` walks 3 parents and can hand back a sibling project's `out`. Left it for a follow-up.

One thing worth flagging: `test_a_populated_default_out_wins_over_the_configured_one` only ever created one of the two directories, so it passed under either policy and pinned nothing. It populates both now.

certora_autosetup/parsers/base.py
shellygr: Claude answers: Renaming it wouldn't have been enough. What makes it read badly is that `extract_logic_contracts` was re-running the decision: it called `_try_read_artifact_dir_from_config` a second time to recover a path `resolve_artifacts_dir` had already computed, and it could only get that right by knowing precedence rules it couldn't see. For Hardhat it also paid the node timeout twice.

So I changed the contract instead. `resolve_artifacts_dir` now always returns the directory the artifacts belong in, whether or not it exists, and the caller does one `is_dir()` check on that path and names that same path in the error. That drops the local and the `Optional`, and puts the last-resort rule (previously split across both halves) in one line. `_try_read_artifact_dir_from_config` is down to a single call site.

I checked it's behaviour-preserving across all 24 combinations of {default holds artifacts / empty / is a file / absent} x {config names a populated dir / an empty dir / a missing dir / a file / unreadable / same as default}: same path, same error text either way. Also added a test for the `is not a directory` branch, which nothing covered.

PR #171, opened 2026-08-17
Move code-explorer prompts onto the ecosystem as shared templates

ericeil:
`Language.code_explorer_prompt` was unused...and anyway was specialized badly for Rust, using a shortened prompt that was incomplete.  The sensible specialization here is per-chain, not per-language, anyway.

This change deletes the unused per-language specialization, and instead specializes this prompt per-ecosystem, assembling the final prompt from templates.  The EVM-specific prompt is unchanged from master.


chandrakananandi (approved):
Thanks for making this change. Looks reasonable to me. For @jtoman we were discussing your comment about the prompt in my PR and Eric took the initiative to clean up the prompts and also put them all in the same place.

composer/spec/code_explorer.py
```diff
@@ -5,44 +5,44 @@
...
+    prior_findings: PriorFindingsMode
 
-When complete, deliver your answer via the `result` tool.
-"""
+
+def render_code_explorer_prompt(
+    template: TypedTemplate[CodeExplorerPromptParams],
+    prior_findings: PriorFindingsMode,
+) -> str:
+    return template.bind({"prior_findings": prior_findings}).render_to(load_jinja_template)
```
jtoman: make this return `PromptInput` or whatever it is, change the body to:

```
return template.bind({"prior_findings": prior_findings}).render_to
```

and you can avoid tying to the `load_jinja_template` (preferable imo)

composer/templates/code_explorer/rust/_common.j2
```diff
@@ -0,0 +1,3 @@
+{% include "code_explorer/_common.j2" %}
+
+The project is a Rust crate or workspace. Navigate from `Cargo.toml` and the crate modules; quote the exact Rust snippets that establish or omit a check.
```
jtoman: what "check"? Why does this keep sneaking into all of these system prompt?

composer/templates/code_explorer/common_fragment.j2
```diff
@@ -0,0 +1,16 @@
+You have access to file tools (list_files, get_file, grep_files) to explore the project.
```
jtoman, line 1: idk how I feel about these `_` prefixes. I've been using the magic *suffix* `fragment` or `partial`. I'd prefer we centralize around a convention...

composer/templates/code_explorer/solidity.j2
```diff
@@ -0,0 +1,22 @@
...
+- Include relevant function signatures, state variable declarations, or code snippets in your answer.
+- If the question asks about behavior, trace through the actual implementation rather than speculating.
+- Be concise: the caller needs a dense, actionable answer, not a walkthrough of your exploration process.
+- If you discover you do not have enough information to fully answer the question, 
+  (e.g., there is a reference to code not available to you) *DO NOT GUESS*. Indicate in your final answer
+  that you cannot fully answer the question due to incomplete information.
+
+If asked a question that cannot be answered by simply looking at the code (e.g., about some completely unrelated
+topic) you must decline to answer, indicating it is out of scope for what you're capable of answering.
+
```
jtoman: why doesn't this on use `_common.j2`?

jtoman (changes requested):
a couple of major nits

composer/templates/code_explorer/rust/_common.j2
ericeil: Yeah, this made a little more sense in context before this change, but we can do better.

composer/templates/code_explorer/solidity.j2
ericeil: It should; this was me being overly zealous about not changing the EVM prompt even a little bit. :)

composer/templates/code_explorer/common_fragment.j2
ericeil: Fixed the new filenames, and also fixed some previous examples of this pattern from my previous PRs.

composer/spec/code_explorer.py
```diff
@@ -38,11 +37,13 @@ class CodeExplorerPromptParams(TypedDict):
     prior_findings: PriorFindingsMode
 
 
-def render_code_explorer_prompt(
+def code_explorer_sys_prompt(
     template: TypedTemplate[CodeExplorerPromptParams],
     prior_findings: PriorFindingsMode,
-) -> str:
-    return template.bind({"prior_findings": prior_findings}).render_to(load_jinja_template)
+) -> Callable[[TemplateLoader], str]:
```
jtoman, line 43: my brother in christ, `type PromptInput = Callable[[TemplateLoader], str] | ...` and `with_sys_prompt` takes the `PromptInput` type. Why u spell out the callable type like this

tests/test_code_explorer_prompts.py
```diff
@@ -18,26 +23,32 @@
 _MODES: tuple[PriorFindingsMode, ...] = ("none", "established", "versioned")
 
 
+def _render(
+    template: TypedTemplate[CodeExplorerPromptParams], mode: PriorFindingsMode
+) -> str:
+    return code_explorer_sys_prompt(template, mode)(load_jinja_template)
```
jtoman, line 29: oh this is why. *sigh* okay

jtoman (approved):

PR #177, opened 2026-08-19
 rules striping

jtoman:
Allow the author agent to get their spec stamped by "striping" the rules across different runs. This requires the AP infrastructure to get access to a list of all the rules in a spec file. The way this PR achieves that currently is ... not great, but can be easily swapped in for a "proper" `certora-cli` invocation whenever that comes down the pipe.

This necessitated adding an `exclude_rules` option to the certora prover tool which maps directly onto the conf flag. While updating the prompts around the `verify_spec`/prover usage, I took the opportunity to point some outdated advice to the context documents and/or delete some advice that was mooted by more thorough treatment in said docs.


composer/spec/source/prover.py
```diff
@@ -186,6 +203,48 @@ def last_prover_run(
...
+) -> bool:
+    """Whether the runs against the current authoring state collectively verify every
+    declared rule (rules expected to fail are forgiven their failures but still count
+    as covered)."""
+    remaining_rules = set(all_rules)
+    for history in _iterate_history(
+        l, curr_digest, curr_status
+    ):
+        for (k, stat) in history:
+            if stat != "VERIFIED" and k.rule not in expected_to_fail:
```
shellygr, line 238: if a rule is expected to fail but yielded timeout, doesn't it mean it's not covered?

shellygr (approved):
one nit

composer/spec/source/prover.py
jtoman: "expected to fail" is "not verified". the new prompt suggests that the agent not use expected to fail for "the rule is broken" but rather "the rule demonstrates a potential bug". We can, if you'd like, allow only "VIOLATED" results for "expected_to_fail" rules if you think that's more coherent?

PR #180, opened 2026-08-19
autosetup: try the optimizer under legacy codegen before reaching for via-ir

shellygr:
## What was wrong

Solc's stack-too-deep message names two remedies: compile with `--via-ir` *while enabling the
optimizer*. The ladder acted on the first half only.

`stack_too_deep_via_ir` (`compilation_workarounds.py:385`) answers `CompilerError: Stack too deep` by
switching via-ir on for the failing contract. The optimizer rung sits behind
`_detect_yul_exception_stack_too_deep`, whose regex requires a literal `YulException:` prefix, and
only the via-ir pipeline emits that. So the loop tried legacy, then via-ir with the optimizer, and
never tried legacy with the optimizer.

That matters because via-ir is not a free substitution. It inlines internal functions and takes the
CVL internal summaries with them, which is why `build_systems/base.py:137` refuses to inherit a
project's declared `via_ir` in the first place.

## What changed

Three rungs now precede via-ir on a legacy stack-too-deep:

1. `stack_too_deep_optimizer` enables the optimizer, keeping legacy codegen.
2. `stack_too_deep_autofinder` handles the case where only the autofinder-instrumented compile is
   over the stack limit. The contracts themselves compile; accepting the finder fallback for those
   files costs local-variable finders there and keeps every contract on legacy codegen.
3. `stack_too_deep_via_ir`, as before, per contract, and only once the two above have been tried.

Per-contract via-ir also stops walking a large scene one compile at a time. Past ten contracts, or
when the project's own build config declares via-ir, the remaining contracts are switched together.
The declared value is still never inherited: it only says where the walk ends, and the optimizer
value emitted is ours.

The optimizer apply is shared with the Yul rung rather than duplicated, so Foundry's
`compilation_restrictions` map keeps its explicit runs values and the scalar is never written beside
the map.

## Measurements

Three experiments ran before the code, in the fleet image, compilation analysis only.

| | before | after |
|---|---|---|
| 43-contract project | ladder ends with via-ir on 1 contract, fails | **rc=0 in 7m40s**, `solc_via_ir_map` all False |
| 171-contract project | ladder ends with via-ir on 7 contracts, fails | optimizer then autofinder rung, no via-ir |
| 105-contract project | 18 accretion passes, 60-minute cap, no result | scene-wide via-ir at our own `solc_optimize 200`, rc=0 in 13m32s |

The 43-contract run went through the real ladder end to end, not a hand-built conf: exactly two rungs
fired, `stack_too_deep_optimizer` then `stack_too_deep_autofinder`, and certoraRun reported
`all solc_via_ir_map values are set to False`.

Two findings worth carrying forward:

- Legacy plus optimizer is **not** enough on its own. With `assert_autofinder_success: True`, which is
  what `setup_prover.py:409` emits, both projects still fail: 35 of 171 and 15 of 43 files fall back
  during autofinder instrumentation. That is why rung 2 exists, and it is a real cost, paid instead of
  the larger one of inlining every contract.
- The declared `optimizer_runs` values (10,000,000 and 100,000) make no difference. 200 compiles the
  same scenes, so nothing needs to be inherited from the project.

## Tests

`pytest -m "not expensive"`: 978 passed, 11 skipped. `pyright`: 0 errors.

Five new cases pin the ladder: the optimizer fires first and sets no via-ir; via-ir follows and stays
scoped to the failing contract; the autofinder rung relaxes the assertion without touching codegen;
the scene flips past the threshold; a declared `via_ir` skips the walk. Three existing tests moved by
one compile each, since the ladder now spends a pass on the optimizer before via-ir. Their assertions
were tightened to the new sequence rather than loosened.

### DIFF
--- a/certora_autosetup/utils/compilation_workarounds.py
+++ b/certora_autosetup/utils/compilation_workarounds.py
@@ -32,6 +32,15 @@ from certora_autosetup.utils.solc_version_resolver import (
 )
 from certora_autosetup.utils.types import ContractHandle
 
+# Solc's legacy-codegen stack-too-deep, as opposed to the YulException the via-ir pipeline
+# raises. Matched against whitespace-normalized output, since solc hard-wraps diagnostics.
+_STACK_TOO_DEEP_RE = re.compile(r"CompilerError:\s*Stack too deep", re.IGNORECASE)
+
+# How many contracts may be given via-ir one at a time before the rest of the scene is
+# switched with them. Each individual escalation costs one full compile of the scene to
+# discover, so this bounds the walk; below it, contracts keep their summaries.
+VIA_IR_SCENE_THRESHOLD = 10
+
 
 class AbstractMainContractError(Exception):
     """Raised when the main (verify-target) contract compiled to no bytecode — it is
@@ -141,6 +150,7 @@ class CompilationWorkaroundManager:
         verbose: int = 0,
         solc_convention: SolcConvention = SolcConvention.CERTORA,
         build_config_dir: Optional[Path] = None,
+        declared_via_ir: bool = False,
     ):
         self.project_root = project_root
         # Where foundry.toml / remappings.txt / package.json are read from. Distinct from
@@ -163,6 +173,13 @@ class CompilationWorkaroundManager:
         # to the scalar on finalize — an unseeded partial map (e.g. a single cancun
         # entry) must never be promoted to a global scalar.
         self._evm_version_seeded = False
+        # Contracts given via-ir one at a time, for the scene-wide threshold in
+        # _apply_via_ir_workaround.
+        self._via_ir_contracts: Set[str] = set()
+        # What the project's own build config asks for. The value is never inherited into
+        # the conf (build_systems/base.py drops it deliberately), but it does say where a
+        # via-ir escalation is going to end up, so the walk can be skipped.
+        self.declared_via_ir = declared_via_ir
         # Convert default version to the detected convention
         if solc_convention == SolcConvention.SOLC_SELECT and solc_default_version.startswith("solc") \
                 and not solc_default_version.startswith("solc-"):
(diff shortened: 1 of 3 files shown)


certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -443,9 +426,52 @@ def run_compilation_with_workarounds(
...
+                    if self._detect_stack_too_deep_errors(output, contracts) is not None
+                    and self._yul_optimizer_pending(compilation_config, contracts)
+                    else None
+                ),
+                apply_fn=self._apply_optimizer,
+                enabled=not global_via_ir_enabled,
+            ),
+            # The optimizer can clear the contract compile and still leave the
+            # autofinder-instrumented one over the stack limit, since instrumentation adds
+            # slots of its own. Accepting the fallback costs local-variable finders in the
```
shellygr: this is only an example. localvars can actually pass. or they can fail. same with internal summaries. it's not clear-cut that enabling via-ir kills both always.

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -443,9 +426,52 @@ def run_compilation_with_workarounds(
                 # contract's entry from its pragma is always safe.
                 enabled=True,
             ),
+            # solc's own advice on a legacy stack-too-deep is to compile via-ir "while
+            # enabling the optimizer" — the optimizer is what reclaims stack slots, and it
+            # does so under legacy codegen too. Enabling it alone keeps legacy codegen,
+            # which via-ir would replace: via-ir inlines internal functions and with them
+            # the internal summaries CVL applies, so it is the more expensive answer.
```
shellygr: it's a maybe, not a definite, but a risk we prefer to avoid

certora_autosetup/utils/compilation_workarounds.py
```diff
@@ -1698,11 +1701,37 @@ def apply_compiler_version_workaround(
...
 
     def _apply_via_ir_workaround(self, contract_needing_via_ir: str, config_dict: Dict) -> Dict:
-        """Add solc_via_ir_map entry for contract that needs via-ir compilation."""
-        # solc_via_ir_map is seeded up front by _seed_compile_maps; just set the
-        # contract that needs via-ir to True.
+        """Enable via-ir for the contract that needs it — or, past the point where naming
+        them one at a time pays off, for the whole scene.
+
+        Per contract is the better answer while the count is small: every contract left on
+        legacy codegen keeps the internal-function summaries via-ir would inline away. But
```
shellygr:
```suggestion
        legacy codegen keeps the internal-function summaries via-ir might inline away. But
```

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: Reworded. It now says via-ir inlines internal functions, "which can leave the internal summaries CVL applies with nothing to attach to", and that it does not always cost them but is a risk worth not taking when the optimizer alone may do. 0fe823b0

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: Fixed. The comment now puts both sides as exposures rather than losses: falling back puts the local-variable finders at risk, via-ir puts the internal summaries at risk, and neither is certain, since a file can fall back and still give useful finders and an inlined contract can still be summarized. What orders the rungs is that the via-ir exposure is the wider one. 0fe823b0

certora_autosetup/utils/compilation_workarounds.py
shellygr: Claude answers: Applied verbatim. I also changed "the finder loss is already broad" two lines down to "the exposure is already broad", for the same reason. 0fe823b0

jar-ben (approved):

PR #97, opened 2026-07-23
PR 2/3: Rust application framework (PyO3)

ericeil:
## PR 2 of 3 — Rust application framework (PyO3)

Part of the stacked split of `eric/crucible`.
**Stack:** `master` → ~~`eric/ecosystem`~~ (#96, **merged**) → **`eric/rust`** → `eric/crucible-app`.
**Base:** `master`. 79 files, +10.7k/-278.

Everything here serves one goal: **let an AutoProver application be implemented in Rust, as requested by the Solana Foundation.  This PR is the host that makes that possible. It carries no verification backend; the
first one (Crucible) is PR 3.

### What this adds

**The generic wheel host — `composer/rustapp/`**
A wheel supplies a declarative `AppDescriptor` and answers pure callouts; the host owns all control
flow, all effects, and the entire vertical an application needs — argparse, service setup, pipeline
wiring, frontend, artifact store, `main()`.
*Why: a Rust application should only have to describe itself and answer questions. Everything a
wheel would otherwise reimplement in Python lives here once.*

**A typed ABI on both sides — `descriptor.py` (load time) and `wire.py` (runtime)**
The descriptor is pydantic-validated at load; the runtime unions are tagged (`CompileOk |
CompileFailed`, `ValidateBuildFailed | ValidateVerdicts`) and the ten callouts are a Protocol checked
at import.
*Why: a field renamed on the Rust side fails at the boundary, instead of reading as `""` three
frames later.*

**The Rust workspace — `rust/`**
`autoprover-sdk` — the ABI serde types, the `Application` / `FormalizeSession` traits, and the
`export_app!` macro that emits the PyO3 module. `example-app` (echoprover) — a demo wheel with zero
bespoke Python. `run-confined` gains a maturin `bin` pyproject.
*Why: the Rust half of the ABI needs a home, and the framework needs something real to round-trip
against in CI.*

**Backend preflight — `composer/pipeline/core.py`**
New `PipelineBackend.preflight`: whatever a backend can do before it knows anything about the
program, run concurrently with system analysis, awaited first, its failure cancelling the analysis.
*Why: a backend that must **build** something can gate a broken workspace before the model is spent,
instead of surfacing it as unfixable compiler errors in the first authored draft.*

**`StagedFormalizer` replaces the `Formalizer.begin` hook**
`prepare_formalization` widens to a union, and the shared setup artifact becomes a constructor
argument to the only object that uses it.
[...]

PR #181, opened 2026-08-19
ci: split the wire round trips out of the `fuzz` marker

ericeil:
The nightly Integration Tests job died in `run template fuzz tests` the first night the wire
round-trip suite existed — [run 32218703320](https://github.com/Certora/AutoProver/actions/runs/32218703320/job/95964997297),
13 errors, one cause.

**`-m 'fuzz' tests` selects two unrelated suites, and only one of them is Python.** The 36 template
tests the step is named for, and the 13 round trips in `tests/test_wire_roundtrip.py`, whose
`wire_echo` fixture builds `wire-echo` with cargo at test time. The step runs `pytest -n 2` and that
fixture is session-scoped *per xdist worker*, so both workers asked rustup to install the pinned 1.96
toolchain at the same instant. rustup's install path is not safe against concurrent invocations: one
call rolled back (`failed to install component: 'rustc-x86_64-unknown-linux-gnu', detected conflict:
'bin/rust-gdb'`) and the other found what the rollback left behind (`the 'cargo' binary ... is not
applicable to the '1.96-x86_64-unknown-linux-gnu' toolchain`). Not a flake that turned:
`rust-toolchain.toml` and the suite both landed in #97 on 08-18, after that morning's scheduled run,
so this was the first nightly to select them — 29 fuzz tests on 08-18, 49 on 08-19.

**Installing the toolchain first would stop the crash and buy nothing.** That is what `pytest.yml`
does, for exactly this hazard, and it is the wrong answer here: this job would pay an uncached cargo
build inside a 20-minute budget to re-run tests it cannot deepen and does not uniquely cover. They
pin `max_examples=200` in their own `@settings`, so this job's `HYPOTHESIS_PROFILE: extended` never
reaches them — that profile is registered in `test_fuzzed_templates.py` alone. And `pytest.yml`
selects `-m 'not expensive'`, which includes them, on every PR and every push to master, with the
toolchain installed and `rust/target` cached.

So the marker splits by requirement rather than by kind. `wire` says nothing about what those tests
do — they are fuzz tests — only about what selecting them commits a job to: having the pinned Rust
toolchain. A job with no Rust can now ask for `fuzz` and stay that way, which is what this one wants.
The `sync-deps` comment claiming the job "doesn't compile the Rust crates" stopped being true the
moment those tests carried `fuzz`; it now says that no toolchain gets installed, and points at the
step that depends on that.

## Tests

The selector counts are the whole behavioural change:

```
-m 'fuzz'          → 36  test_fuzzed_templates.py   the nightly, Rust-free again
-m 'wire'          → 13  test_wire_roundtrip.py
-m 'not expensive' → 35  test_wire_roundtrip.py     pytest.yml, unchanged
```

`pytest -m wire tests/test_wire_roundtrip.py`: 13 passed. `pytest -m "not expensive"`: 1036 passed,
11 skipped. `pyright`: 0 errors.

Nothing in PR CI exercises the nightly's selector — `pytest.yml` never runs `-m fuzz` — so the first
real proof is the next scheduled run.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -41,11 +41,9 @@ jobs:
           java-version: "21"
 
       - name: sync-deps
-        # --no-dev skips the `dev` group and the `apps` group it includes, so this
-        # job never builds the `rust/` wheels and never installs the pinned
-        # toolchain. Anything here that shelled out to cargo would make rustup
-        # fetch that toolchain mid-test — the fuzz step below is scoped to avoid
-        # exactly that. Every `uv run` below needs the flag too: a bare `uv run`
+        # --no-dev skips the `dev` group (and the `apps` group it includes), so this
+        # job doesn't compile the Rust crates — nothing under `expensive` or `fuzz`
+        # touches them. Every `uv run` below needs the flag too: a bare `uv run`
         # re-syncs with the default groups and would pull `apps` back in.
         run: |
           uv sync --group test --extra prover --no-dev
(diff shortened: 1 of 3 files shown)


chandrakananandi (approved):
Thanks for fixing this!

chandrakananandi (approved):
Seems like a better fix. Thanks!

PR #182, opened 2026-08-19
Fix issues that made long runs fail or hang

chandrakananandi:
I was trying to run AutoProver (crucible on the solana_vault demo) via Fable and it had to do these fixes to get it working. I wanted to make a PR with the fixes, in case these might help others too.

From Claude:

1. LLM calls died at 10 min, or hung forever. Non-streaming calls hit the Anthropic SDK's 10-minute limit, and our `timeout=None` disables timeouts completely, so a dropped connection hung the run indefinitely. Fix: stream responses (no limit) and time out after 5 min of silence. This will affects every Anthropic call in the repo. Is that ok?

2. Doc search: broke under parallel agents. The embedding model isn't thread-safe: concurrent searches returned no results on CPU and segfaulted on Mac GPUs. Fix was to run one encode at a time; add COMPOSER_EMBED_DEVICE=cpu to skip the GPU.

3. Report: empty despite successful work. State read back from Postgres sometimes arrives as plain dicts instead of typed objects, crashing the final step after all the paid LLM work was done. Fix was to re-validate on read. (Why the round-trip does this, and why an all-failed run still exits 0, are left open.)

### DIFF
--- a/composer/diagnostics/usage_callback.py
+++ b/composer/diagnostics/usage_callback.py
@@ -20,7 +20,7 @@ from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.messages import AIMessage
 from langchain_core.outputs import ChatGeneration, LLMResult
 
-from graphcore.utils import TokenUsageDict, get_token_usage
+from graphcore.utils import TokenUsageDict, get_normalized_token_usage, get_token_usage
 from composer.diagnostics.timing import get_run_summary
 
 
@@ -29,29 +29,22 @@ def _usage_of(msg: AIMessage) -> TokenUsageDict:
 
     ``get_token_usage`` reads the raw Anthropic ``response_metadata["usage"]`` dict,
     which only a non-streamed response carries; a streamed response reports usage
-    solely through the provider-normalized ``usage_metadata``. Fall back to that,
-    translating back to the raw shape: normalized ``input_tokens`` is the total
-    including both cache buckets, where the raw count excludes them."""
+    solely through the provider-normalized ``usage_metadata``. Fall back to
+    ``get_normalized_token_usage`` over that, translating back to the raw shape:
+    normalized input is the total including both cache buckets, where the raw
+    count excludes them."""
     usage = get_token_usage(msg)
-    if any(
-        usage[k]
-        for k in (
-            "input_tokens", "output_tokens",
-            "cache_read_input_tokens", "cache_creation_input_tokens",
-        )
-    ):
+    if "usage" in msg.response_metadata:
         return usage
-    if (um := msg.usage_metadata) is None:
-        return usage
-    details = um.get("input_token_details") or {}
-    cache_read = details.get("cache_read", 0)
-    cache_creation = details.get("cache_creation", 0)
+    norm = get_normalized_token_usage(msg)
+    cache_read = norm["cache_read_tokens"]
+    cache_write = norm["cache_write_tokens"]
     return {
-        "model_name": usage["model_name"],
-        "input_tokens": max(0, um["input_tokens"] - cache_read - cache_creation),
-        "output_tokens": um["output_tokens"],
+        "model_name": usage["model_name"] or norm["model_name"],
+        "input_tokens": max(0, norm["total_input_tokens"] - cache_read - cache_write),
+        "output_tokens": norm["total_output_tokens"],
         "cache_read_input_tokens": cache_read,
-        "cache_creation_input_tokens": cache_creation,
+        "cache_creation_input_tokens": cache_write,
     }
 
 
--- a/composer/diagnostics/usage_callback.py
+++ b/composer/diagnostics/usage_callback.py
@@ -20,8 +20,39 @@ from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.messages import AIMessage
 from langchain_core.outputs import ChatGeneration, LLMResult
 
-from graphcore.utils import get_token_usage
+from graphcore.utils import TokenUsageDict, get_token_usage
 from composer.diagnostics.timing import get_run_summary
+
+
+def _usage_of(msg: AIMessage) -> TokenUsageDict:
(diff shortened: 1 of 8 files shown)


ericeil:
> Report: empty despite successful work. State read back from Postgres sometimes arrives as plain dicts instead of typed objects, crashing the final step after all the paid LLM work was done. Fix was to re-validate on read. (Why the round-trip does this, and why an all-failed run still exits 0, are left open.)

I ran into this myself today.  It looks like it's an issue with the new "tool family" thing.  I don't think this is the right fix though; I will open a separate PR for this.

composer/llm/anthropic.py
```diff
@@ -293,7 +293,16 @@ def builder_for(
...
+            # An explicit None DISABLES the SDK's timeouts (None != not-given), so a
+            # socket that dies silently mid-stream hangs the session forever. A float
+            # is a per-phase httpx timeout — for a streamed response, the max silence
+            # between chunks, not a cap on the whole turn.
+            timeout=300.0,
+            # Stream every request: a long authoring turn (Opus + thinking on a large
+            # prompt) can exceed the SDK's 600s non-streaming ceiling, and a silent
+            # 10-minute wait is long enough for NAT/idle killers to drop the socket
+            # (surfaces as APIConnectionError mid-run). Streaming keeps bytes flowing.
+            streaming=True,
```
jtoman, line 305: I happen to know this causes the langgraph api to produce streaming results in its own API. Please double check this doesn't utterly break the TUI and console display handlers, i.e. we aren't streaming chunks that our handlers have no idea what to do with, this somehow opts us out of the complete results, etc.

composer/rag/db.py
```diff
@@ -96,19 +97,32 @@ async def get_manual_section(self, headers: list[str]) -> str | None:
...
 
     async def embed_docs(
         self, doc: list[BlockChunk]
     ) -> list[ndarray]:
-        return cast(list[ndarray], await asyncio.to_thread(
-                self.tr.encode_document, [f"search_document: {d.chunk}" for d in doc], show_progress_bar=False
-            ))
+        return await asyncio.to_thread(
+                self._encode_docs_locked, [f"search_document: {d.chunk}" for d in doc]
+            )
```
jtoman, line 123: this should also be applied to the `DefaultEmbedder` found in the indexed research agents, although there the API is `sync` so we can't just offload to `to_thread`....

composer/rustapp/session.py
```diff
@@ -941,14 +941,28 @@ async def run_session[K: (RustFormalResult, RustSetupSpec)](
...
-        property_checks=[(m.property_title, m.checks) for m in state["property_checks"]],
-        verdicts=state["verdicts"],
-        ran=state["ran"],
+        skipped=_revive(SkippedProperty, state["skipped"]),
+        property_checks=[
+            (m.property_title, m.checks)
+            for m in _revive(PropertyCheckMapping, state["property_checks"])
+        ],
+        verdicts={k: _revive_one(WireVerdict, v) for k, v in state["verdicts"].items()},
+        ran=_revive(Target, state["ran"]),
```
jtoman: ugh, this has bitten us before. It's unclear to me why langgraph sometimes deserializes the type, sometimes hands us back the raw dict. Having to add this defensive coding is madness, but idk what to do ...

ericeil: Hm, I guess if this is a pre-existing condition then ignore my previous comment. :)  I'll look closer at the "tool family" problem.

ericeil: Ok, yeah, this was a pre-existing condition that also affected the new shared tools. :)  See https://github.com/Certora/graphcore/pull/35

ericeil: From the description of https://github.com/Certora/graphcore/pull/35:

> A checkpoint serializer (JsonPlusSerializer) names a pydantic value by cls.__module__ and
> cls.__name__, then restores it with getattr(import_module(module), name)(**kwargs). On failure
> it returns the kwargs dict.
> 
> create_model takes __module__ from the calling frame, so both the bound clone and a rendering
> of it claimed graphcore.tools.schemas. That name is not there, so a rendered value comes back as
> a bare dict and the next read of it fails on an attribute it no longer has:

chandrakananandi:
> > Report: empty despite successful work. State read back from Postgres sometimes arrives as plain dicts instead of typed objects, crashing the final step after all the paid LLM work was done. Fix was to re-validate on read. (Why the round-trip does this, and why an all-failed run still exits 0, are left open.)
> 
> I ran into this myself today. It looks like it's an issue with the new "tool family" thing. I don't think this is the right fix though; I will open a separate PR for this.

Should I undo this then?

composer/rustapp/session.py
chandrakananandi: so should we keep this change or no? Seems like it can help, no?

ericeil: I think we should revert this part of the change; my graphcore change should make this unnecessary.  BTW the AutoProver side of that change is here, if anyone wants to approve: https://github.com/Certora/AutoProver/pull/184

composer/diagnostics/usage_callback.py
```diff
@@ -20,10 +20,41 @@
...
+    details = um.get("input_token_details") or {}
+    cache_read = details.get("cache_read", 0)
+    cache_creation = details.get("cache_creation", 0)
+    return {
+        "model_name": usage["model_name"],
+        "input_tokens": max(0, um["input_tokens"] - cache_read - cache_creation),
+        "output_tokens": um["output_tokens"],
+        "cache_read_input_tokens": cache_read,
+        "cache_creation_input_tokens": cache_creation,
+    }
```
jtoman: `get_normalized_token_usage` exists tho?

chandrakananandi: addressed, thanks.

ericeil (approved):
Looks good!

jtoman (approved):

PR #186, opened 2026-08-22
autosetup: keep via-ir off contracts whose pinned compiler cannot take its settings

shellygr:
A fleet sweep of ~400 projects on current master turned up one regression against the previous
sweep, and this is it: two contracts in one project went from ok to a failed compilation analysis,
at roughly double the runtime.

### What happens

The scene mixes compilers through `compiler_map`. One contract is pinned to solc 0.8.4; the rest
are on 0.8.24. Under #180 the stack-too-deep ladder now (a) enables the optimizer globally on its
first rung and (b) takes via-ir scene-wide as soon as the build config declares `via_ir`, rather
than naming contracts one at a time. Neither step consults the per-contract pin.

With via-ir and the optimizer both on for a file, certoraRun emits
`settings.optimizer.details = {"inliner": false, "yulDetails": …}`. The `inliner` key exists from
solc 0.8.5, so the 0.8.4 binary rejects the input:

```
solc8.4 had an error:
Unknown key "inliner"
```

That fails the compile for the whole scene, not just that contract. No rung matches the text
(`unsupported_solc_via_ir` is gated on the scalar `solc_via_ir`, which is never set when via-ir
arrived through the map), so the loop falls to its catch-all, the import patcher runs as a last
resort, the compile fails identically, and the run ends as a compilation-analysis failure with the
patch reverted.

### The change

`_apply_via_ir_workaround` now checks each contract's pinned compiler before switching it, in both
the single-contract and the scene-wide paths, and logs which contracts it kept on legacy codegen.
Unpinned contracts run on the environment's solc and are unaffected, so the common case does not
change.

The floor is a new `VIA_IR_SETTINGS_MIN_VERSION = 0.8.5` next to `VIA_IR_MIN_VERSION`, which covers
something else: 0.7.5 is where solc learned `settings.viaIR`, 0.8.5 is where it learned the settings
we send with it.

### Tests

Two new cases in `tests/test_compilation_workarounds.py`, both failing without the change: a
scene-wide escalation that must hold back a below-floor contract, and a below-floor contract that
reported the stack-too-deep itself. A third asserts unpinned scenes still go via-ir. 77 pass.

### The other half

certora-cli emits the details block with no version guard when via-ir is on, while its own
non-via-ir branch guards exactly this at 0.8.5. Guarding both would make this floor redundant, and
would also cover callers that set via-ir without going through this ladder. That change is not in
this repo.

Draft: #180 is a clear net win in the same sweep (stack-too-deep failures went from 34 to 1), so
this is about paying its edge case, and I would like a second opinion on the floor's placement
before it lands.

### DIFF
--- a/certora_autosetup/utils/compilation_workarounds.py
+++ b/certora_autosetup/utils/compilation_workarounds.py
@@ -34,11 +34,14 @@ from certora_autosetup.utils.logger import logger
 from certora_autosetup.utils.paths import user_harness_path, user_harnesses_dir
 from certora_autosetup.utils.remappings import build_packages_from_remapping_sources
 from certora_autosetup.utils.solc_version_resolver import (
+    VIA_IR_SETTINGS_MIN_VERSION,
     extract_pragma_spec,
     pragma_admits,
     read_pragma_from_source_file,
     resolve_pragma_to_version,
 )
+from certora_autosetup.utils.config_manager import certora_format_to_raw_version
+from packaging.version import Version
 from certora_autosetup.utils.types import ContractHandle
 
 # Solc's legacy-codegen stack-too-deep, as opposed to the YulException the via-ir pipeline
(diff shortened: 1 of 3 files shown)


shellygr:
this will likely become obsolete in the next certora-cli version. keeping as draft for now

PR #187, opened 2026-08-22
autosetup: anchor on the project whose artifacts these actually are

shellygr:
Follow-up to the artifact-directory work in #169, from the same sweep that produced #186.

### What goes wrong

`find_build_config_dir` walks up from the contract and stops at the first directory that holds a
build config and has its artifact directory on disk. That test is satisfiable by a directory that
never built anything, because two configs can name the same physical directory:

```
foundry.toml            src = 'pkg/src'   out = 'pkg/out'     <- this one ran
pkg/foundry.toml        src = 'src'       out = 'out'         <- this one did not
```

`pkg/out` exists and `pkg/foundry.toml` is a config, so the walk stops at `pkg/`. Everything
downstream then reads those artifacts in the wrong frame: the extractor builds its project-file set
relative to `pkg/`, while the artifacts record `pkg/src/Widget.sol` relative to the root. No source
path matches, the scope filter drops every artifact, and the run dies with "No project-local logic
contracts found in FoundryManager build output" — from a build that had no errors at all.

In the sweep this cost one project 110 matching artifacts, and the class it lands in is the run's
largest.

### The change

The artifacts say which project wrote them: Foundry records `metadata.settings.compilationTarget`,
Hardhat records `sourceName`, and both are relative to the project the build ran in. A candidate
directory now has to pass that test — at least one recorded source path has to resolve inside it —
so a directory holding somebody else's artifacts is skipped and the walk continues to the one they
belong to.

This extends #169's own principle rather than reversing it: decide from what is on disk, not from
the directory layout. A genuinely nested project that built its own artifacts records paths relative
to itself, passes, and still anchors where #169 put it. Artifacts with no recorded source (older
Foundry, stripped metadata) answer yes, so nothing moves for them.

Sampling is capped at 20 artifacts per candidate, so the check costs a few JSON reads on a path that
already walks the tree.

### Tests

`tests/test_project_dir.py` gains three cases: the shared-out-dir layout above (fails without the
change), a nested project that really did write its own artifacts (must keep anchoring there), and
artifacts with no metadata (must not move). 24 pass in that file, 111 across the parser and
workaround suites.

The repo's full run fails 14 for me and 15 on a clean tree, all in integration tests that vary
between runs; none of them is unique to this change.

Draft, and worth reviewing alongside #186: both come out of the same sweep, and this one touches the
path #169 introduced.

### Review

`recorded_source` is an abstract method on `BuildSystemManager` next to the existing
`holds_artifacts`. Each manager names its own field, and the base does the sampling walk and the
ownership test once. Truffle is covered, which it was not before: it records `sourcePath`, so it
answered every candidate yes and the check never applied to it.

Truffle's `sourcePath` is absolute where the other two are project-relative, so the base tests an
absolute path for containment rather than joining it onto the candidate. Joining discards the
candidate under pathlib and would accept any artifact whose source still existed anywhere.

The extractors parse the same three fields for their own purposes (`parsers/foundry.py`,
`parsers/hardhat.py`, `parsers/truffle.py`) and are left alone. They do more than read the path,
and folding them in is a separate change.

### DIFF
--- a/certora_autosetup/utils/project_dir.py
+++ b/certora_autosetup/utils/project_dir.py
@@ -9,24 +9,20 @@ owns a contract — the remappings, pinned solc and artifact layout all come fro
 monorepos vendor a per-package ``foundry.toml`` under ``modules/`` or ``lib/`` while the
 root config is what actually builds the tree, so the nearest config is often not the one
 that ran. See ``find_build_config_dir``.
+
+Reading the artifacts is the build systems' own business, so this module locates the
+directory and asks the matching ``BuildSystemManager`` whether what is in it is theirs.
 """
 
-import json
 import os
 import tomllib
 from pathlib import Path
 from typing import Optional
 
-# Build config filenames that mark a directory as a project root, in no particular order —
-# presence of any one of them is enough to anchor there. Truffle has two: `truffle.js` is
-# the v4 spelling, `truffle-config.js` everything since.
-BUILD_CONFIG_FILENAMES = (
-    "foundry.toml",
-    "hardhat.config.js",
-    "hardhat.config.ts",
-    "truffle-config.js",
-    "truffle.js",
-)
+from certora_autosetup.build_systems.foundry import FoundryManager
+from certora_autosetup.build_systems.hardhat import HardhatManager
+from certora_autosetup.build_systems.manager import BuildSystemManager
+from certora_autosetup.build_systems.truffle import TruffleManager
 
 # Hardhat writes artifacts here unless the config sets `paths.artifacts`, and Truffle
 # unless it sets `contracts_build_directory`. Reading either override means running node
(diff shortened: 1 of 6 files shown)


certora_autosetup/utils/project_dir.py
```diff
@@ -85,11 +86,56 @@ def _artifact_dir_of(config_dir: Path) -> Optional[Path]:
     )
 
 
+def _declared_sources(artifacts: Path, limit: int = 20) -> list[str]:
+    """Source paths a sample of *artifacts* say they were compiled from.
+
+    Foundry records them as the keys of ``metadata.settings.compilationTarget``; Hardhat as
```
jar-ben: 1. We support more build systems than just hardhat and foundry, so perhaps we should handle all of them also here. 
2. it would be better to have this as an abstract method implemented by the particular build managers

jar-ben (changes requested):

certora_autosetup/utils/project_dir.py
shellygr: Claude answers: Done, both points.

jar-ben (approved):

PR #188, opened 2026-08-24
Fix imports whose letter case does not match the file on disk

shellygr:
## What this fixes

A Solidity project written on macOS or Windows can import a file by a path that does not match
the real filename, usually differing only in letter case. Those filesystems are case-insensitive,
so the project compiles for its authors and cannot compile anywhere case-sensitive. Compilation
analysis used to stop there, with a source-not-found error and nothing to try next.

## Approach

New module `certora_autosetup/setup/import_case_fix.py`. It checks each component of an import
path against the real directory entries with `os.scandir` instead of asking whether the path
exists. An existence check on a case-insensitive filesystem resolves the mis-cased path happily,
which would make the defect invisible on the machine where the code is written and leave the fixer
firing only in production.

Handled: letter case in the basename, in any directory component, and in the extension, plus
Windows separators and doubled slashes. A path that names no file even case-insensitively is left
alone, because that is an absent file rather than a case problem.

Three bails are worth naming, since none of them is visible from the path alone:

- An import that resolves through a remapping or package prefix. Where it leads is the conf's
  business, not the project tree's, and repairing it is the packages workaround's job.
- A file inside a dependency checkout. Its sources are read, since an import may legitimately
  point into one, but never rewritten: a dependency install overwrites them.
- Two directory entries matching the same component case-insensitively. That is legal on a
  case-sensitive filesystem, so an ambiguous match rewrites nothing.

## Integration

In `run_compilation_analysis` the case fix gets a retry of its own. If it rewrote anything, the
compilation runs again, and on success the import patcher never runs at all. Only if that retry
still fails does the patcher go on top of the case fix rather than instead of it: the patcher
resolves each relative import against its map of real files and skips the ones that miss, so a
corrected spelling is one more import it can canonicalize. If the last attempt fails, both revert
in reverse order of application.

Rewrites are scoped to the quoted path literal, checked against the recorded bytes before anything
is written, kept single-line so the patcher's line-indexed revert stays correct, never applied
inside a dependency tree, and logged at WARNING with the old and new spelling.

Also adds a small `package_prefixes()` helper to `utils/import_diagnostics.py` so the fixer can
leave remapped prefixes alone.

## Testing

28 unit tests. They assert on planned and applied rewrites built from one real file plus a
wrong-cased import string, rather than creating two files that differ only in case. The latter is
impossible on a case-insensitive filesystem and would make the suite behave differently per
developer machine.

Full suite: 1125 passed, 10 skipped, 11 deselected, with 21 errors from the testcontainers-backed
`test_rag_db` and `test_indexed_tool` cases that need a database this machine does not start.
pyright: 0 errors.

## Known limitation

`extract_imports_multiline` accumulates lines until the next `;`, so a commented-out `import` with
no semicolon can make the scan land on an ordinary string constant further down the file and
rewrite it. Comment masking closes the common shape of this, but the accumulation is still there.
Closing it properly wants a real parser; the typed solc AST is not an option at this point because
it only exists after a successful build, and here the build has failed.

### DIFF
--- a/certora_autosetup/setup/import_spelling_fix.py
+++ b/certora_autosetup/setup/import_spelling_fix.py
@@ -1,0 +1,552 @@
+"""Rewrite imports whose spelling does not match the file that exists on disk.
+
+A project authored on a case-insensitive filesystem (macOS, Windows) can carry an import
+whose path differs from the real on-disk path — most often in letter case — and compile for
+its authors while failing everywhere else. This module compares every component of an
+import path against the actual directory entries, and when the filesystem names exactly one
+thing the import was meant to name, rewrites the quoted path literal to that spelling.
+
+Detection reads directory entries (``os.scandir``) instead of asking whether a path exists.
+On a case-insensitive filesystem an existence check resolves the mis-spelled path, so a
+fixer built on ``Path.exists`` would report nothing where it is developed and fire only in
+production.
+
+In scope: the letter case of the basename, of any directory component, and of the extension;
+Windows separators and repeated slashes; and the right basename in the wrong directory, when
+exactly one file in the project carries that basename.
+
+Out of scope, each an explicit bail rather than a silent miss:
+
+- Fuzzy or edit-distance matching. A near-miss name is a different contract, and pointing an
+  import at it yields a project that compiles while verifying code nobody asked about.
+- Guessing at name shape: CamelCase versus snake_case, singular versus plural.
+- Inventing or changing an extension. Only its letter case may change, which follows from
+  matching whole basenames case-insensitively.
+- Renaming files on disk.
+- Imports that resolve through a remapping or package prefix. Repairing those is the
+  ``source_not_found_packages`` workaround's job, which rebuilds the conf's packages list.
+- Imports naming a path inside a dependency checkout that does not resolve. That is a
+  dependency that is absent or laid out differently, and the project's own same-named mock or
+  interface is not what the import asked for.
+- More than one candidate. Two files differing only in case are legal on a case-sensitive
+  filesystem, so an ambiguous match rewrites nothing and says why.
+"""
+
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Callable, Dict, List, Optional, Sequence, Tuple
+
+from certora_autosetup.setup.solidity_import_patch import extract_imports_multiline
+from certora_autosetup.setup.solidity_utils import DEPENDENCIES, walk_files_by_suffix
+from certora_autosetup.utils.import_diagnostics import package_prefixes, parse_unresolved_imports
+from certora_autosetup.utils.logger import logger as _logger
+
+LogFunc = Callable[..., None]
+
+SOLIDITY_SUFFIX = ".sol"
+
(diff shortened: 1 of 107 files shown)


certora_autosetup/setup/import_spelling_fix.py
```diff
@@ -0,0 +1,620 @@
...
+
+In scope: the letter case of the basename, of any directory component, and of the extension;
+Windows separators and repeated slashes; and the right basename in the wrong directory, when
+exactly one file in the project carries that basename.
+
+Out of scope, each an explicit bail rather than a silent miss:
+
+- Fuzzy or edit-distance matching. A near-miss name is a different contract, and pointing an
+  import at it yields a project that compiles while verifying code nobody asked about.
+- Guessing at name shape: CamelCase versus snake_case, singular versus plural.
```
jar-ben: Why do you even include these "out of scope" here? They are obviously out of scope. Or are you saying that mixing CamelCase and snake_case works just fine on Windows/macOS?

jar-ben: Perhaps we can rename the class/file to something obviously saying os-specifing/agnostic spelling issues

certora_autosetup/setup/setup_prover.py
```diff
@@ -451,34 +460,49 @@ def run_compilation_analysis(
...
+                # The spelling fix has to run before the import patcher, because the patcher
+                # resolves each relative import against its map of real files and skips the
+                # ones that miss (solidity_import_patch.create_patch). A misspelled import
+                # never resolves, so it is exactly what the patcher cannot canonicalize.
+                # Correct spellings first, canonical paths second.
+                spelling_fix_applied = self._run_import_spelling_fix(
+                    output, compilation_config.get("packages") or []
+                )
+                import_patch_applied = self._run_import_patch()
+                import_patcher_applied = import_patch_applied
```
jar-ben: Don't run both together? what if spelling fix helps (applies), but import pach fails (and let's suppose we need only the spelling fix).

certora_autosetup/setup/import_spelling_fix.py
shellygr: Claude answers: Fair. They are not case problems at all, and listing them only made the module look like a general spelling fixer. Dropped. What is left is the three bails you cannot see from the path alone: a remapped prefix, a dependency checkout, and two entries matching one component case-insensitively.

shellygr: Claude answers: Done. It is import_case_fix.py now, with ImportCaseRewrite and plan/apply/revert_import_case_fixes. I also dropped the right-basename-in-the-wrong-directory branch, which was the one part that was not an OS issue, so the name now covers everything the module does. No observed run ever needed it; the failure behind this PR is a plain basename case mismatch.

certora_autosetup/setup/setup_prover.py
shellygr: Claude answers: Done. The case fix gets its own retry, and the patcher only runs if that retry still fails, on top of the case fix rather than instead of it. That costs a third compile only when the case fix applied and was not enough.

jar-ben (approved):

PR #190, opened 2026-08-25
support openrouter as an llm provider

naftali-g:
It matches any model with the `author/model` naming scheme.


composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+little carries over: vendor-qualified ids, no Files API, a different request shape.
+
+Three things differ from the OpenAI backend:
+
+* **Probing is live, not parsed.** OpenRouter fronts 400+ models whose names encode
+  nothing and whose roster turns over weekly, so ``GET /api/v1/models`` supplies the
+  window, output cap, reasoning support and price card instead of a name parser and
+  a hand-maintained table. See :func:`_catalog`.
+* **Files are inlined.** There is no Files API — see :class:`InlineFileUploader`.
+* **Every route goes through the Responses API**, not Chat Completions — see
```
jtoman: > "not chat completions"

what is bro talking about, the openai provider uses the responses API????

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+
+_PROBE_TIMEOUT_SECONDS = 15
+
+# The only fields read off a roster record.
+_PROBED_KEYS = ("context_length", "top_provider", "supported_parameters", "pricing")
+
+# Transient stream failures and how many times to re-issue the request. The SDK's
+# own `max_retries` covers *establishing* a request; these surface while the SSE
+# body is being consumed, by which point the request has already succeeded, so
+# nothing below us retries them and they take the whole run down. `TimeoutError`
```
jtoman: not true anymore, fwiw, we have graph level retry mechanisms....

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+    # disables thinking silently.
+    reasoning: bool
+
+
+@cache
+def _catalog() -> dict[str, Mapping[str, Any]]:
+    """OpenRouter's model roster, by id: one blocking unauthenticated GET, no retry,
+    from :meth:`OpenRouterModelProvider.create` at startup. Any failure yields an
+    empty catalog and a run on conservative defaults."""
+    try:
```
jtoman, line 151: why not async?

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+        with urllib.request.urlopen(_MODELS_URL, timeout=_PROBE_TIMEOUT_SECONDS) as resp:
+            payload = json.load(resp)
+    except (OSError, json.JSONDecodeError) as exc:
+        logger.warning(
+            "Could not fetch OpenRouter model metadata from %s (%s); falling back to "
+            "a %d-token context window and no price card.",
+            _MODELS_URL, exc, _FALLBACK_CONTEXT_WINDOW,
+        )
+        return {}
+    records = payload.get("data") if isinstance(payload, dict) else None
```
jtoman: assuming this json is returned according to a documented schema, just use pydantic for heaven's sake rather than this "hand" rolled parsing...

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+    if (exact := catalog.get(model_name)) is not None:
+        return exact
+    # Variant suffixes (`:free`, `:nitro`, `:floor`) select a routing policy, not a
+    # different model; most are absent from the roster under their suffixed id.
+    base, _, variant = model_name.partition(":")
+    return catalog.get(base) if variant else None
+
+
+def _positive_int(value: Any) -> int | None:
+    return value if isinstance(value, int) and not isinstance(value, bool) and value > 0 else None
```
jtoman: what the blue hell? again, all of this can be declaratively described using pydantic I think ...

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+        ),
+        max_output_tokens=_positive_int(top.get("max_completion_tokens")),
+        reasoning="reasoning" in supported,
+    )
+
+
+# --- pricing ---------------------------------------------------------------
+
+# OpenRouter quotes USD per token; PriceTier is USD per million.
+_TOKENS_PER_MTOK = 1_000_000
```
jtoman: "tokens per million tokens == 1 million"

word?

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+        return None
+    # A route that publishes no cache bucket bills those tokens as fresh input, and
+    # one with no separate 1h rate charges the 5m one. `or` would not do here: a
+    # real 0.0 is a free route.
+    if (cache_read := _per_mtok(prices.get("input_cache_read"))) is None:
+        cache_read = prompt
+    if (cache_write := _per_mtok(prices.get("input_cache_write"))) is None:
+        cache_write = prompt
+    if (cache_write_1h := _per_mtok(prices.get("input_cache_write_1h"))) is None:
+        cache_write_1h = cache_write
```
jtoman: so wait, if we don't see these numbers, we just make them up? that seems... wrong?

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+                tiers.append((floor, tier))
+    tiers.sort(key=lambda t: t[0], reverse=True)
+
+    def provider(input_tokens: int) -> PriceTier | None:
+        for floor, tier in tiers:
+            if input_tokens > floor:
+                return tier
+        return short
+
+    return provider
```
jtoman: something something pydantic, almost certainly assuming this data conforms to any kind of advertised schema

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+    renderer: ContentRenderer = field(default_factory=OpenRouterRenderer)
+
+    async def _upload_bytes(
+        self, crc_basename: str, file_data: bytes, mime: str
+    ) -> str:
+        # No dedup cache: the "id" *is* the content, so there is nothing to reuse.
+        # Off-thread because the encode is ~1.5ms/MB of blocked loop, matching how
+        # `composer.input.files` already offloads the read.
+        encoded = await asyncio.to_thread(base64.b64encode, file_data)
+        return f"data:{mime};base64,{encoded.decode('ascii')}"
```
jtoman: oh. I hate it. yes it works, but its hijacking the fact that the type we're using to represent a file id (`str`) happens to *also* allow us to smuggle through the Base64 encoding that is necessary for an "inline" block. Far better to make the file upload optional rather than lying about it.

Luckily, I think the fix is pretty easy; make an analog of `InMemoryTextFile` called `InMemoryBytesFile` which invokes a new API on `FileRenderer`: `inline_file_block` or something. then the "regular" `file_block` becomes a `NotImplemented` error thrown by the openrouter content renderer.

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+
+class OpenRouterService(ProviderServiceBase):
+    """Provider services for OpenRouter. ``cache_marker`` stays the base no-op:
+    OpenRouter caches prompts itself on the Responses API, so an explicit
+    ``cache_control`` breakpoint would add nothing."""
+
+    def __init__(self):
+        from graphcore.tools.memory import openai_async_memory_tool
+        super().__init__(
+            # The OpenAI-flavored memory tool is a plain client-side function tool
```
jtoman: FFS. Its so weird seeing claude get this wrong. The anthropic memory tool is *also* client side. It is *convinced* that the anthropic memory tool we use is server side, it isn't!!!!! Stop implying it is!

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+                        "re-issuing in %.0fs (attempt %d/%d).",
+                        len(chunks), type(exc).__name__, exc, delay,
+                        attempt + 1, _STREAM_ATTEMPTS,
+                    )
+                    await asyncio.sleep(delay)
+                    continue
+                for chunk in chunks:
+                    yield chunk
+                return
+
```
jtoman: Is this mooted by the retry mechanisms I recently landed? Or is it a middle level of retrying (in which case we are probably retrying at three different levels...)

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+    model_name: str
+    options: ModelConfiguration
+    features: OpenRouterModelFeatures
+    price_provider: PriceProvider
+    api_key: str
+    provider: OpenRouterService = field(default_factory=_openrouter_service)
+
+    @staticmethod
+    def create(model_name: str, options: ModelConfiguration) -> "OpenRouterModelProvider":
```
jtoman, line 498: oh, because this isn't async. hum.

composer/llm/openrouter.py
```diff
@@ -0,0 +1,464 @@
...
+            }
+            # Ask for the encrypted chain of thought, which is what langchain echoes
+            # back with the next tool result so the model can resume it.
+            kwargs["include"] = ["reasoning.encrypted_content"]
+
+        return _chat_model_cls()(
+            model=self.model_name,
+            base_url=BASE_URL,
+            api_key=SecretStr(self.api_key),
+            # Load-bearing, not a default: the Responses API is the only surface on
```
jtoman: take a drink

jtoman (changes requested):

composer/llm/openai.py
```diff
@@ -162,16 +162,23 @@ def text_block(self, text: str, *, cache_level: CacheLevel = CacheLevel.NONE) ->
...
             },
         }
 
+    def inline_file_block(
+        self, basename: str, contents: bytes, mime: str,
+        *, cache_level: CacheLevel = CacheLevel.NONE
+    ) -> dict:
+        raise NotImplementedError(
+            "OpenAI content is uploaded to the Files API, not inlined."
+        )
```
jtoman, line 180: technically we could allow inline blocks too, they are supported just fine on anthropic/openai. But I don't think we do; its usually a disaster for our checkpointing...

composer/llm/openrouter.py
```diff
@@ -97,35 +91,98 @@ class OpenRouterModelFeatures:
...
+    input_cache_write_1h: float | None = None
+
+
+class _PriceOverride(_PriceCard):
+    """A conditional price, restating only the fields it changes. Only prompt-size
+    floors are modelled; OpenRouter also publishes time-of-day windows
+    (``utc_start``/``utc_end``), which a per-call price curve cannot express, so
+    those arrive with ``min_prompt_tokens`` unset and are skipped."""
+
+    min_prompt_tokens: PositiveInt | None = None
```
jtoman, line 118: oh, and positiveint is even built in to pydantic. nice!

composer/llm/openrouter.py
```diff
@@ -97,35 +91,98 @@ class OpenRouterModelFeatures:
...
+
+
+class _Pricing(_PriceCard):
+    overrides: list[_PriceOverride] = Field(default_factory=list)
+
+
+class _ModelRecord(BaseModel):
+    id: str
+    context_length: PositiveInt | None = None
+    top_provider: _TopProvider = Field(default_factory=_TopProvider)
```
jtoman: It doesn't matter, but I'm trying to figure out what "Top" means here. Like, the top provider of a specific model?

jtoman (approved):

composer/llm/openrouter.py
naftali-g: openrouteer uses several providers for each model (e.g. kimi-k3 has 17 different possible providers) and it uses some sort of weighted load balancing to choose between them for each call. the top provider is the one openrouter thinks is best. but looking at this again, this actually exposed an issue - some providers may have a smaller context lengths, which would have caused issues if it would try to route to them. So I'm changing this to instead take only providers that provide a long enough context.

PR #199, opened 2026-08-30
Recover from CVL type-check failures, and fix a deadlock in the same path

naftali-g:
`declared_rules_list` (added in #177) runs `certoraRun --compilation_steps_only` before every prover call. Two defects in it:

It killed the run on a repairable error. The subprocess output was discarded and a bare `ValueError("Type check failed?")` raised — a type not in graph.py's handle_tool_errors, so langgraph re-raised it. That regressed behavior `run_prover` still has fifteen lines below: return the compiler's complaints as the tool result so the author agent can repair the spec.

It could hang forever. Both streams were piped with nobody draining them, so a child outfilling the ~64KB buffer blocked on write while we blocked on `wait()`.

### DIFF
--- a/composer/prover/core.py
+++ b/composer/prover/core.py
@@ -449,6 +449,40 @@ async def run_prover_inner(
         run_result = cast(ProverResult, json.load(output_file))
         return run_result, stdout
 
+# How much of a failed compile to hand back. Trimmed from the end because the errors come last.
+_COMPILE_OUTPUT_TAIL_CHARS = 8_000
+
+
+class SpecCompilationError(Exception):
+    """The spec did not compile.
+
+    Carries the compiler's own output, which names the offending lines. An authoring
+    agent can repair a spec from that, so callers driving one should surface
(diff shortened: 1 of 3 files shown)


composer/prover/core.py
```diff
@@ -449,6 +449,40 @@ async def run_prover_inner(
...
+        *argv,
+        cwd=str(cwd),
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+    stdout, stderr = await proc.communicate()
+    output = b"".join(part for part in (stdout, stderr) if part).decode(
+        "utf-8", errors="replace"
+    )
+    return proc.returncode or 0, output[-_COMPILE_OUTPUT_TAIL_CHARS:]
```
jtoman: just return the whole thing! WHO CARES! WHY TF DOES CLAUDE KEEP TRUNCATING STRINGS

jtoman (approved):

PR #200, opened 2026-08-31
Add a prioritized run mode

shellygr:
## Why

Today there is nono guaranteed correlation between the effort put by the agent and the value of the output. Property inference fans out per
component, every property it invents gets the full CVL authoring + prover + counterexample
treatment, and `budget_integration.md` already names `formalization` the dominant cost. That phase
scales with (components x properties), so an inference agent that thinks of forty properties bills
for forty.

## What

Today's behaviour becomes run mode `comprehensive` and stays the default. `prioritized` ranks every
candidate property across every component once, then spends the run on the highest-contribution
property plus the properties needed to state or discharge it. One property, pursued properly,
instead of everything pursued shallowly.

Concretely, a prioritized batch is the primary property plus **at most two** dependencies drawn from
its own component. All three are first-class: the feedback judge demands rules for each, they count
in the report, and all of them are skip-protected. The model does not pick the primary. It scores
every candidate and records what each rests on; `priority = score + 15 if critical_match` is applied
in code, so `property_ranking.json` and the property the run actually spends itself on are the same
thing by construction.

The insertion point is one seam in `composer/pipeline/core.py`, between `_extract_all` and
`staged.begin`. That is the first and only point where every component's candidates exist together:
inference never sees more than the unit it was given, and the per-component plugin hook cannot
compare across units either.

## How it is turned on

`--run-mode prioritized`, or `AUTOPROVER_RUN_MODE=prioritized`.

The value rides on `PipelineRun`, which the driver and the CVL author both already hold, so
`run_pipeline` and the backend protocols needed no new parameter.

## The ranking is not trusted

A single structured call, cached under the properties namespace, validated in Python before anything
acts on it: every candidate ranked exactly once, no invented keys, and every `depends_on` naming a
property of its own component. An unusable ranking raises rather than falling back to the whole
candidate set, because a silent widening would spend exactly what this mode exists to save. The
failure is cheap anyway, since nothing has been formalized yet.

## The author may not walk away

Two escape hatches would each be a total loss when the run is one property, so both tighten:
`record_skip` refuses the focus properties, and `give_up` rejects an `exhausted` stop until the
prover has actually been run a few times, with the attempts enumerated.

Two escapes are deliberate, and both are the difference between a gate and a deadlock:

- `sort='environment'` is never gated. A missing `certoraRun` or `solc` produces no prover run at
  all, so a bare attempt floor would be unsatisfiable and would hold the agent against its recursion
  limit instead of letting it report a broken toolchain.
- The budget monitor's wrap-up lifts the skip protection. Its order is "delete what does not work,
  skip the rest, publish", which is exactly what the protection forbids.

The floor counts `verify_spec` tool calls rather than `prover_history`. `prover_history` only grows
when the prover returns a report, and `verify_spec` returns early with a bare string when the spec
will not type-check, which is the failure most worth escalating through.

## The output says what it did not look at

A prioritized `report.json` would otherwise be indistinguishable from a contract with three
properties. So `report.json` carries `run_mode` and the deprioritized properties with their scores
and rationales, `coverage` carries a `deprioritized_count` and a warning, the full ranking is written
to `property_ranking.json`, and `job_info.json` records the mode on every path including a crash.

## What does not get cheaper

`prepare_formalization` is untouched, so the harness lift,
AutoSetup, custom summaries, structural invariants **and the invariant CVL/prover loop** cost exactly
what they cost today, as does per-component property inference (it has to run before anything can be
ranked). The saving is the formalization fan-out, which is the dominant cost. Two more caveats: in
`--cloud` mode components already verify in parallel, so the wall-clock win is smaller than the money
win, and a `GaveUp` is never cached, so a prioritized run against a genuinely unprovable property is
the worst case for this design.

## Scope

Prover flow only. The foundry and rust-app entry points keep their own parsers and do not expose the
flag, because the focus protections live in the CVL author: a prioritized foundry run would get the
narrowing without the discipline, which is the worst of both.

## Also in here

`FINAL_PROPERTIES_KEY` now forks on the mode. Its leaf was fixed apart from the threat-model,
interactive and extra-context parameters, so a prioritized run would have overwritten a
comprehensive run's record for the same component with its one-property batch, and an offline walker
would have read the pruned list as if it were everything inference found.

## Testing

`tests/test_prioritize.py` covers the validation and the cut, including that a supporting property
from another component is rejected and that the selection can never return an empty batch list (the
driver raises on that). `tests/test_focus_exits.py` covers the gate and the skip protection on
**both** binding paths, since the author reaches its skip tools either through `skip_tools` directly
or through `property_tools`, and a protection on only one is silently bypassable. Driver-level tests
in `tests/test_pipeline_staged_formalizer.py` prove one batch reaches formalization, that `begin`
still runs once beforehand, and that the ranker never runs in comprehensive mode.

`pytest -m "not expensive"`: 1126 passed. `pyright`: 0 errors. The 21 errors in the suite are a local
`pgvector` testcontainer that will not stay up; they are in `test_rag_db` / `test_indexed_tool` and
predate this branch.

## Not validated here

The ranker's judgement. The mechanism is tested; whether it picks the right property is what the
first real runs are for, and `property_ranking.json` is the artifact to read when checking that.

### DIFF
--- a/composer/authoring/tools.py
+++ b/composer/authoring/tools.py
@@ -11,16 +11,18 @@ What varies on skip and give-up is LLM-facing text, so those are
 instantiate time. Unskip does not vary.
 """
 
-from typing import Callable, Sequence, override
+from dataclasses import dataclass
+from typing import Callable, Literal, Sequence, override
 
-from langchain_core.messages import ToolMessage
+from langchain_core.messages import AIMessage, AnyMessage, ToolMessage
 from langchain_core.tools import BaseTool
 from langgraph.types import Command
 from pydantic import Field
 
 from graphcore.graph import tool_state_update
 from graphcore.tools.schemas import (
-    ToolFamilyParams, WithAsyncDependencies, WithImplementation, WithInjectedId, tool_family,
+    ToolFamilyParams, WithAsyncDependencies, WithImplementation, WithInjectedId,
+    WithInjectedState, tool_family,
 )
 
 from composer.authoring.state import SkippedProperty
@@ -38,6 +40,19 @@ def _as_titles(titles: Titles | Sequence[PropertyTitle]) -> Titles:
         return titles
     snapshot = titles
     return lambda: snapshot
+
+
+@dataclass(frozen=True)
+class SkipScope:
+    """What ``record_skip`` is allowed to retire: the batch's titles, and the subset it must
+    refuse.
+
+    ``protected`` is a thunk, not a set, because the answer changes during a session — the
+    budget monitor's wrap-up order tells the agent to skip everything that does not work, so a
+    protection that outlived that order would deadlock the session it was meant to curtail.
+    A backend with nothing to protect passes a thunk returning ``()``."""
+    titles: Titles
+    protected: Titles
 
 
 # ---------------------------------------------------------------------------
@@ -61,7 +76,7 @@ def _skip_result(
 
 @tool_family_display(_skip_label, _skip_result)
 @tool_family(SkipParams)
-class RecordSkip(WithInjectedId, WithAsyncDependencies[Command, Titles]):
+class RecordSkip(WithInjectedId, WithAsyncDependencies[Command, SkipScope]):
     """{description}"""
     property_title: PropertyTitle = Field(
         description="The snake_case title of the property from the batch listing"
@@ -70,13 +85,21 @@ class RecordSkip(WithInjectedId, WithAsyncDependencies[Command, Titles]):
 
     @override
     async def run(self) -> Command:
-        with self.tool_deps() as titles:
-            known = titles()
+        with self.tool_deps() as scope:
+            known = scope.titles()
+            protected = scope.protected()
         if self.property_title not in known:
             return tool_state_update(
                 self.tool_call_id,
                 f"Unknown property title {self.property_title!r}. Must be one "
                 f"of: {', '.join(known)}.",
+            )
+        if self.property_title in protected:
+            return tool_state_update(
+                self.tool_call_id,
+                f"Property {self.property_title!r} is the focus of this run and cannot be "
+                "skipped. Decompose it into lemmas, strengthen the harness, or formalize a "
+                "weaker core of it and say so in your commentary — but do not retire it.",
             )
         if not self.reason.strip():
             return tool_state_update(
@@ -122,12 +145,18 @@ def skip_tools(
     *,
     skip_description: str,
     skip_reason: str,
+    protected: Titles | Sequence[PropertyTitle] = (),
 ) -> list[BaseTool]:
-    """The skip / unskip pair, bound to the batch's property titles."""
+    """The skip / unskip pair, bound to the batch's property titles.
+
+    ``protected`` names titles ``record_skip`` must refuse — the focus of a prioritized run,
+    whose whole point is that it is pursued rather than retired. Unskip is deliberately not
+    constrained: un-retiring a property is always allowed."""
     get = _as_titles(titles)
+    scope = SkipScope(titles=get, protected=_as_titles(protected))
     return [
         RecordSkip.with_template(description=skip_description, reason=skip_reason)
-        .bind(get)
+        .bind(scope)
         .as_tool("record_skip"),
         Unskip.bind(get).as_tool("unskip_property"),
     ]
@@ -180,3 +209,92 @@ def give_up_tool(
         reason=reason_description,
         label=label,
     ).as_tool(name)
+
+
+# ---------------------------------------------------------------------------
+# Give up, behind an escalation gate
+# ---------------------------------------------------------------------------
+
+def _gated_give_up_label(p: dict, *, description: str, reason: str, label: str) -> str:
+    return f"Giving up on {label}: {p['reason']}"
+
+
+def _verify_attempts(state: dict) -> int:
+    """How many times this session has put a spec in front of the prover.
+
+    Counted from the message history rather than from ``prover_history``: that list only grows
+    when the prover returns a *report*, and the failures worth escalating through — a spec that
+    will not type-check, a toolchain that will not run — return early without one. Counting
+    calls counts attempts, which is what the gate is about. The same walk over ``messages`` is
+    how the prover tool identifies its own prior calls."""
+    return sum(
+        1
+        for msg in state.get("messages", [])
+        if isinstance(msg, AIMessage)
+        for call in msg.tool_calls
+        if call["name"] == "verify_spec"
+    )
+
+
+@tool_family_display(_gated_give_up_label, None)
+@tool_family(GiveUpParams)
+class GatedGiveUp(
+    WithInjectedId,
+    WithInjectedState[dict],
+    WithAsyncDependencies[Command, int],
+):
+    """{description}"""
+    sort: Literal["environment", "exhausted"] = Field(
+        description="Why you are stopping. 'environment' means the toolchain itself is unusable "
+        "— the prover or compiler is missing or broken — and nothing you could write would "
+        "succeed. 'exhausted' means you have tried to verify this and cannot find a way."
+    )
+    reason: str = Field(description="{reason}")
+    attempts: list[str] = Field(
+        min_length=1,
+        description="One entry per distinct approach you actually tried, saying what you "
+        "attempted and how it failed.",
+    )
+
+    @override
+    async def run(self) -> Command:
+        with self.tool_deps() as floor:
+            if self.sort == "exhausted" and (ran := _verify_attempts(self.state)) < floor:
+                return tool_state_update(
+                    self.tool_call_id,
+                    f"Rejected: you have run the prover {ran} time(s) on this task, and this run "
+                    f"requires at least {floor} before a property may be abandoned. This is the "
+                    "one property the run exists to establish. Decompose it into lemmas, add or "
+                    "strengthen a harness, or formalize a weaker core of it and state the gap — "
+                    "then verify. If the toolchain itself is broken, stop with "
+                    "sort='environment' instead.",
+                )
+        return tool_state_update(
+            self.tool_call_id,
+            "Accepted",
+            failed=True,
+            result=self.reason,
+        )
+
+
+def gated_give_up_tool(
+    *,
+    name: str,
+    description: str,
+    label: str,
+    min_attempts: int,
+    reason_description: str = "The reason for giving up on your task",
+) -> BaseTool:
+    """:func:`give_up_tool` with a floor under it: an ``exhausted`` surrender is refused until
+    the session has actually put a spec in front of the prover ``min_attempts`` times, and the
+    reason must enumerate what was tried.
+
+    For a run that has staked itself on one property, where an early surrender is the whole run.
+    The ``environment`` escape is not optional: a missing prover or compiler produces no prover
+    run at all, so a bare attempt floor would be unsatisfiable and would hold the agent against
+    its recursion limit instead of letting it report a broken toolchain."""
+    return GatedGiveUp.with_template(
+        description=description,
+        reason=reason_description,
+        label=label,
+    ).bind(min_attempts).as_tool(name)
--- a/composer/authoring/tools.py
+++ b/composer/authoring/tools.py
@@ -14,6 +14,8 @@ instantiate time. Unskip does not vary.
 from dataclasses import dataclass
 from typing import Callable, Literal, NotRequired, Sequence, override
 
+from typing_extensions import ReadOnly
+
 from langchain_core.messages import AIMessage, AnyMessage, ToolMessage
 from langchain_core.tools import BaseTool
 from langgraph.graph import MessagesState
@@ -233,10 +235,15 @@ def _gated_give_up_label(p: dict, *, description: str, reason: str, label: str) 
 
 
 class GatedGiveUpState(MessagesState):
-    """The authoring state :class:`GatedGiveUp` reads and writes: the message history it counts
-    prover attempts from, and the two fields a surrender records."""
-    result: NotRequired[str]
-    failed: NotRequired[bool]
+    """What a session must carry for :class:`GatedGiveUp` to run on it: the message history the
+    gate counts prover attempts from, and the two fields a surrender records.
+
+    ``result`` and ``failed`` are ``ReadOnly`` because the tool never writes them through the
+    state object; it returns a ``Command`` and the graph applies it. Declaring them anyway is
+    what makes the dependency visible, so a session missing them fails to typecheck rather than
+    at runtime."""
+    result: ReadOnly[NotRequired[str]]
+    failed: ReadOnly[bool | None]
 
 
 def _verify_attempts(state: GatedGiveUpState) -> int:
@@ -258,9 +265,9 @@ def _verify_attempts(state: GatedGiveUpState) -> int:
 
 @tool_family_display(_gated_give_up_label, None)
 @tool_family(GiveUpParams)
-class GatedGiveUp(
+class GatedGiveUp[S: GatedGiveUpState](
     WithInjectedId,
-    WithInjectedState[GatedGiveUpState],
+    WithInjectedState[S],
     WithAsyncDependencies[str | Command, int],
 ):
     """{description}"""
@@ -297,12 +304,13 @@ class GatedGiveUp(
         )
 
 
-def gated_give_up_tool(
+def gated_give_up_tool[S: GatedGiveUpState](
     *,
     name: str,
     description: str,
     label: str,
     min_attempts: int,
+    state_ty: type[S],
     reason_description: str = "The reason for giving up on your task",
 ) -> BaseTool:
     """:func:`give_up_tool` with a floor under it: an ``exhausted`` surrender is refused until
@@ -313,8 +321,12 @@ def gated_give_up_tool(
     The ``environment`` escape is not optional: a missing prover or compiler produces no prover
     run at all, so a bare attempt floor would be unsatisfiable and would hold the agent against
     its recursion limit instead of letting it report a broken toolchain."""
+    # The subscript comes after the render, not before: the family decorator hands back an
+    # already-specialized type, so ``GatedGiveUp[state_ty]`` is not a thing but
+    # ``GatedGiveUp.with_template(...)[state_ty]`` is. Naming the session's own state type here
+    # means the fields this tool reads are checked against the state it actually runs on.
     return GatedGiveUp.with_template(
         description=description,
         reason=reason_description,
         label=label,
-    ).bind(min_attempts).as_tool(name)
+    )[state_ty].bind(min_attempts).as_tool(name)
--- a/composer/authoring/tools.py
+++ b/composer/authoring/tools.py
@@ -12,10 +12,11 @@ instantiate time. Unskip does not vary.
 """
 
 from dataclasses import dataclass
-from typing import Callable, Literal, Sequence, override
+from typing import Callable, Literal, NotRequired, Sequence, override
 
 from langchain_core.messages import AIMessage, AnyMessage, ToolMessage
 from langchain_core.tools import BaseTool
+from langgraph.graph import MessagesState
 from langgraph.types import Command
 from pydantic import Field
 
@@ -42,17 +43,28 @@ def _as_titles(titles: Titles | Sequence[PropertyTitle]) -> Titles:
     return lambda: snapshot
 
 
+class CurtailableState(MessagesState):
+    """The authoring state :class:`RecordSkip` reads.
+
+    ``budget_curtailed`` is where the budget monitor records its wrap-up order, and the focus
+    protection stands down when it is set. Optional because only the prover and foundry authors
+    have a budget to be curtailed by; the natspec and rustapp sessions share this tool and carry
+    no such field, and absent reads as "still trying"."""
+    budget_curtailed: NotRequired[bool]
+
+
 @dataclass(frozen=True)
 class SkipScope:
     """What ``record_skip`` is allowed to retire: the batch's titles, and the subset it must
     refuse.
 
-    ``protected`` is a thunk, not a set, because the answer changes during a session — the
-    budget monitor's wrap-up order tells the agent to skip everything that does not work, so a
-    protection that outlived that order would deadlock the session it was meant to curtail.
-    A backend with nothing to protect passes a thunk returning ``()``."""
+    The protection is unconditional only while the run is still trying. Once the budget monitor
+    orders a wrap-up it tells the agent to skip everything that does not work, so a protection
+    that outlived that order would deadlock the session it was meant to curtail. That state lives
+    in ``budget_curtailed``, which is where the tool reads it — not in a flag on this object,
+    which would not survive a checkpoint and could disagree with the state it shadows."""
     titles: Titles
-    protected: Titles
+    protected: frozenset[PropertyTitle] = frozenset()
 
 
 # ---------------------------------------------------------------------------
@@ -76,7 +88,11 @@ def _skip_result(
 
 @tool_family_display(_skip_label, _skip_result)
 @tool_family(SkipParams)
-class RecordSkip(WithInjectedId, WithAsyncDependencies[Command, SkipScope]):
+class RecordSkip(
+    WithInjectedId,
+    WithInjectedState[CurtailableState],
+    WithAsyncDependencies[str | Command, SkipScope],
+):
     """{description}"""
     property_title: PropertyTitle = Field(
         description="The snake_case title of the property from the batch listing"
@@ -84,28 +100,24 @@ class RecordSkip(WithInjectedId, WithAsyncDependencies[Command, SkipScope]):
     reason: str = Field(description="{reason}")
 
     @override
-    async def run(self) -> Command:
+    async def run(self) -> str | Command:
         with self.tool_deps() as scope:
             known = scope.titles()
-            protected = scope.protected()
+            protected = scope.protected
         if self.property_title not in known:
-            return tool_state_update(
-                self.tool_call_id,
+            return (
                 f"Unknown property title {self.property_title!r}. Must be one "
-                f"of: {', '.join(known)}.",
+                f"of: {', '.join(known)}."
             )
-        if self.property_title in protected:
-            return tool_state_update(
-                self.tool_call_id,
+        if self.property_title in protected and not self.state.get("budget_curtailed"):
+            return (
                 f"Property {self.property_title!r} is the focus of this run and cannot be "
-                "skipped. Decompose it into lemmas, strengthen the harness, or formalize a "
-                "weaker core of it and say so in your commentary — but do not retire it.",
+                "skipped. Prove a smaller statement it is built out of, or formalize the "
+                "strongest version of it that holds and say what you could not prove — but do "
+                "not retire it."
             )
         if not self.reason.strip():
-            return tool_state_update(
-                self.tool_call_id,
-                "A non-empty justification is required when skipping a property.",
-            )
+            return "A non-empty justification is required when skipping a property."
         return tool_state_update(
             self.tool_call_id,
             f"Recorded skip for property {self.property_title}.",
@@ -145,15 +157,16 @@ def skip_tools(
     *,
     skip_description: str,
     skip_reason: str,
-    protected: Titles | Sequence[PropertyTitle] = (),
+    protected: Sequence[PropertyTitle] = (),
 ) -> list[BaseTool]:
     """The skip / unskip pair, bound to the batch's property titles.
 
     ``protected`` names titles ``record_skip`` must refuse — the focus of a prioritized run,
-    whose whole point is that it is pursued rather than retired. Unskip is deliberately not
-    constrained: un-retiring a property is always allowed."""
+    whose whole point is that it is pursued rather than retired. The refusal lifts on its own
+    once the run is budget-curtailed. Unskip is deliberately not constrained: un-retiring a
+    property is always allowed."""
     get = _as_titles(titles)
-    scope = SkipScope(titles=get, protected=_as_titles(protected))
+    scope = SkipScope(titles=get, protected=frozenset(protected))
     return [
         RecordSkip.with_template(description=skip_description, reason=skip_reason)
         .bind(scope)
@@ -219,7 +232,14 @@ def _gated_give_up_label(p: dict, *, description: str, reason: str, label: str) 
     return f"Giving up on {label}: {p['reason']}"
 
 
-def _verify_attempts(state: dict) -> int:
+class GatedGiveUpState(MessagesState):
+    """The authoring state :class:`GatedGiveUp` reads and writes: the message history it counts
+    prover attempts from, and the two fields a surrender records."""
+    result: NotRequired[str]
+    failed: NotRequired[bool]
+
+
+def _verify_attempts(state: GatedGiveUpState) -> int:
     """How many times this session has put a spec in front of the prover.
 
     Counted from the message history rather than from ``prover_history``: that list only grows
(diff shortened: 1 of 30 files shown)


shellygr:
Claude answers:

Four of the five are right and are fixed in b89217b0. One I have pushed back on, with the part of it that was fair taken.

**1. The selected primary is not derived from the ranking. Agreed, fixed, though not quite as proposed.**

The bug is real: nothing forced `primary` to be the top of `ranked`, so `property_ranking.json` could rank A first while the run pursued D.

The recommendation was a combined priority score, validated ordering, and `primary = ranked[0]`. The formula has to exist either way, so rather than define it, tell the model to sort by it, and then check the model's arithmetic, the formula now lives in code and code does the picking:

```python
CRITICAL_MATCH_BONUS = 15
def priority(rp): return rp.score + (CRITICAL_MATCH_BONUS if rp.critical_match else 0)
```

`primary` and `supporting` are off the schema entirely. Each ranked property now carries its own `depends_on`, which is what keeps this to one LLM call: the model cannot name dependencies for a primary that code has not picked yet. `select` takes `max(ranked, key=priority)` and reads that entry's dependencies. Disagreement is now unrepresentable rather than merely detected, and `validate_ranking` got shorter.

**2. Cache hits should be checked for focus compatibility. Agreed, and this was worse than P2.**

Right, and the consequence is sharper than the finding says. A replay never calls `formalize`, so none of the focus protections apply to it: not the gated `give_up`, not the skip protection, not the rewritten prompt. A prioritized run could therefore return a comprehensive result that had skipped the one property it existed to establish, with exit code 0 and nothing proved.

Implemented as recommended. `_focus_satisfied` checks that every focus property is mapped to at least one check and none was retired, using `property_checks()` and `skipped` off `BackendResult`, so it is backend-agnostic. A hit that fails is treated as a miss and re-formalized. It cannot loop: the live tools refuse to retire the focus, and the one path that overrides them is a budget wrap-up, which yields `Curtailed` and is never cached.

**3. Cache inspection cannot locate prioritized results. Agreed, and it is broader than reported.**

`cache_autoprove.py` rebuilt the key as `FORMALIZATION_KEY(GeneratedCVL, bug_cache.items)` with no `plugins` argument, while the driver writes it from `batch.props` and `contributing_plugins`. Those already disagreed on master in two ways before this branch existed: `batch.props` is post-`POST_PROPERTY_KEY` plugin rewrites, and a contributing plugin adds a suffix. Pruning was a third.

Fixed exactly as recommended: `_resolve_formalization_key` reads `FINAL_PROPERTIES_KEY` (mode-parameterized) and keys off `final.items` + `final.tool_plugins`, falling back to the bug-analysis path for older records. Worth noting `FinalProperties` was written for precisely this and had no reader anywhere in the repo, and `AutoProveCacheTags.run_mode` was write-only; both now have one. Also fixed three docstrings pointing at `composer.meta.run` as the offline walker, which does not exist and never did.

**4. Five properties instead of one. Pushing back on the reframing, taking the magnitude.**

This one re-litigates a decision that was made deliberately. The design question was put explicitly and "top-1 plus its supporting cluster" was chosen over "exactly one property", with the batch specified as 1..k. So the behaviour matches intent rather than drifting from it.

On the merits: a dependency handed to the author as context is never verified, so the primary's proof could rest on an unchecked assumption. Keeping the supporting properties first-class means the judge demands rules for them and they are skip-protected, which is the point of pulling them in at all.

What is fair in the finding is the magnitude. "One property plus what it rests on" should not quietly mean five, so `MAX_SUPPORTING` drops from 4 to 2. A prioritized batch is now the primary plus at most two dependencies, and the PR body says so rather than leaving it to be inferred.

**5. Generated component labels can collide. Agreed, fixed.**

Confirmed by construction: `["Vault", "Vault", "Vault (2)"]` produced `Vault`, `Vault (2)`, `Vault (2)`. Since `by_key` is a dict, one entry shadowed the other and `select` took whichever came first.

`unit_index` is the real identity, as noted. Rather than restructure the key, which would cost the `PropertyKey` alignment shared with the report and the grouping model, labels are now suffixed until actually unused and asserted unique, so the label is an identity by construction. Test covers the pathological input.

---

`pytest -m "not expensive"`: 1133 passed. `pyright`: 0 errors. The 21 errors in the suite are a local pgvector testcontainer that will not stay up; they are in `test_rag_db` / `test_indexed_tool` and predate the branch.

Finding 3 is the one the suite cannot really cover, since it walks a live LangGraph store. Verifying that one by hand against a prioritized run and an existing comprehensive run.

composer/authoring/tools.py
```diff
@@ -180,3 +209,92 @@ def give_up_tool(
...
+
+# ---------------------------------------------------------------------------
+# Give up, behind an escalation gate
+# ---------------------------------------------------------------------------
+
+def _gated_give_up_label(p: dict, *, description: str, reason: str, label: str) -> str:
+    return f"Giving up on {label}: {p['reason']}"
+
+
+def _verify_attempts(state: dict) -> int:
```
jtoman: you can do better on this type I think

composer/authoring/tools.py
```diff
@@ -180,3 +209,92 @@ def give_up_tool(
...
+        for call in msg.tool_calls
+        if call["name"] == "verify_spec"
+    )
+
+
+@tool_family_display(_gated_give_up_label, None)
+@tool_family(GiveUpParams)
+class GatedGiveUp(
+    WithInjectedId,
+    WithInjectedState[dict],
```
jtoman: ibid, this is a weak sauce type bound.

composer/authoring/tools.py
```diff
@@ -180,3 +209,92 @@ def give_up_tool(
...
+            if self.sort == "exhausted" and (ran := _verify_attempts(self.state)) < floor:
+                return tool_state_update(
+                    self.tool_call_id,
+                    f"Rejected: you have run the prover {ran} time(s) on this task, and this run "
+                    f"requires at least {floor} before a property may be abandoned. This is the "
+                    "one property the run exists to establish. Decompose it into lemmas, add or "
+                    "strengthen a harness, or formalize a weaker core of it and state the gap — "
+                    "then verify. If the toolchain itself is broken, stop with "
+                    "sort='environment' instead.",
+                )
```
jtoman: for this, its my preference to make the return type of the tool be `str | Command`. And then do `return "Error message"` on the error path. It's more explicit about which branches are updating state and which are "oopsie didn't work"

composer/authoring/tools.py
```diff
@@ -180,3 +209,92 @@ def give_up_tool(
...
+
+
+def gated_give_up_tool(
+    *,
+    name: str,
+    description: str,
+    label: str,
+    min_attempts: int,
+    reason_description: str = "The reason for giving up on your task",
+) -> BaseTool:
```
jtoman: tioli, I prefer to make the dependencies of the state of a tool like this explicit. You can do this with:
```
class GatedGiveUpDeps(TypedDict, MessagesState):
   result: ReadOnly[str]
   failed: ReadOnly[bool]

class gated_give_up_tool[T: GatedGiveUpToo](
  *,
  ...,
 state_ty: type[T]
) -> BaseTool:
   return GatedGiveUp[state_ty].with_template(...)
```

composer/pipeline/keys.py
```diff
@@ -99,6 +103,7 @@ def _final_properties_key(
     threat_model_digest: str | None,
     with_refinement: bool,
     extra_context_digest: str | None = None,
+    run_mode: str = "comprehensive",
```
jtoman: literal type, remove the default.

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+        "properties whose violation would be a critical bug. Score the property's importance, "
+        "not how hard it looks to verify.",
+    )
+    critical_match: bool = Field(
+        description="True if this property corresponds to a concern the user explicitly "
+        "raised in the design document, threat model, or supplied context."
+    )
+    depends_on: list[PropertyTitle] = Field(
+        default_factory=list,
+        description="Titles of properties IN THIS PROPERTY'S OWN COMPONENT that would be needed "
```
jtoman: WHY ARE YOU SHOUTING ABOUT THIS

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+    )
+    rationale: str = Field(
+        description="One or two sentences justifying the score. Say what breaks if the "
+        "property does not hold."
+    )
+
+
+class PropertyRanking(BaseModel):
+    """Every candidate, scored. Which one the run pursues is derived from this, not stated
+    alongside it: see :func:`priority` and :func:`select`."""
```
jtoman: "see :func:`priority`" is a useless thing to send to an llm as a doc string in the output structure :)

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+    return rp.score + (CRITICAL_MATCH_BONUS if rp.critical_match else 0)
+
+
+@dataclass(frozen=True)
+class Candidate:
+    """One component's extracted properties, as offered to the ranker.
+
+    ``label`` is what the model sees and names in a :class:`PropertyKey`. It is derived
+    from the component's display name but forced unique across the run: display names are
+    free text the analysis agent produced under no uniqueness constraint, so resolving a
```
jtoman: ~I am pretty sure this "no uniqueness constraint" is not true. I mean, I think we insist on unique slugs, which *effectively* constrains the uniqueness of the free text. So the `unit_index` trick is *fine* but redundant wrt the component slug. If this is a defense against some ecosystems' component analysis potentially not having unique names, we need to make it part of the api to provide some stable, guaranteed way to identify a component within a contract without relying on "order in the parent list" (which is brittle)~

Nevermind, I'm realizing someone (eric I think?) added `unit_index` to be exactly this id. carry on.

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+    unit_index: int
+    label: ComponentName
+    props: list[PropertyFormulation]
+
+
+@dataclass(frozen=True)
+class Selection:
+    """The focus: which unit survives, and which of its properties, in order."""
+    unit_index: int
+    titles: list[PropertyTitle]
```
jtoman, line 121: is the order important?

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+    taken: set[str] = set()
+    out: list[Candidate] = []
+    for unit_index, name, props in units:
+        label, n = name, 1
+        while label in taken:
+            n += 1
+            label = f"{name} ({n})"
+        taken.add(label)
+        out.append(Candidate(unit_index, ComponentName(label), props))
+    assert len({c.label for c in out}) == len(out), "component labels must be unique"
```
jtoman: I think my mooted point above *DOES* apply here. Let's just require that component names have to be unique within each unit.

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+            "These ranked entries name properties that are not in the candidate listing: "
+            f"{unknown}. Use the [component, title] pairs exactly as given."
+        )
+    if (dupes := [k for k, n in seen.items() if n > 1]):
+        return f"These properties appear more than once in `ranked`: {dupes}. Rank each exactly once."
+    if (missing := [k for k in by_key if k not in seen]):
+        return (
+            f"These candidate properties are missing from `ranked`: {missing}. "
+            "Every candidate must be ranked, including the ones you consider unimportant."
+        )
```
jtoman: I recognize this is fiddling at the edges, but wouldn't it be better to collect all such errors in case there is some *catastrophic* screw up by the llm? that way it isn't fixing the most recent such error.

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+            if dep == rp.key[1]:
+                return (
+                    f"{rp.key} lists itself in `depends_on`. A property does not depend on itself."
+                )
+            if dep not in siblings:
+                return (
+                    f"{rp.key} lists {dep!r} in `depends_on`, which is not a property of its own "
+                    f"component ({home.label}). A run pursues one component's properties, so a "
+                    "dependency must come from the same component as the property that needs it."
+                )
```
jtoman: Feel free to wave this off, but as a dependency graph validation, this ain't it. Nothing stop cyclic dependencies, there is nothing said about what to do for transitive lemmas etc. If I had to guess, the "supporting lemmas" properties are probably going to end up as "these are important too" carriers, not *really* supporting lemmas. Further, the notion of a "supporting lemma" property isn't really a thing in our conceptual model; supporting lemmas usually fall out of trying to prove the high level property, and the property agent is actively discouraged from authoring properties that "follow from" other properties.

I think I understand what you're going for here, but I think this is a coarse approximation of what we really want, which is some notion of "these properties, collectively, define the core 'meta-property'".

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+        if len(titles) - 1 == MAX_SUPPORTING:
+            break
+        if dep in titles:
+            continue
+        titles.append(dep)
+
+    # Emit them in the component's own order so the author reads them as it would any
+    # batch, with the primary first.
+    order = {p.title: i for i, p in enumerate(home.props)}
+    head, tail = titles[0], sorted(titles[1:], key=lambda t: order[t])
```
jtoman: is... this actually important? the property order is kinda whatever, we don't do anything to constrain it.

composer/spec/types.py
```diff
@@ -103,6 +103,12 @@ def sort_description(self) -> str:
             case "safety_property":
                 return "Safety Property"
 
+type PropertyKey = tuple[ComponentName, PropertyTitle]
```
jtoman, line 106: wait, I swear this type is already defined somewhere. in the `report/` module? did you move it here?

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+        problem = validate_ranking(candidates, result)
+        if problem is None:
+            return result
+        _log.warning("property ranking rejected (attempt %d): %s", attempt + 1, problem)
+        messages = [
+            *messages,
+            HumanMessage(
+                f"Your ranking was rejected: {problem}\n\nProduce a corrected ranking."
+            ),
+        ]
```
jtoman, line 266: so we're clear on the mechanism here: the llm will just see the initial prompt, and then "your ranking was wrong"; but it will never actually see the ranking that was rejected. maybe that's fine and what you meant! but it's worth being clear eyed on what we're actually doing; we're telling the llm to fix a problem (and naming it) but as far as the poor llm knows, it never made this mistake and doesn't even know what we're talking about.

composer/spec/prioritize.py
```diff
@@ -0,0 +1,255 @@
...
+        max_supporting=MAX_SUPPORTING,
+    )
+    docs = [d for d in (design_doc, threat_model, *extra_context) if d is not None]
+    content: list[dict | str] = [
+        {"type": "text", "text": user},
+        *(d.to_dict(CacheLevel.SHORT) for d in docs),
+    ]
+    bound = llm.with_structured_output(PropertyRanking)
+
+    messages: list = [SystemMessage(system), HumanMessage(content=content)]
```
jtoman: `list` **huffs**

composer/templates/property_generation_prompt.j2
```diff
@@ -131,6 +131,27 @@ VERIFIED *or* that rule must be explicitly marked as "expected to fail" using th
...
+This run is being spent on the properties above and nothing else. They were chosen out of everything
+inferred for this contract because they are the ones worth verifying, so there is no other work to fall
+back on: whatever you establish here is the entire result.
+
+You therefore may not retire them. `record_skip` will refuse them, and the `give_up` tool rejects a
+`sort='exhausted'` stop until you have actually put a spec in front of the prover several times and can
+enumerate what you tried. When you are stuck, the move is to make the problem smaller, not to put it down:
+
+  - Decompose it. Prove the lemmas the property rests on as their own rules and build up to it.
+  - Strengthen the setup. A harness, a summary, or a `require` that rules out states the contract itself
```
jtoman: this bit about `require` might conflict with the extensive rag context we provide now basically saying "use require as a last resort". In fact, thinking about it more, this whole piece of advice is **awful** because it reads as license for "if you've got timeouts, add 'realistic' starting state assumptions until the rule passes" and I've spent months convincing this f---ing thing to not do that lol

composer/templates/property_generation_prompt.j2
```diff
@@ -131,6 +131,27 @@ VERIFIED *or* that rule must be explicitly marked as "expected to fail" using th
...
+    cannot reach will often turn an unprovable rule into a provable one.
+  - Weaken it honestly. If the full property will not go through, formalize the strongest core of it that
+    will, and say plainly in your commentary what you proved and what you did not.
+
```
jtoman: worth tying this back to the constitutional principle we state above. I'm under no illusions about this being a power word, but it's worth being very clear about what we want here. This reads as "don't give up, you have to deliver something". But a "weakened to be useless" or "says nothing" property is still bad.

composer/spec/source/harness.py
```diff
@@ -163,7 +163,7 @@ def lift_harnessed(
 
     Declared beside the models because every consumer of the harnessed view
     must build it identically — the prover pipeline, and any offline walker
-    re-deriving component cache keys (``composer.meta.run``): the harnessed
```
jtoman, line 166: did I never commit this module? 🤔

composer/spec/source/report/schema.py
```diff
@@ -15,17 +15,15 @@
...
+    ComponentName, PropertyFormulation, PropertyKey, PropertyTitle, PropertyType, RuleName,
+)
 
 type RuleRef = tuple[str, RuleName]
 """A rule's identity: ``(spec_file, name)``. A name is only unique within a spec, so the defining
 spec file disambiguates a rule re-stated under the same name in another spec (and collapses a single
 definition — e.g. an imported structural invariant — seen through several component runs)."""
 
-type PropertyKey = tuple[ComponentName, PropertyTitle]
-"""A property's identity: ``(component, title)`` — the cross-reference key groups use for members."""
```
jtoman, line 26: you did move it, okay

composer/spec/source/report/schema.py
```diff
@@ -304,10 +309,30 @@ class Finding(BaseModel):
     provenance: FindingProvenance | None = None
 
 
+class DeprioritizedProperty(BaseModel):
```
jtoman: I feel like this could and should just inherit from `PropertyFormulation` right?

composer/spec/source/author.py
```diff
@@ -697,14 +733,22 @@ async def run(
...
     fires any pending reminders are dropped — moot, since the warning tells
-    the agent to ignore prover/feedback outcomes anyway."""
+    the agent to ignore prover/feedback outcomes anyway.
+
+    The wrap-up also lifts a prioritized run's focus protection: the order it gives is to skip
+    what does not work, which the protection would otherwise refuse."""
+    def _wrap_up(_s: SourceCVLGenerationState, _c: object) -> dict:
+        if focus is not None:
+            focus.lifted = True
+        return {"required_validations": [], "budget_curtailed": True}
```
jtoman: OOOOOOOF. This is spooky action at a distance, and some magic side effect on an in-memory object determining the behavior of the tool. just have the focus tools read the "budget_curtailed" part of the state, it already records the information you want?

jtoman: oh, I guess there the "protected name" is a callable that doesn't see the state. But there's gotta be a better way...

(not for nothing this in memory flag isn't checkpointed. not the worst thing in the world but it does mean its possible to be in a state where one state flag thinks we're in "wrap it up" and the other flag doesn't)

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
+You are deciding where a single formal-verification budget should be spent on a smart contract
+system.
+
+You are given every candidate property that property inference produced, grouped by the component
```
jtoman: friendly reminder that the llm comes into existence having no earthly idea what "property inference is", what the "formal verification budget" we're talking about is, etc. It can probably pick it up from background context, but pretend you put this prompt in front of a random 2nd year PhD PL student who doesn't work at certora. They would have so many questions!

(Aside, I've noticed this is a pretty pervasive failure mode when you ask an llm to write a prompt, it writes a prompt with all of its own background knowledge assumed, it leaves out "obvious" stuff that's only obvious because it has all the context for the task...)

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+it was inferred for. Each carries a unique title, a `sort` ("invariant", "safety_property",
+"attack_vector") and an English description. You may also be given the system's design document, a
+threat model, and further context the user supplied.
+
+Score every candidate, and record what each one rests on. You are not choosing a winner: the run
```
jtoman: "what each one rests on"

I'm a flesh and blood human and even I don't know what this means? What the score rests on? What the candidate rests on (dependencies?)

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+picks the highest-priority property from your scores, so the scores are the decision.
```
jtoman: I meeeeean. You're telling it that its not picking a winner, and then telling it that its picking a winner, but via the scores it chooses. If the agent wants to pick a winner, it can rank its "favorite" as score 100 and the rest as 1. I'm honestly not sure what this whole bit is trying to say; why is it important the llm isn't thinking its picking a winner?

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+
+For each candidate give:
+  - `score`: 0-100, how much *proving it* would raise confidence in the overall correctness of the
+    system. Score importance, not difficulty.
```
jtoman: There is absolutely no scale provided here. Any ordering within 0-100 provides information for relative positioning I suppose, but what does a score 100 look like? 5? 25? Are we fine just leaving that up to the agent?

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+  - `critical_match`: whether it answers a concern the user actually raised, in the design
+    document, the threat model, or the supplied context. This is worth a fixed bonus on top of the
+    score, so mark it honestly rather than generously.
+  - `depends_on`: the properties from the SAME component that would be needed to state or discharge
+    this one — the lemmas it assumes, the invariants its argument leans on. Only the run's chosen
+    property has its dependencies pursued, and only the first {{ max_supporting }} of them, so put
+    the ones the proof genuinely cannot do without first. A property that is merely *related* is
+    not a dependency; leave it out.
```
jtoman: see my comments elsewhere about my dubiousness around the utility of what we're asking for here. There is a sort mismatch here in what we're telling it to find within "depends on".

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+  - `rationale`: one or two sentences. Say what breaks if the property does not hold.
+
+WHAT MAKES A PROPERTY WORTH THE RUN:
+  W1. Its violation is a real loss — funds, access, or an accounting identity — not a cosmetic
+      deviation.
```
jtoman: > "real loss"

It is at this point, I realize we've never even told the agent this is for a Web3/Defi protocol. So it has no idea what we're talking about here.

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+  W2. It constrains behaviour across many entry points rather than one narrow path.
```
jtoman: Commounweaulth speulling jump scaure.

jtoman: But also, what does this mean? This sounds like you're asking for "cross cutting", but I'd argue that's not a really important metric. A `rebalanceShares` function that is called in one entry point is probably the most important in the whole contract.

composer/templates/property_prioritization_system.j2
```diff
@@ -0,0 +1,44 @@
...
+    document, the threat model, or the supplied context. This is worth a fixed bonus on top of the
+    score, so mark it honestly rather than generously.
+  - `depends_on`: the properties from the SAME component that would be needed to state or discharge
+    this one — the lemmas it assumes, the invariants its argument leans on. Only the run's chosen
+    property has its dependencies pursued, and only the first {{ max_supporting }} of them, so put
+    the ones the proof genuinely cannot do without first. A property that is merely *related* is
+    not a dependency; leave it out.
+  - `rationale`: one or two sentences. Say what breaks if the property does not hold.
+
+WHAT MAKES A PROPERTY WORTH THE RUN:
```
jtoman: what run! this is the first time you've mentioned a "run"

jtoman (changes requested):

composer/templates/property_prioritization_system.j2
shellygr: so I should add headers similar to the ones in other prompts?

composer/templates/property_prioritization_system.j2
shellygr: with how tied we are to solidity I don't think this is not sensible

composer/authoring/tools.py
shellygr: Claude answers: done

composer/authoring/tools.py
shellygr: Claude answers: done

composer/authoring/tools.py
shellygr: Claude answers: done

composer/authoring/tools.py
shellygr: Claude answers: not done as sketched. The state type is now GatedGiveUpState rather than dict, but the generic parameter does not survive the tool_family decorator: it returns an already-specialized _TemplatedTool, so GatedGiveUp[state_ty] is a pyright error.

composer/pipeline/keys.py
shellygr: Claude answers: done

composer/spec/prioritize.py
shellygr: Claude answers: done

composer/spec/prioritize.py
shellygr: Claude answers: done

composer/spec/prioritize.py
shellygr: Claude answers: yes, unit_index. The label the model sees now carries it.

composer/spec/prioritize.py
shellygr: Claude answers: no. dropped.

composer/spec/prioritize.py
shellygr: Claude answers: not done. The label is now "<unit_index>: <name>", so two components sharing a name are already distinct and nothing has to be required of the analysis.

composer/spec/prioritize.py
shellygr: Claude answers: done

composer/spec/prioritize.py
shellygr: Claude answers: done. depends_on is gone; the ranker groups candidates into claims and the run takes the group holding the highest-scoring property.

composer/spec/prioritize.py
shellygr: Claude answers: no, it was not. dropped.

composer/spec/types.py
shellygr: Claude answers: moved, from report/schema.py.

composer/spec/prioritize.py
shellygr: Claude answers: done. The rejected ranking is replayed before the correction.

composer/spec/prioritize.py
shellygr: Claude answers: done

composer/templates/property_generation_prompt.j2
shellygr: Claude answers: done. Removed.

composer/templates/property_generation_prompt.j2
shellygr: Claude answers: done

composer/spec/source/harness.py
shellygr: Claude answers: composer.meta.run never existed. The references point at cache_autoprove now, which is the only walker there is.

composer/spec/source/report/schema.py
shellygr: Claude answers: yes.

composer/spec/source/report/schema.py
shellygr: Claude answers: done

composer/spec/source/author.py
shellygr: Claude answers: done. Reads budget_curtailed from the state.

shellygr: Claude answers: done, same fix. The flag is gone, so there is nothing left to disagree with the state.

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done. That phrasing is gone with depends_on.

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done. Dropped.

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done. Anchored with a scored example at each level.

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done

shellygr: Claude answers: done. Dropped.

composer/templates/property_prioritization_system.j2
shellygr: Claude answers: done

composer/authoring/tools.py
jtoman: I think `GatedGaveUp.with_template(...)[t]` might work? I honestly forget; but I think eric used parametricity and tool families successfully.

composer/pipeline/core.py
```diff
@@ -549,24 +549,33 @@ async def _prepare_formalization() -> Formalizer[FormT, U] | StagedFormalizer[Fo
...
+    if batches and run.run_mode is RunMode.PRIORITIZED:
         with named_budget_or_nop("property_extraction"):
             batches, deprioritized = await _prioritize(
                 backend.analysis_spec.properties_key, backend.artifact_store,
                 batches, run, phases["extraction"], threat_model, extra_context,
             )
 
+    staged = await staged_task
+    if not batches:
+        raise ValueError("No properties extracted from any component.")
```
jtoman, line 580: run in parallel imo

jtoman (approved):
much better imo

composer/authoring/tools.py
shellygr: Claude answers: done. You were right, the subscript just has to come after with_template rather than before it. The state type is SourceCVLGenerationState now, so `failed` turned out to be `bool | None` and required.

composer/pipeline/core.py
shellygr: Claude answers: not done. It already overlaps: staged_task is a live task started before extraction, so the setup keeps running while the ranking waits on its model call. Gathering them explicitly reads better but changes nothing at runtime, and the fuller version breaks two tests that pin this pair as overlapped-but-not-gated.

PR #202, opened 2026-09-02
Check that declared source paths exist during component analysis

shellygr:
## What

The `path` fields on `SourceExplicitContract`, `SourceExternalActor` and `ExistingFromSource` are
written by the component-analysis agent. Nothing checked them against the filesystem.
`validate_solidity_connectivity` covered names, duplicates and the interaction graph, and it was
typed over `BaseApplication` with no project root to check anything against.

So a declared path that resolves to nothing travels a long way before anyone notices. It goes
through `run_setup_part1`'s `name_to_path`, into the transitive closure, into `extra_files` in
`run_autosetup_phase`, and finally into the AutoSetup subprocess argv, where `parse_contract_files`
raises `File ... does not exist`. By then an hour has passed and several analysis and bug-analysis
phases have run.

The shape that prompted this: a repository whose build project sits one or more directories below
the repository root. Paths are project-root-relative, so every declared path has to carry that
leading directory. A path assembled from an import statement or from a build tool's source
directory instead of from the agent's own file-tool output loses it, and every contract in the
submission loses it the same way.

## How

The validator takes a third `project_root: Path | None` and checks the declared paths inside its
existing single pass over components. Path complaints join the same accumulated message as the
graph ones, so a submission that is wrong in both ways is corrected in one retry instead of two.
`project_root` threads through `Ecosystem.validate_analysis`, which follows the existing
`locate_main` precedent of handing ecosystem callables the run's facts. Solana and Soroban take an
ignored parameter; nothing in their models carries a source path.

Containment is lexical, meaning relative and never stepping up out of the root. Resolving instead
would follow symlinks, and the forbidden-read predicate deliberately keeps `.sol` readable under
`lib/` and `node_modules/`, which are routinely symlinked. The agent is entitled to name a file
there and has no way to restate it as a non-symlinked path, so a resolving check would reject it
with no way to comply. There is a test pinning that.

The error says where to look instead. One walk of the tree collects readable files carrying the
declared name, narrowed to those whose path ends with the declared path, so the dropped-directory
case names exactly one real file rather than reporting "not found". One walk covers every bad path
in the submission, and wholly withheld directories are pruned during descent rather than filtered
out afterwards, because this runs while the agent waits on its retry.

Cache hits get the same check. `root_cache_key` is
`sha256(project_root | doc_hash | relative_path | contract_name)`, which covers neither the source
tree nor the rules in force, and the cloud namespaces the cache per user and repository. Without
this, a model written before the check existed is replayed unvalidated on every rerun and the gate
never fires at all. An entry that fails validation is stale, so it is re-derived and the existing
`cache_put` replaces it.

The prompt and the `path` field descriptions now state the convention the validator enforces. They
reject nothing on their own; they say what the check requires.

## Scope note for reviewers

`SourceExternalActor.path` is checked too, which is wider than the failure above. It is the same
class of unchecked agent-written string and it reaches the summarizer prompt. The error offers
"omit the path instead" as the remedy, and an actor with no path is passed over silently
downstream, so a repository that passes today can start reporting a validation error. That is
intended, not a regression.

## Testing

`tests/test_solidity_component_paths.py` is new: 17 tests over acceptance, a path missing its
leading directory (asserting the hint names the one real file), several files sharing a name, no
file of that name anywhere, a directory, absolute and escaping paths, a path under a withheld
directory, a symlinked dependency being accepted, several bad paths arriving in one message, graph
errors and path errors arriving together, the external actor with and without a path,
`project_root=None`, greenfield, and `FromSourceApplication` where the existing contract is checked
and the fresh one is not. Two more cover the cache branch in both directions.

`tests/test_analysis_input_phrase.py` gained rendering assertions for the two path bullets and the
omit-the-path clause. `tests/test_pipeline_overlap.py` now asserts the driver forwards a real
project root.

Each new test was checked by breaking the code it covers and confirming it fails.

Targeted run: 91 passed. `pyright composer/ analyzer sanity_analyzer certora_autosetup`: 0 errors.

## Review round

`forbidden_read` now comes off the pipeline input rather than the module-level `fs_forbidden_read`,
so the check asks the question the run's own file tools answer. It threads beside `project_root`
through `Ecosystem.validate_analysis`, and natspec carries it on its `MentalModel`. Normalizing the
two surface forms of a `GlobalExcludeArg` uses `make_exclude_pred`, exported from graphcore in
Certora/graphcore#37, now merged, and the pin sits on graphcore master.

The directory prune in the relocation walk stays on `fs_withheld_subtree`, which is a walk-cost
measure over prover and VCS scratch rather than a readability rule. A prune cannot be derived from
an arbitrary exclusion: `lib/` is excluded whole while the `.sol` under it stays readable. Four new
tests pin the split, including the regex form of `forbidden_read` and a hint that must not name a
file the run withholds.

Advancing the graphcore pin also brings its token-usage accounting fix, so the fake message in
`tests/test_token_usage.py` carries `usage_metadata`, whose input total includes the cache buckets.

### DIFF
--- a/composer/spec/system_analysis.py
+++ b/composer/spec/system_analysis.py
@@ -2,10 +2,11 @@ import enum
 import logging
 import os
 from dataclasses import dataclass
-from pathlib import Path, PurePath, PurePosixPath
+from pathlib import Path, PurePosixPath
 from typing import Mapping, NotRequired, Any, Callable, TypedDict
 
 from graphcore.graph import MessagesState, FlowInput
+from graphcore.tools.vfs import GlobalExcludeArg, make_exclude_pred
 
 
 from composer.spec.context import (
@@ -19,7 +20,7 @@ from composer.spec.system_model import (
 )
 from composer.spec.types import SourceIdentifier
 from composer.spec.service_host import ServiceHost, Sort
-from composer.spec.util import fs_forbidden_read, fs_withheld_subtree, slugify_filename
+from composer.spec.util import fs_withheld_subtree, slugify_filename
 from composer.tools.thinking import RoughDraftState, get_rough_draft_tools
 from composer.diagnostics.budget import budget_monitor
 
@@ -59,9 +60,10 @@ class _PathFault(enum.Enum):
     WITHHELD = "is not readable through your file tools; name a file they hand back instead."
 
 
-def _path_fault(root: Path, declared: str) -> _PathFault | None:
+def _path_fault(root: Path, declared: str, withheld: Callable[[str], bool]) -> _PathFault | None:
     """What is wrong with a declared source path, or ``None`` when it names a file under *root*
-    that the agent's own file tools hand back.
+    that the agent's own file tools hand back. *withheld* is the run's own ``forbidden_read``,
+    normalized, so the check asks the same question the tools serving those files answer.
 
     Containment is lexical — relative, and never stepping up out of the root. Resolving the path
     would follow symlinks, and a dependency tree symlinked in under ``lib/`` or ``node_modules/``
@@ -81,29 +83,34 @@ def _path_fault(root: Path, declared: str) -> _PathFault | None:
         return _PathFault.DIRECTORY
     if not candidate.is_file():
         return _PathFault.MISSING
-    if fs_forbidden_read(PurePath(declared)):
+    if withheld(declared):
         return _PathFault.WITHHELD
     return None
 
 
-def _readable_files_named(root: Path, names: set[str]) -> dict[str, list[str]]:
+def _readable_files_named(
+    root: Path, names: set[str], withheld: Callable[[str], bool]
+) -> dict[str, list[str]]:
     """Project-root-relative paths of every readable file whose base name is in *names*, from a
     single walk of the tree.
 
     One walk for all the wanted names rather than one per name: a project that sits a directory
     below the root drops the same leading component from every path it declares, so a whole
-    submission's worth of distinct base names goes wrong together. Wholly withheld directories are
-    pruned on the way down instead of filtered out of the results, which keeps a report directory
-    or a ``.certora_internal`` tree off the walk entirely. This runs while the agent waits on its
-    retry, so both matter.
+    submission's worth of distinct base names goes wrong together. This runs while the agent waits
+    on its retry, so the cost of the walk itself matters: ``fs_withheld_subtree`` prunes prover and
+    VCS scratch on the way down, keeping a report directory or a ``.certora_internal`` tree off the
+    walk entirely. That prune is a cost measure, not the readability rule — *withheld* is, and it
+    cannot serve as a prune, since a directory it excludes whole (``lib/``, say) still holds .sol
+    the agent may read.
     """
     found: dict[str, list[str]] = {name: [] for name in names}
     for dirpath, dirnames, filenames in os.walk(root):
         rel_dir = Path(dirpath).relative_to(root)
         dirnames[:] = [d for d in dirnames if not fs_withheld_subtree(rel_dir / d)]
         for filename in filenames:
-            if filename in found and not fs_forbidden_read(rel_dir / filename):
-                found[filename].append((rel_dir / filename).as_posix())
+            rel = (rel_dir / filename).as_posix()
+            if filename in found and not withheld(rel):
+                found[filename].append(rel)
     return {name: sorted(paths) for name, paths in found.items()}
 
 
@@ -152,19 +159,23 @@ class _PathComplaint:
         )
 
 
-def _path_complaints(app: BaseApplication, project_root: Path) -> list[str]:
+def _path_complaints(
+    app: BaseApplication, project_root: Path, forbidden_read: GlobalExcludeArg
+) -> list[str]:
     """Every declared source path in *app* that names no readable file under *project_root*,
-    worded for the agent's retry."""
+    worded for the agent's retry. *forbidden_read* is normalized once here rather than per path,
+    so its regex form is compiled once for the whole submission."""
+    withheld = make_exclude_pred(forbidden_read)
     complaints: list[_PathComplaint] = []
     for c in app.components:
         if isinstance(c, SourceExplicitContract | ExistingFromSource):
-            fault = _path_fault(project_root, c.path)
+            fault = _path_fault(project_root, c.path, withheld)
             if fault is not None:
                 complaints.append(
                     _PathComplaint(f"Contract {c.solidity_identifier}", c.path, fault)
                 )
         elif isinstance(c, SourceExternalActor) and c.path is not None:
-            fault = _path_fault(project_root, c.path)
+            fault = _path_fault(project_root, c.path, withheld)
             if fault is not None:
                 complaints.append(_PathComplaint(
                     f"External actor {c.name}", c.path, fault,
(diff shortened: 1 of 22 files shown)


composer/spec/system_analysis.py
```diff
@@ -28,15 +38,166 @@ class AnalysisPromptParams(TypedDict):
...
+    parsed = PurePosixPath(declared)
+    if parsed.is_absolute() or ".." in parsed.parts:
+        return _PathFault.OUTSIDE
+    candidate = root / declared
+    if candidate.is_dir():
+        return _PathFault.DIRECTORY
+    if not candidate.is_file():
+        return _PathFault.MISSING
+    if fs_forbidden_read(PurePath(declared)):
+        return _PathFault.WITHHELD
```
jtoman: I'm confused how this happened?

composer/spec/system_analysis.py
```diff
@@ -28,15 +38,166 @@ class AnalysisPromptParams(TypedDict):
...
+    below the root drops the same leading component from every path it declares, so a whole
+    submission's worth of distinct base names goes wrong together. Wholly withheld directories are
+    pruned on the way down instead of filtered out of the results, which keeps a report directory
+    or a ``.certora_internal`` tree off the walk entirely. This runs while the agent waits on its
+    retry, so both matter.
+    """
+    found: dict[str, list[str]] = {name: [] for name in names}
+    for dirpath, dirnames, filenames in os.walk(root):
+        rel_dir = Path(dirpath).relative_to(root)
+        dirnames[:] = [d for d in dirnames if not fs_withheld_subtree(rel_dir / d)]
```
jtoman, line 109: I was today years old when I learned you could control the recursion behavior of `os.walk` by just mutating `dirnames`. This language is *insane* dude.

composer/spec/system_analysis.py
```diff
@@ -28,15 +38,166 @@ class AnalysisPromptParams(TypedDict):
...
+
+
+def _relocate(declared: str, same_named: Mapping[str, list[str]]) -> str:
+    """The half of a missing-path complaint that says where to look instead: the readable files
+    carrying the declared file name, narrowed to those whose path *ends* in the declared path.
+    A path that lost a leading directory tail-matches exactly one real file, which turns the
+    complaint into an answer the agent can copy."""
+    name = PurePosixPath(declared).name
+    candidates = same_named[name]
+    tail_matches = [c for c in candidates if c.endswith("/" + declared)]
```
jtoman, line 124: Wait, I'm confused. This is basically finding candidates where the LLM clearly forgot a leading directory?

composer/spec/system_analysis.py
```diff
@@ -28,15 +38,166 @@ class AnalysisPromptParams(TypedDict):
...
+    """
+    parsed = PurePosixPath(declared)
+    if parsed.is_absolute() or ".." in parsed.parts:
+        return _PathFault.OUTSIDE
+    candidate = root / declared
+    if candidate.is_dir():
+        return _PathFault.DIRECTORY
+    if not candidate.is_file():
+        return _PathFault.MISSING
+    if fs_forbidden_read(PurePath(declared)):
```
jtoman: FWIW, we should be using the `forbidden_read` on `SourceFields` which makes up the input to the pipeline. `fs_forbidden_read` is just the default (that we happen to use always, so its worth considering whether we need to keep `forbidden_read` as a separate part of the input)

composer/spec/system_analysis.py
jtoman, line 294: FWIW, this is the `input` field which holds the source field to use.

composer/spec/system_analysis.py
```diff
@@ -115,9 +278,10 @@ async def run_component_analysis[T: BaseApplication](
     extra_input: list[str | dict],
     expected_main_id: SourceIdentifier | None = None,
     *,
+    project_root: Path | None,
```
jtoman, line 299: doesn't this duplicate the data available on `SourceCode`? When will this be non-none and `input !is SourecCode`?

composer/spec/system_analysis.py
```diff
@@ -126,9 +290,26 @@ async def run_component_analysis[T: BaseApplication](
     ``has_doc`` below reflects whether a design document is actually present, not
     merely whether an ``input`` object was passed, so a source-only run renders the
     no-doc prompt branch and never dereferences a missing ``content``.
+
+    ``project_root`` is the tree the validator resolves any source paths the model declares
+    against; it is required (not defaulted) so a new call site has to say what the model's paths
+    mean, and is ``None`` only when the run has no source tree.
```
jtoman: which is also the semantics (if you will) of input == SystemDoc or input == None. I think we're duplicating data.

jtoman (changes requested):

composer/spec/system_analysis.py
shellygr: Claude answers: I don't have a case of the model actually doing it. The branch is there for consistency with the hints: the walk behind them prunes `.certora_internal` and the report directories, so a path into one is something the validator would never suggest, and accepting it would point AutoSetup at a prover's own copy of a contract that is readable at its canonical path anyway. Every prover run leaves such a copy behind, so the files really are there. tests/test_solidity_component_paths.py covers it.

composer/spec/system_analysis.py
shellygr: Claude answers: It is the only way to prune a directory without walking it first, which is what this loop is after.

composer/spec/system_analysis.py
shellygr: Claude answers: Yes. Candidates start as every readable file with that base name, and the tail match keeps the ones whose path ends with what was declared. That is the dropped-leading-directory case, and it usually leaves exactly one file for the agent to copy.

composer/spec/system_analysis.py
shellygr: Claude answers: Done in 4f7c2f1e. The validator takes the run's `forbidden_read` and normalizes it with `make_exclude_pred`, which I exported from graphcore in Certora/graphcore#37, so the regex and the callable form are read the way the tools read them. The pin here points at that branch commit until #37 merges. The directory prune in the walk stays on `fs_withheld_subtree`: a prune cannot come from an arbitrary exclusion, since `lib/` is excluded whole while the .sol under it stays readable.

composer/spec/system_analysis.py
shellygr: Claude answers: They come apart in the natspec from-source workflow, which calls this with `input` a `SystemDoc` and a real root off `mental_model.source_root` (composer/spec/natspec/system_analysis.py). That workflow is generating contracts, so it has no `contract_name` or `relative_path` and cannot hand over a `SourceCode`. If you would rather it were one object, the shape is splitting `(project_root, forbidden_read)` out of `SourceFields` into a base both callers pass. Say so and I will do that instead.

composer/spec/system_analysis.py
shellygr: Claude answers: Same as the parameter above: natspec from-source has a root and no `SourceCode`, so the two are not the same condition.

composer/spec/system_analysis.py
jtoman: oh. hmmm. Heck! okay then

jtoman (approved):

composer/spec/system_analysis.py
shellygr: Claude answers: graphcore#37 is merged, so the pin here now points at graphcore master (2177f57) instead of the branch commit.

PR #208, opened 2026-09-04
Bound the AutoSetup wait, and stop the task holding it from outliving the run

shellygr:
## What went wrong

Two cloud runs sat for 22 hours until the job's hard time cap killed them, and that cap skips the results upload, so both were a total loss.

An Anthropic API error killed one extraction task in each run. The failure unwound out of the pipeline, the run's exit handlers ran, and then `asyncio.run`'s teardown cancelled what was left and waited for it. One of the two runs had an AutoSetup subprocess live inside an orphaned task at that moment; its log ends with the `CancelledError` coming out of `_drain` and then the debug line in the `finally`, so it was last seen going into `await proc.wait()` with nothing bounding it. The other had finished AutoSetup cleanly twelve seconds earlier and froze at the same point in teardown with no subprocess anywhere on its stack.

So this PR is not the fix for the incident. The process-level backstop is AIAutoProver#80.

## The change

**`_reap`**: wait up to 30s, kill the child, wait another 30s, then give up and let the run continue. Nothing is killed on the normal path, where the first wait returns as soon as the pipes are at EOF. A child that could not be reaped is reported as a failed AutoSetup. The bound is a loop timer rather than anything about the child, so it holds whether or not the child can be reaped.

**The orphan**: the formalization task is started to overlap with extraction and awaited after it, and nothing cancels it in between. An extraction that raises therefore leaves it running into teardown. `_alongside` ties it to the block it accompanies, so it is cancelled and awaited while the loop is still healthy rather than after the exit handlers have run.

**The log**: the reap said where it began and nothing on the way out, which is why the incident logs are ambiguous about whether the wait ever returned. It now reports the exit status and how long it took. `console-autoprove` drives the loop through an explicit `Runner`, which gives the close a before and an after and a place to name the tasks still running when it starts; those are what the unbounded cancel-and-wait inside `close()` is spent on.

## Tests

`tests/test_autosetup_reap.py`: a child that exits reports its status, a child that ignores SIGTERM is killed and reaped within the bound, the reap ends when it runs in the `finally` of an already-cancelled task, and it says how it ended.

### DIFF
--- a/composer/pipeline/core.py
+++ b/composer/pipeline/core.py
@@ -36,7 +36,8 @@ from typing import (
     Protocol, Any, ClassVar, Concatenate, cast, Awaitable, Sequence, Callable, ContextManager, overload
 )
 from abc import ABC, abstractmethod
-from contextlib import nullcontext
+from contextlib import nullcontext, asynccontextmanager
+from collections.abc import AsyncIterator, Coroutine
 
 
 from composer.io.multi_job import TaskInfo
@@ -458,6 +459,24 @@ async def _run_pipeline_inner[P: enum.Enum, FormT: BackendResult, H, A: Artifact
             extra_context=extra_context, max_bug_rounds=max_bug_rounds, ecosystem=ecosystem,
         )
 
+
+@asynccontextmanager
+async def _alongside[T](coro: Coroutine[Any, Any, T]) -> AsyncIterator[asyncio.Task[T]]:
+    """Run ``coro`` beside the body, and end it with the body rather than letting it outlive it.
+
+    A task still running when its scope is left is reachable only by ``asyncio.run``'s teardown,
+    which cancels it after the run's exit handlers have finished and then waits for it with no
+    bound of its own. Whatever the task is holding at that point is held there.
+    """
+    task = asyncio.create_task(coro)
+    try:
+        yield task
+    except BaseException:
+        task.cancel()
+        await asyncio.gather(task, return_exceptions=True)
+        raise
+
+
 # ---- the driver --------------------------------------------------------------
 async def run_pipeline_inner[P: enum.Enum, FormT: BackendResult, H, A: ArtifactIdentifier, U: FeatureUnit, Main, App: BaseApplication, Pre](
     backend: PipelineBackend[P, FormT, H, A, U, Main, App, Pre],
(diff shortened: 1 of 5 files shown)


composer/spec/source/autosetup.py
```diff
@@ -176,7 +206,12 @@ def log_complete(self, returncode: int):
             raise
         finally:
             _logger.debug("AutoSetup process complete, waiting for exit")
-            returncode = await proc.wait()
+            returncode = await _reap(proc)
```
jtoman, line 220: I'm confused, do we know why this hung? how did `proc.kill(); proc.wait()` not work??

shellygr: I am also confused how it happened, but either the kill in the exception didn't kill the process on a timely manner, or `gather` returned without the process actually ending. The question is whether we think that this fix is hiding a more serious issue.

Claude's commentary:
 ▎ We don't. The run sat in that wait until the job's hard time cap killed
 ▎ the process, so there was nothing left to inspect.
 ▎
 ▎ On the kill: it only runs in the except BaseException branch. If the gather returns
 ▎ normally with the child still alive, nothing kills it, and the old await proc.wait()
 ▎ in the finally just blocks. And where we do kill, the wait after it was that same
 ▎ unbounded wait, so a child that doesn't die on SIGKILL parks the run the same way. The
 ▎ bound is what ends it either way.

shellygr: Claude answers: We do now, from the container logs. One of the two runs ends with the CancelledError coming out of `_drain`, then the debug line in the `finally`. So `asyncio.run`'s teardown cancelled the drain, `proc.kill()` did run, and the run was last seen going into the unbounded `proc.wait()`.

The kill only exists in the except branch, and the wait after it is that same unbounded wait, so nothing there was a bound on anything. Whether the wait itself is what held for 22h I can't tell from that log, because nothing logged when it returned. The other run froze at the same point in teardown with its autosetup already exited 0, so this isn't the whole incident.

I've pushed the rest: the task holding the subprocess is now cancelled with the extraction it runs beside instead of being left for teardown, the reap logs how it ended, and there's a test for the path that actually happened.

composer/pipeline/core.py
```diff
@@ -458,6 +465,24 @@ async def _run_pipeline_inner[P: enum.Enum, FormT: BackendResult, H, A: Artifact
...
+    bound of its own. Whatever the task is holding at that point is held there.
+    """
+    task = asyncio.create_task(coro)
+    try:
+        yield task
+    except BaseException:
+        task.cancel()
+        await asyncio.gather(task, return_exceptions=True)
+        raise
+
```
jtoman, line 484: I'm confused; this is just gather? or a task group? What is this doing that either of those two abstractions don't already do for us?

jtoman (commented):
more being annoying

PR #212, opened 2026-09-04
Retry a rejected report grouping instead of falling back

### LINKED ISSUE #211: Report grouping falls back to one bucket when the model string-wraps its structured output
## What happened

A cloud AutoProver run produced a report whose `coverage.warnings` began with
`FALLBACK GROUPING APPLIED`, collapsing all 11 formalized properties into the single
`general` bucket (`total_groups: 1`, `properties_per_group_min/max: 11`). The report is
still correct about what was proved; it just loses every high-level heading.

The cause, from the executor log:

```
composer.spec.source.report.build WARNING report: grouping failed: 1 validation error for GroupingResult
groups
  Input should be a valid list [type=list_type, input_value='{"groups":[\n{"slug":"...  ...}]}', input_type=str]
```

The model returned the entire `GroupingResult` document as a JSON **string** in the
`groups` field rather than as a list of objects. The content was complete and
well-formed; only the encoding was wrong.

## Why the report gave up immediately

`call_grouping_llm` (`composer/spec/source/report/grouping.py:76`) makes one structured
call and returns whatever comes back. There is no validation feedback and no retry, so
`build` (`composer/spec/source/report/build.py:79-106`) catches the pydantic error and
goes straight to `build_fallback_grouping`.

Two things are worth noting:

1. **A syntactic defect costs the whole grouping.** The model's answer was usable. A
   string that parses as the expected object should not end up in the same bucket as a
   transport failure.
2. **The ranker already solves this and grouping does not.** `rank_properties`
   (`composer/spec/prioritize.py`) validates its structured output and replays the
   rejection to the model for one corrective attempt, precisely because a single bad
   structured response is recoverable. Grouping has no equivalent, so one blip is
   terminal for the report's headings.

## Suggested direction

Decode a string that parses as the expected shape, and give grouping the corrective
retry the ranker has, before the single-bucket fallback is reached. The fallback should
stay as the last resort it was designed to be, not the first thing a recoverable
response hits.

## Reproducing

The failure is a property of the model's response, so it does not reproduce on demand. A
unit test can drive it directly by having the bound LLM return
`{"groups": "<the JSON document as a string>"}` and asserting the report still gets real
groups.

shellygr:
Fixes #211.

A cloud run produced a report whose `coverage.warnings` began with `FALLBACK
GROUPING APPLIED`, collapsing all 11 formalized properties into the single
`general` bucket. The cause:

```
report: grouping failed: 1 validation error for GroupingResult
groups
  Input should be a valid list [type=list_type, input_value='{"groups":[...]}', input_type=str]
```

The model serialized the whole `GroupingResult` document into the one field
meant to hold the list. The grouping was complete and correct; only the
encoding was wrong, and it landed in the same bucket as a transport failure.

## What changes

Grouping gets the corrective retry the ranker already has.
`call_grouping_llm` previously made one structured call and let any schema miss
reach `build`'s fallback. It now retries once with the rejection appended,
mirroring `rank_properties` in `composer/spec/prioritize.py`. The correction is
appended to the original request rather than sent as a second turn, so each
attempt stays one self-contained user message.

The single-bucket fallback is unchanged and still covers a grouping that could
not be obtained at all.

## Tests

Two added, both failing if `max_attempts` drops to 1:

- the retry fires once, carries the rejection, and the first attempt does not
- after the retry it raises, so the caller's fallback still applies

1200 pass, pyright clean.

### DIFF
--- a/composer/spec/source/report/grouping.py
+++ b/composer/spec/source/report/grouping.py
@@ -1,23 +1,32 @@
 """LLM-driven grouping of inferred properties into high-level audit claims.
 
-A single structured LLM call takes the `FormalizedProperty` list and partitions it into high-level
+A structured LLM call takes the `FormalizedProperty` list and partitions it into high-level
 `PropertyGroup`s (the "P-NN" headings) — each property in exactly one group, while the rules those
 properties are formalized by may surface under several groups. Each group's status is rolled up from
 its members' rules' verdicts. Groups are identified by the slug the LLM assigns — a per-run snapshot.
 
+A response that misses the schema costs the report every heading it has, so two things stand
+between a malformed answer and the fallback: a grouping the model serialized into a string is
+decoded rather than rejected, and a response that still does not validate is retried once with the
+rejection appended.
+
 A single ``general`` fallback group (every property in one group) is used by `build` when the LLM
 call raises, validation rejects the grouping, or the grouping covers no properties.
 """
+import json
+import logging
 from typing import Iterable
 
 from langchain_core.language_models.chat_models import BaseChatModel
 from langchain_core.messages import HumanMessage, SystemMessage
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, ValidationError, field_validator
 
 from composer.templates.loader import load_jinja_template
 from composer.spec.source.report.schema import (
     FormalizedProperty, GroupStatus, Outcome, PropertyGroup, PropertyKey, RuleRef,
 )
+
+_log = logging.getLogger(__name__)
 
 FALLBACK_SLUG = "general"
 FALLBACK_TITLE = "General"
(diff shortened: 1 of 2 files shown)


composer/spec/source/report/grouping.py
```diff
@@ -72,25 +81,67 @@ class GroupingResult(BaseModel):
...
+        handed back untouched for pydantic to reject as it normally would."""
+        if not isinstance(v, str):
+            return v
+        try:
+            decoded = json.loads(v)
+        except json.JSONDecodeError:
+            return v
+        if isinstance(decoded, dict) and "groups" in decoded:
+            return decoded["groups"]
+        return decoded
```
jtoman: you already have the ability to make the LLM correct its mistakes; hardcoding a post-hoc correction for a schema mangling observed *once* in real deployments is an oversteer imo.

jtoman (changes requested):

composer/spec/source/report/grouping.py
shellygr: Claude answers: Agreed, dropped the validator and its three tests in e452d86c. The retry covers the case on its own.

jtoman (approved):

PR #214, opened 2026-09-05
Drop the eager structural-invariant phase

shellygr:
## What was blocking

The structural-invariant phase sat inside `prepare_formalization`, and the driver joins that
before any component is formalized. Its formulation half already ran alongside AutoSetup, so
that was never the cost. The cost was the invariant CVL generation: a full
`batch_cvl_generation` with real prover jobs, serial after the gather because it needs
AutoSetup's `prover_config`. #200's own commit message measured it at two of four hours on one
run, and it was paid whether or not any property turned out to need an invariant.

It was also the run's only fatal path, since a give-up there raised and killed everything, and
it was all-or-nothing: a budget curtailment quarantined the spec and no component imported it.

## What replaces it

The component author writes the invariant it needs, in its own spec, and the prover checks it in
the same run as the rule that cites it. The author prompt already said to do this on an
impossible starting state. What it lacked was any reason to prefer that over a bare `require`,
which was cheap and always available. The ladder now makes the invariant the default rung and
spells out what the last rung costs. A `structural_invariant_guidance` tool carries the
selection knowledge the deleted judge held, pulled only when the author hits an unreachable
prestate rather than sitting in every prompt.

Two ways this could have gone green while proving nothing, both closed:

- An invariant that no property names is dropped from the report as an orphan
  (`collect.py`). `PublishResultTool` now reads the declared rule set off the last prover run
  and passes it to `validate_property_rules`, which already had the both-directions branch
  behind that argument and no caller on this path. An unmapped invariant blocks publication.
- `expect_rule_failure` on an invariant that a `requireInvariant` cites turns every citing rule
  into an unproven assumption while the run reports green. Refused in the author prompt, on the
  tool's own description, and in the judge's laundering practice.

The judge's overconstrained-inputs criterion was scoped to `preserved` blocks, which is where a
bare `require` used to land. Without a pre-proven `invariants.spec` it will land in rule bodies
too, so that criterion widens to any `require` that fixes the shape of the prestate.

## Notes for review

- **`prepare_formalization` is AutoSetup plus the summaries built on its config.** No `await`
  moved. The overlap with property inference and #200's prioritization already existed; this
  makes the thing being overlapped cheap. `tests/test_pipeline_staged_formalizer.py` still pins
  the ranking ahead of the join, unchanged.
- **`tests/test_prover_prepare_formalization.py` is the new load-bearing test.** It asserts no
  `CVL_GEN`-phase task is launched before the barrier, which is the precise statement of the
  change.
- **The deleted tape lane was the Counter harness's designated tool-coverage lane.** Its turns
  (manual and recipe search, the research sub-agent, the typechecker rejection path,
  skip/unskip, expect-fail/passage) move into the component lane rather than being lost.
- **A warm local cache from before this commit will mis-hit.** `FORMALIZATION_KEY` is keyed on
  the property batch and plugin ids only, so a replayed `GeneratedCVL` still imports
  `invariants.spec`, which nothing writes now. Use a fresh `--cache-ns`. Cloud is unaffected: a
  git+ install puts the commit in the root key.
- **Old cache namespaces keep `structural-inv` and `invariant-cvl` subtrees** that the walkers
  no longer render, and a project re-run in place keeps a stale `certora/specs/invariants.spec`
  and its conf and properties files. Both are expected, not data loss.
- **`formalization_preparation` keeps its budget key** because `BUDGET_PHASES` is derived from
  it and dropping it would reject existing budget files. Nothing left under that phase installs
  a monitor, so it is attribution-only now, and the budget matrix's T3 scenario is deleted
  rather than reworded.
- **No frontend change needed.** The webapp's `invariants.spec` handling is a display allowlist
  that `.spec` already covers, so historical reports keep rendering.
- **`scripts/autoprove_cache_explorer.py` still imports the deleted module.** It has been
  unimportable for unrelated reasons for a while and is outside both pyright and pytest. Left
  alone deliberately; say the word and I will delete it.

## Verification

`COMPOSER_STRICT_TEMPLATES=1 uv run pytest -m 'not expensive' tests/` passes 1163, and pyright
reports 0 errors, against current master with the updated graphcore pin. The expensive
integration lane still needs a run, and so does one real end-to-end run on a project with a
pre-change baseline.

The acceptance criterion for that run is not in `report.json`: a rule that requires its way past
a counterexample reports VERIFIED and GOOD. Compare the generated `autospec_*.spec` files
against the baseline for invariants authored, `requireInvariant` citations, and bare `require`s
in rule bodies, parsed with the real CVL parser rather than grep.


composer/templates/structural_invariant_guidance.j2
```diff
@@ -0,0 +1,61 @@
+A counterexample that starts from a state the contract could never actually reach is telling you that the
```
shellygr: probably needs the standard headers for prompts as it had before

source_edits_integration.md
```diff
@@ -51,10 +51,13 @@ The new top-level field:
...
-  with editing denied by design, so its outcomes are always about the on-disk
-  source. Foundry-backend reports never carry entries.
+  applied**. Un-edited and gave-up components never appear. Foundry-backend reports
+  never carry entries.
+- An invariant a component authored to support one of its own properties is covered by
+  that component's edit record, so it is established against exactly the source that
+  component's rules were proven against. There is no longer a phase that proves
+  invariants against the pristine source ahead of everything else, and so no separate
+  entry for one.
 - `applied_edits` lists every edit that survived to the final working copy, in the
```
shellygr:
```suggestion
  component's rules were proven against.
- `applied_edits` lists every edit that survived to the final working copy, in the
```

composer/templates/structural_invariant_guidance.j2
shellygr: Claude answers: done, split into <background> and <guidance> and gave the four failure modes Good/Bad pairs. No role preamble though: cvl_system_prompt.j2 already opens the author's system prompt with the FV-engineer-at-Certora role, so the Web3 framing is in context already.

composer/templates/property_judge_prompt.j2
```diff
@@ -176,13 +176,18 @@ Evaluate each rule for its *preconditions*. Remember that these rules/invariants
...
 NOTE: portions of the input state may be unconstrained/underconstrained if they don't matter for the purposes of the property being verified.
 
 ### Criteria 4: Overconstrained Inputs
 
-You may also consider the inverse; over-constraining the input space may cause the prover to miss real bugs. In particular, any constraints in
-`preserved` blocks that are *not* a `requireInvariant` statement (i.e., a *proven* assumption) should be subject to intense scrutiny.
+You may also consider the inverse; over-constraining the input space may cause the prover to miss real bugs. In particular, any constraint that fixes the
+*shape of the prestate* — a relationship among the contract's own state fields that the contract maintains — and is *not* a `requireInvariant` statement
+(i.e., a *proven* assumption) should be subject to intense scrutiny, wherever it appears: in a `preserved` block or in the body of a rule. Such a
+relationship is exactly what an invariant is for, and a bare `require` asserts it where an invariant would prove it.
```
jtoman: "asserts": somewhat confusing, as `assert` roughly means "proven" in cvl.

composer/templates/property_judge_prompt.j2
```diff
@@ -176,13 +176,18 @@ Evaluate each rule for its *preconditions*. Remember that these rules/invariants
...
+A `require` that merely bounds a rule's own free inputs is not this; that is Criteria 3's business.
```
jtoman: tighten this

composer/templates/property_judge_prompt.j2
```diff
@@ -293,6 +298,8 @@ against a *contract* output is the cure, not the defect.
 
 Verify that the specification covers all {{ properties | length }} listed properties. For each property:
 - If the spec contains rules/invariants addressing it: evaluate those rules under criteria 1-7.
+- A supporting invariant listed alongside a rule is evidence about that rule's proof, not coverage of
+  the property in its own right; judge coverage by the rules that assert the property.
```
jtoman: this doesn't really make sense as an item under the "For each property:" directive. I'm also not entirely sure what this trying to say.

composer/templates/structural_invariant_guidance.j2
```diff
@@ -0,0 +1,80 @@
+<background>
+A counterexample that starts from a state the contract could never actually reach is telling you that the
+prover does not know something you do: the contract's fields stand in some relationship that the code
+maintains, and nothing in your spec says so. For example, the prover may hand you a starting state where a
+user's balance of a token is significantly greater than that token's total supply — clearly infeasible for a
+well-behaved token, but perfectly admissible to a tool that was never told otherwise.
```
jtoman: this is too "cute"; "perfectly admissible to a tool" is a weird way of talking about the prover. "never told otherwise" is likewise a weird way to talk around "adding a constraint"

jtoman (commented):
some nits.

composer/templates/property_judge_prompt.j2
shellygr: Claude answers: done, it says "assumes the relationship instead of establishing it" now. Same slip was in two more places you did not see, the author's ladder and the Criteria 8 bullet; both fixed.

composer/templates/property_judge_prompt.j2
shellygr: Claude answers: done, four lines instead of five, and dropped the sentence that restated the one before it.

composer/templates/property_judge_prompt.j2
shellygr: Claude answers: done, moved out of the list. It is about property_rules entries that name a supporting invariant the rules cite with requireInvariant, so it now says that below the list instead of pretending to be a fourth branch.

composer/templates/structural_invariant_guidance.j2
shellygr: Claude answers: done, it now says the spec is missing a constraint and that the prover explores every state the spec admits.

jtoman (approved):
RIP king

PR #215, opened 2026-09-05
Multi-spec buffers: parallel, per-precision CVL verification

jar-ben:
## Summary

Today the CVL-generation agent authors **one** spec with **one** global `methods{}` block, so
every rule is verified under the *intersection* of what all rules need precise — one property
that needs an expensive function (nonlinear math, hashing, a heavy external) kept exact forces
**every** rule to pay that cost. That coupling is a common source of timeouts, and a single
combined run cannot escape it.

This PR replaces the single-spec model with **named spec buffers**: the agent partitions its
properties into several self-contained CVL buffers, each with its **own** `methods{}` (so a
function summarized in one buffer can stay exact in another), and each **verified and reviewed
independently and in parallel**. A function's precision cost is now paid only by the buffer that
needs it.

It also converts the prover interaction from a blocking, one-spec-at-a-time call into an
**async submit/collect** model, so buffers prove concurrently and the agent keeps working
(authoring the next buffer, or processing a finished result) instead of blocking on the slowest
job.

Validated end-to-end on a real project (see *Validation*).

## The buffer model

- **Named buffers.** `put_buffer` / `edit_buffer` / `get_buffer` / `list_buffers` /
  `delete_buffer` author CVL buffers; `submit_buffer` / `collect_results` verify them (replacing
  `verify_spec`).
- **Run-target vs shared.** A **run-target** buffer declares `property_rules` (the properties it
  verifies + their rule names) and is run by the prover. A **shared** buffer (`is_run_target=false`)
  carries infrastructure (ghosts, common invariants, helper CVL, token/oracle models, summaries)
  and is never run itself.
- **Selective imports.** A shared buffer is imported only by the run-targets that declare it — so
  a subset of run-targets can share one base while another subset shares a different one, and
  editing a shared buffer re-verifies only its importers.
- **Coverage is enforced.** Every non-skipped property must appear in **exactly one** run-target,
  and each rule lives in exactly one buffer (`validate_coverage` / `validate_disjoint_rules`).
- **Per-buffer completion.** Publishing requires, per run-target, both a **prover** and a
  **feedback** stamp keyed to a content digest over the buffer's *import closure* — so editing a
  buffer (or a shared buffer it imports) invalidates exactly the buffers affected, and nothing
  else is re-run.

## The async prover model

- `submit_buffer(name)` launches a prover job in the **background** and returns immediately
  (concurrency-throttled; idempotent; supersedes a stale in-flight job for the same buffer).
- `collect_results()` **drains finished jobs without blocking** and returns a status board
  (`complete` / `running` / `needs (re)submission`); it blocks only when there is nothing else to
  do. The agent works a strict priority order — process a finished result, else author/submit the
  next buffer, else block on the next completion.
- **Snapshot isolation.** Each job materializes its buffers to tagged `{name}__{digest}.spec`
  files with sibling imports retargeted, so concurrent jobs never read each other's in-flight
  edits.
- **Shared-base refine.** Editing a shared buffer invalidates every importer (including
  already-verified ones); the board lists them under `needs (re)submission` so the agent re-runs
  exactly those.

## Guidance & advisories

- **Prompt guidance** now leads with the intersection argument and **strongly recommends**
  splitting by precision need + adding over-approximating performance summaries *preemptively*
  (before a monolithic run), including the `assert`-only vs `satisfy` soundness split. It is a
  recommendation, not a mandate.
- **Duplication linter** (advisory, report-once): flags `methods{}`/ghost declarations duplicated
  verbatim across run-targets and suggests hoisting them into a shared buffer. It only *tells* —
  it never blocks.

## Reliability fixes (independent; bundled here — can be split if preferred)

- **Anthropic mid-stream retry** (`composer/llm/anthropic.py`): a mid-stream connection drop
  surfaces as a raw `httpx.RemoteProtocolError`; treat it as retryable.
- **Cloud re-fetch, never re-prove** (`composer/prover/cloud.py`): a transient failure while
  downloading a *completed* job's results used to bubble up and re-run the whole (often
  hours-long) proof. Now the fetch itself is retried with capped exponential backoff against the
  already-`SUCCEEDED` job.
- **Configurable sanity timeout** (`certora_autosetup/setup/sanity.py`): `AUTOPROVER_SANITY_TIMEOUT`
  (default 1200s) caps the sanity phase's per-job timeout.

## Config knobs

- `AUTOPROVER_MAX_SPEC_BUFFERS` — max run-target buffers (default **6**).
- `AUTOPROVER_SANITY_TIMEOUT` — sanity-phase per-job timeout in seconds (default **1200**).

## Files

New substrate & tools:
- `composer/spec/source/spec_buffers.py` — pure buffer substrate (model, import closure, digests,
  coverage/disjointness validation, completion stamps, dup detection).
- `composer/spec/source/buffer_tools.py` — the authoring tools (`put/get/edit/list/delete_buffer`,
  cap enforcement).

Wiring & orchestration:
- `composer/spec/source/prover.py` — async `submit_buffer` / `collect_results`, materialization,
  job supersession, dup-linter surfacing.
- `composer/spec/source/author.py` — buffer-only authoring mode, per-buffer review/feedback,
  buffer guidance.
- `composer/spec/source/pipeline.py`, `autoprove_common.py`, `composer/spec/cvl_generation.py`,
  `composer/certora_env.py`, `certora_autosetup/cache/content_cache.py` — supporting wiring.
- `composer/templates/property_generation_system_prompt.j2` — prompt updates.

## Testing

**51 new unit tests** across 7 files, all pure (no live prover / no jar):
- `test_spec_buffers.py` (26) — substrate: ownership, import closure, digests, coverage,
  disjoint rules, completion stamps, dup detection, buffer-map reducer.
- `test_buffer_completion.py` (6), `test_spec_buffer_prover.py` (5), `test_buffer_tools.py` (5) —
  completion tracking, async submit/collect + per-buffer feedback, cap enforcement.
- `test_cloud_fetch_retry.py` (2), `test_sanity_timeout.py` (4), `test_anthropic_retry.py` (3) —
  the reliability fixes.

Full non-expensive suite green (**1219 passed**; the only errors are pre-existing
`test_rag_db.py [postgres]` env cases that need the local RAG container). `pyright` clean.

## Validation (live cloud run)

Run on a testing real project:
- The agent **split into 6 run-target buffers across multiple property classes** (
  init/ownership, impl/upgrade-auth, attack-vectors; combinatorial-collateral: core, operations,
  attack) — i.e. multiple buffers *within* a property class, not merely one per class.
- **52 cloud jobs driven in parallel** (up to 3 concurrent in flight), with re-submissions as
  buffers were refined.
- **0 rules timed out** (146 VERIFIED / 75 VIOLATED instantiations; the VIOLATED set is the
  expected-failure attack vectors + refinement CEXes).
- The **dup-linter fired advisorily** (e.g. `owner()` across the 3 buffers) and the agent
  correctly treated it as a non-blocking suggestion.
- The **cloud re-fetch path saw 0 connection failures** across the full 52-job fetch load (the
  same volume that previously surfaced transient stalls).

## Follow-ups (out of scope)

- **Comment-only re-verification** (TODO in `spec_buffers.py`): a pure-comment edit changes the
  buffer digest and re-runs an identical proof. The fix is a separate comment-stripped *prover*
  digest distinct from the raw *feedback* digest — noted, not done here.
- A spurious-CEX issue observed on the run traces to an **agent-authored summary in
  `custom_summaries.spec`** (SafeTransferLib path not covered) and the summarizer↔CVL-agent
  one-way handoff — a **pre-existing, unrelated** issue, not introduced by this PR; handed off
  separately.

### DIFF
--- a/composer/spec/source/artifacts.py
+++ b/composer/spec/source/artifacts.py
@@ -100,7 +100,7 @@ class ProverArtifactStore(ArtifactStore[SpecIdentity, GeneratedCVL]):
         self, spec: SpecIdentity, base_config: dict | None, spec_path: Path,
     ) -> None:
         """The prover conf for the run: the generation's final ``state["config"]`` plus
-        the fixed run overlay (shared with the live ``verify_spec`` run). No-op if no
+        the fixed run overlay (shared with the live prover run). No-op if no
         base config."""
         if base_config is None:
             _log.warning("no base config for %s; skipping conf dump", spec.stem)
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -28,7 +28,9 @@ from composer.spec.cvl_generation import (
 from composer.prover.core import run_prover, CexHandler, ProverCallbacks, ProverReport
 from composer.spec.source.live_explorer import VersionedHistory, LiveEditTools, WIPE_HISTORY
 from composer.spec.source.prover import setup_prover_config_in
-from composer.spec.source.spec_buffers import SpecBuffersExtra, spec_buffers_enabled
+from composer.spec.source.spec_buffers import (
+    SpecBuffersExtra, combined_buffers_view, run_targets, spec_buffers_enabled,
+)
 from composer.spec.source.buffer_tools import (
     put_buffer, get_buffer, edit_buffer, list_buffers, delete_buffer,
 )
@@ -653,7 +655,10 @@ class EditorAwareFeedbackTool(
                 vfs=self.state["vfs"],
                 version_history=self.state["version_history"],
             )
-            return await judge(snap, spec, skipped, self.rebuttals, self.tool_call_id)
+            # In multi-buffer mode the single curr_spec is empty; review the buffers as one document.
+            buffers = self.state.get("buffers") or {}
+            review_spec = combined_buffers_view(buffers) if run_targets(buffers) else spec
+            return await judge(snap, review_spec, skipped, self.rebuttals, self.tool_call_id)
 
     @override
     def _version_history(self) -> Sequence[str]:
@@ -666,6 +671,35 @@ class PropertyGenSystemParams(TypedDict):
     source_editing: bool
 
 _PropertyGenSysTemplate = TypedTemplate[PropertyGenSystemParams]("property_generation_system_prompt.j2")
+
+#: Appended to the system prompt only when multi-buffer authoring is enabled (spec_buffers_enabled).
+_SPEC_BUFFERS_GUIDANCE = """
+## Splitting the spec into verification buffers
+
+Instead of one spec, you may author several named CVL **buffers**, each a self-contained spec that is
+verified and reviewed on its own. Use `put_buffer` / `edit_buffer` / `get_buffer` / `list_buffers` /
+`delete_buffer` instead of the single-spec put/get/edit tools when you split.
+
+When to split:
+- Properties have **conflicting precision needs** — one buffer can summarize a function that another
+  keeps exact (each buffer has its own `methods{}`), with no global-methods-block conflict.
+- To **isolate the hard/slow properties**: put the many easy properties in one buffer that verifies and
+  is approved once and never re-touched, while you iterate on a small hard buffer in isolation — its
+  re-verification and re-review cost only that buffer.
+
+How:
+- Put **shared** infrastructure (ghosts, common invariants, helper CVL functions, token/oracle models)
+  in a buffer with `is_run_target=false`, and have run-target buffers `import "<shared>.spec"` and list
+  it in their `imports`.
+- Each **run-target** buffer declares its `property_rules` (the properties it verifies + their rule
+  names). Across all run-target buffers, every non-skipped property must appear in **exactly one**
+  buffer, and each rule lives in exactly one buffer.
+- `verify_spec` then verifies every run-target buffer independently and in parallel; a buffer already
+  verified at its current content is skipped. Editing a shared buffer re-verifies every buffer that
+  imports it.
+
+A single run-target buffer is exactly the one-spec case, so only split when it helps.
+"""
 
 #: The prover's tool extension: contributions come from plugins deriving
 #: ``CertoraProverTools``, dispatched via their ``certora_prover_tools`` hook.
@@ -794,6 +828,8 @@ async def batch_cvl_generation(
     sys_prompt : list[RawPromptInput | type[CacheMarker]] = [
         _PropertyGenSysTemplate.bind({"source_editing": editing is not None}).render_to
     ]
+    if spec_buffers_enabled():
+        sys_prompt.append(_SPEC_BUFFERS_GUIDANCE)
 
     added_tools : list[BaseTool] = []
     if editing_tools is not None:
@@ -1012,8 +1048,13 @@ async def batch_cvl_generation(
             # unformalizable" judgment — it's the budget talking. Keep the agent's account.
             return Curtailed(None, detail=res_state["result"])
         return GaveUp(reason=res_state["result"])
-    d = res_state["curr_spec"]
-    assert d is not None
+    # In multi-buffer mode the artifact is the buffers combined into one document; else curr_spec.
+    _buffers = res_state.get("buffers") or {}
+    if run_targets(_buffers):
+        d = combined_buffers_view(_buffers)
+    else:
+        d = res_state["curr_spec"]
+        assert d is not None
     applied_edits: list[AppliedEdit] = []
     if editing is not None:
         for edit_id in res_state["version_history"]:
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -28,6 +28,10 @@ from composer.spec.cvl_generation import (
 from composer.prover.core import run_prover, CexHandler, ProverCallbacks, ProverReport
 from composer.spec.source.live_explorer import VersionedHistory, LiveEditTools, WIPE_HISTORY
 from composer.spec.source.prover import setup_prover_config_in
+from composer.spec.source.spec_buffers import SpecBuffersExtra, spec_buffers_enabled
+from composer.spec.source.buffer_tools import (
+    put_buffer, get_buffer, edit_buffer, list_buffers, delete_buffer,
+)
 from composer.spec.context import WorkflowContext, CVLGeneration, CacheKey, SourceCode
 from composer.spec.types import PropertyFormulation, PropertyTitle
 from composer.pipeline.core import GaveUp, ToolBinder, InjectingToolExtension, Curtailed
@@ -86,7 +90,7 @@ class SourceAuthorExtra(TypedDict):
 # ``vfs`` comes from ProverStateExtra (NotRequired, no merge op — replaced
 # wholesale by commit_edit / revert_to_edit); the generation input always
 # seeds it explicitly.
-class SourceCVLGenerationExtra(CVLGenerationExtra, ProverStateExtra, SourceAuthorExtra, VersionedHistory):
+class SourceCVLGenerationExtra(CVLGenerationExtra, ProverStateExtra, SourceAuthorExtra, VersionedHistory, SpecBuffersExtra):
     pass
 
 class SourceCVLGenerationInput(SourceCVLGenerationExtra, FlowInput):
@@ -893,8 +897,17 @@ async def batch_cvl_generation(
         )
     else:
         b = b.with_tools(env.source_tools)
+    # Multi-buffer authoring tools (opt-in): the agent partitions the spec into several named
+    # buffers verified independently. Off by default so the single-curr_spec flow is untouched.
+    buffer_authoring: list[BaseTool] = [
+        put_buffer(SourceCVLGenerationState), get_buffer(SourceCVLGenerationState),
+        edit_buffer(SourceCVLGenerationState), list_buffers(SourceCVLGenerationState),
+        delete_buffer(SourceCVLGenerationState),
+    ] if spec_buffers_enabled() else []
     task_graph = b.with_tools(
         static_tools()
+    ).with_tools(
+        buffer_authoring
     ).with_tools(
         feedback_suite
     ).with_tools(
@@ -972,7 +985,8 @@ async def batch_cvl_generation(
                 reminders_channel=[],
                 vfs=restored_vfs,
                 version_history=restored_history,
-                spec_stem=spec_stem
+                spec_stem=spec_stem,
+                buffers={},
             )
         )
     except BudgetExceeded as e:
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -1,4 +1,4 @@
-from typing import AsyncIterator, NotRequired, override, Literal, Annotated, Sequence, Protocol, Callable
+from typing import AsyncIterator, NotRequired, override, Literal, Annotated, Sequence, Protocol, Callable, cast
 
 from typing_extensions import TypedDict
 from contextlib import asynccontextmanager
@@ -14,7 +14,7 @@ from graphcore.tools.schemas import (
     WithAsyncImplementation, WithImplementation, WithInjectedId, WithInjectedState,
     WithAsyncDependencies
 )
-from graphcore.graph import tool_state_update, RawPromptInput, CacheMarker, SummaryConfig
+from graphcore.graph import tool_state_update, tool_return, RawPromptInput, CacheMarker, SummaryConfig
 from graphcore.tools.vfs import VFSAccessor, VFSState
 
 from composer.authoring.judge import PropertyFeedbackProtocol
@@ -29,13 +29,15 @@ from composer.prover.core import run_prover, CexHandler, ProverCallbacks, Prover
 from composer.spec.source.live_explorer import VersionedHistory, LiveEditTools, WIPE_HISTORY
 from composer.spec.source.prover import setup_prover_config_in
 from composer.spec.source.spec_buffers import (
-    SpecBuffersExtra, combined_buffers_view, run_targets, spec_buffers_enabled,
+    SpecBuffersExtra, buffer_review_text, buffer_state_digest, check_buffer_completion,
+    combined_buffers_view, run_targets, spec_buffers_enabled, validate_coverage,
+    validate_disjoint_rules,
 )
 from composer.spec.source.buffer_tools import (
     put_buffer, get_buffer, edit_buffer, list_buffers, delete_buffer,
 )
 from composer.spec.context import WorkflowContext, CVLGeneration, CacheKey, SourceCode
-from composer.spec.types import PropertyFormulation, PropertyTitle
+from composer.spec.types import PropertyFormulation, PropertyTitle, RuleName
 from composer.pipeline.core import GaveUp, ToolBinder, InjectingToolExtension, Curtailed
 from composer.pipeline.plugin_api import ProvidedTools
 from composer.spec.source.plugin import CertoraProverTools, CVLAuthorState
@@ -162,16 +164,38 @@ class PublishResultTool(
 
     @override
     async def run(self) -> Command | str:
-        if (err := check_completion(self.state, self.state["version_history"])) is not None:
-            return err
-        with self.tool_deps() as titles:
-            if (err := validate_property_rules(self.property_rules, self.state["skipped"], titles)) is not None:
+        buffers = self.state.get("buffers") or {}
+        if run_targets(buffers):
+            # Multi-buffer completion: every run-target buffer verified AND reviewed at its current
+            # digest (per-buffer stamps), plus a clean property/rule partition across buffers.
+            skipped_pairs = [(str(s.property_title), str(s.reason)) for s in self.state["skipped"]]
+            if (err := check_buffer_completion(
+                buffers, self.state["validations"], self.state["required_validations"],
+                skipped=skipped_pairs, version_history=self.state["version_history"],
+            )) is not None:
                 return err
+            with self.tool_deps() as titles:
+                skip_titles = {str(s.property_title) for s in self.state["skipped"]}
+                if (err := validate_coverage(buffers, all_properties=set(titles), skipped=skip_titles)) is not None:
+                    return f"Completion REJECTED: {err}"
+            if (err := validate_disjoint_rules(buffers)) is not None:
+                return f"Completion REJECTED: {err}"
+            pr = [
+                PropertyRuleMapping(property_title=cast(PropertyTitle, p), rules=cast(list[RuleName], rs))
+                for b in run_targets(buffers) for p, rs in b.property_rules.items()
+            ]
+        else:
+            if (err := check_completion(self.state, self.state["version_history"])) is not None:
+                return err
+            with self.tool_deps() as titles:
+                if (err := validate_property_rules(self.property_rules, self.state["skipped"], titles)) is not None:
+                    return err
+            pr = self.property_rules
         return tool_state_update(
             self.tool_call_id,
             "Accepted",
             result=self.commentary,
-            property_rules=self.property_rules,
+            property_rules=pr,
             failed=False,
         )
 
@@ -655,14 +679,44 @@ class EditorAwareFeedbackTool(
                 vfs=self.state["vfs"],
                 version_history=self.state["version_history"],
             )
-            # In multi-buffer mode the single curr_spec is empty; review the buffers as one document.
-            buffers = self.state.get("buffers") or {}
-            review_spec = combined_buffers_view(buffers) if run_targets(buffers) else spec
-            return await judge(snap, review_spec, skipped, self.rebuttals, self.tool_call_id)
+            return await judge(snap, spec, skipped, self.rebuttals, self.tool_call_id)
 
     @override
     def _version_history(self) -> Sequence[str]:
         return self.state["version_history"]
+
+    @override
+    async def run(self) -> Command:
+        """Single-spec: the base flow. Multi-buffer: review each run-target buffer whose feedback
+        stamp is missing or stale (its text or an import changed) in isolation, and stamp
+        ``feedback:<buffer>`` per approved buffer — so an approved, unchanged buffer is never
+        re-reviewed and the hard buffer is reviewed alone."""
+        buffers = self.state.get("buffers") or {}
+        targets = run_targets(buffers)
+        if not targets:
+            return await super().run()
+
+        skipped = self.state["skipped"]
+        skipped_pairs = [(str(s.property_title), str(s.reason)) for s in skipped]
+        vh = self.state["version_history"]
+        validations = self.state["validations"]
+
+        def digest(name: str) -> str:
+            return buffer_state_digest(buffers, name, skipped=skipped_pairs, version_history=vh)
+
+        stale = [b for b in targets if validations.get(f"feedback:{b.name}") != digest(b.name)]
+        if not stale:
+            return tool_return(
+                self.tool_call_id, "All buffers already reviewed and approved at their current state."
+            )
+        new_stamps: dict[str, str] = {}
+        blocks: list[str] = []
+        for b in stale:
+            verdict = await self._get_feedback(buffer_review_text(buffers, b.name), skipped)
+            blocks.append(f"=== buffer {b.name} ===\nGood? {verdict.good}\nFeedback {verdict.feedback}")
+            if verdict.good:
+                new_stamps[f"feedback:{b.name}"] = digest(b.name)
+        return tool_state_update(self.tool_call_id, "\n\n".join(blocks), validations=new_stamps)
 
 
 _PropertyGenTemplate = TypedTemplate[PropertyGenParams]("property_generation_prompt.j2")
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -730,9 +730,9 @@ _PropertyGenSysTemplate = TypedTemplate[PropertyGenSystemParams]("property_gener
 _SPEC_BUFFERS_GUIDANCE = """
 ## Splitting the spec into verification buffers
 
-Instead of one spec, you may author several named CVL **buffers**, each a self-contained spec that is
-verified and reviewed on its own. Use `put_buffer` / `edit_buffer` / `get_buffer` / `list_buffers` /
-`delete_buffer` instead of the single-spec put/get/edit tools when you split.
+Instead of one spec, you author several named CVL **buffers**, each a self-contained spec verified and
+reviewed on its own. Use `put_buffer` / `edit_buffer` / `get_buffer` / `list_buffers` / `delete_buffer`
+to author them, and `submit_buffer` / `collect_results` (NOT `verify_spec`) to verify them.
 
 When to split:
 - Properties have **conflicting precision needs** — one buffer can summarize a function that another
@@ -741,16 +741,60 @@ When to split:
   is approved once and never re-touched, while you iterate on a small hard buffer in isolation — its
   re-verification and re-review cost only that buffer.
 
-How:
+Structuring buffers:
 - Put **shared** infrastructure (ghosts, common invariants, helper CVL functions, token/oracle models)
   in a buffer with `is_run_target=false`, and have run-target buffers `import "<shared>.spec"` and list
   it in their `imports`.
 - Each **run-target** buffer declares its `property_rules` (the properties it verifies + their rule
   names). Across all run-target buffers, every non-skipped property must appear in **exactly one**
   buffer, and each rule lives in exactly one buffer.
-- `verify_spec` then verifies every run-target buffer independently and in parallel; a buffer already
-  verified at its current content is skipped. Editing a shared buffer re-verifies every buffer that
-  imports it.
+
+Summaries — over-approximate up front (this is your main lever for tractability):
+- For each run-target buffer, look at the functions its properties exercise, and **summarize now**, in
+  that buffer's `methods{}`, every function the buffer's properties do not need exact — an
+  over-approximation (e.g. a `NONDET`/havoc return, or a weaker constraint than the real body).
+  **Do this preemptively, before the first `submit_buffer`** — do not wait for a timeout to force it.
+  Aggressive preemptive summarization is how the hard properties become tractable; a nonlinear-math,
+  hashing, or heavy-external function a property only reads through is the prime candidate.
+- **Soundness rule: an over-approximating summary is sound for `assert` rules but UNSOUND for `satisfy`
+  rules.** An over-approximation *adds* behaviors: an `assert` that holds over the larger behavior set
+  still holds over the real one (sound), whereas a `satisfy` may be witnessed only by an added, non-real
+  behavior (unsound — its witness need not correspond to a real execution). So over-approximate freely
+  in a buffer whose rules are **all `assert`**; in a buffer that contains any `satisfy` rule (or a
+  reachability check), keep the functions it exercises **exact**. This is a first-class reason to
+  partition: group `assert`-only properties (which share aggressive over-approximations) apart from
+  `satisfy` properties (which need those functions exact).
+- **Refine on a spurious counterexample.** A too-coarse over-approximation produces a *spurious* cex —
+  one reachable only because the summary admits behavior the real function cannot. When a buffer returns
+  VIOLATED, analyze the cex: if it hinges on behavior your summary allows but the real function forbids,
+  it is spurious → tighten that summary (add the missing constraint) or drop the over-approximation for
+  that function with `edit_buffer`, then re-`submit_buffer`. If the cex reproduces against the exact
+  function, it is a real bug. Start from the simplest sound over-approximation and tighten only as
+  spurious cexes force you to.
+
+Verifying — submit / collect (buffers prove in parallel; never wait on a slow buffer to work on a fast
+one):
+- **Settle the shared base before submitting anything that imports it.** Submit a run-target buffer only
+  once BOTH (i) that buffer is finished AND (ii) every shared buffer it imports is finished — you do not
+  expect to edit them again. Editing a shared buffer after submitting invalidates every buffer that
+  imports it (submitted or already verified): those prover jobs are wasted and must be re-run. So author
+  and stabilize the shared infrastructure first, then submit the run-target buffers that depend on it.
+- `submit_buffer(name)` starts a buffer's prover job in the **background** and returns immediately.
+  Submit each run-target buffer as soon as it and its shared imports are ready — they prove concurrently.
+- `collect_results()` returns the outcomes of jobs that have **finished so far** (without blocking) plus
+  a status board: which buffers are `complete`, `running`, or `needs (re)submission`. Process a finished
+  buffer immediately — fix its counterexample and `submit_buffer` it again — while the others keep
+  proving.
+- Work this priority order, and only ever block at the last step:
+  1. A finished result not yet processed → handle it. If the fix is in a **shared** buffer (e.g. an
+     under-approximation), edit the shared buffer; that invalidates every buffer importing it — the
+     board lists them under `needs (re)submission` (including ones already verified) — so re-submit
+     each of them.
+  2. Else a buffer still to author/submit → author it and `submit_buffer` it.
+  3. Else everything is submitted and running with nothing finished to process → call
+     `collect_results(wait=true)` to sleep until the next job finishes.
+- Editing a buffer (or a shared buffer it imports) makes its prior verification stale; re-submit it. A
+  buffer already verified at its current content is not re-run.
 
 A single run-target buffer is exactly the one-spec case, so only split when it helps.
 """
@@ -764,6 +808,9 @@ _PROVER_TOOLS = InjectingToolExtension(
 @dataclass
 class ProverTool:
     lg_tool: BaseTool
+    #: The async multi-buffer tools (submit_buffer / collect_results), bound in place of ``lg_tool``
+    #: when spec-buffer authoring is enabled.
+    buffer_tools: list[BaseTool]
     options: ProverOptions
 
 @dataclass
@@ -987,13 +1034,20 @@ async def batch_cvl_generation(
         )
     else:
         b = b.with_tools(env.source_tools)
-    # Multi-buffer authoring tools (opt-in): the agent partitions the spec into several named
-    # buffers verified independently. Off by default so the single-curr_spec flow is untouched.
-    buffer_authoring: list[BaseTool] = [
-        put_buffer(SourceCVLGenerationState), get_buffer(SourceCVLGenerationState),
-        edit_buffer(SourceCVLGenerationState), list_buffers(SourceCVLGenerationState),
-        delete_buffer(SourceCVLGenerationState),
-    ] if spec_buffers_enabled() else []
+    # Multi-buffer authoring (opt-in): the agent partitions the spec into several named buffers,
+    # each submitted and verified independently via the async submit_buffer / collect_results tools
+    # (bound in place of the single-spec verify_spec). Off by default so the single-curr_spec flow is
+    # untouched.
+    if spec_buffers_enabled():
+        buffer_authoring: list[BaseTool] = [
+            put_buffer(SourceCVLGenerationState), get_buffer(SourceCVLGenerationState),
+            edit_buffer(SourceCVLGenerationState), list_buffers(SourceCVLGenerationState),
+            delete_buffer(SourceCVLGenerationState),
+        ]
+        prover_binding: list[BaseTool] = list(prover_tool.buffer_tools)
+    else:
+        buffer_authoring = []
+        prover_binding = [prover_tool.lg_tool]
     task_graph = b.with_tools(
         static_tools()
     ).with_tools(
@@ -1001,7 +1055,7 @@ async def batch_cvl_generation(
     ).with_tools(
         feedback_suite
     ).with_tools(
-        [prover_tool.lg_tool,
+        [*prover_binding,
          ExpectRulePassage.as_tool("expect_rule_passage"),
          ExpectRuleFailure.as_tool("expect_rule_failure"),
          give_up_tool(name="give_up", description=_GIVE_UP_DESCRIPTION, label="CVL generation")
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -990,7 +990,7 @@ async def batch_cvl_generation(
         # Run-root strategy (see ProjectDirectory): an empty working copy is read
         # in-situ, a non-empty one against a temporary materialization whose lifetime
         # is the contributed tool's invocation.
-        project_directory = materializing_project(source.project_root, kit.live.mat)
+        project_directory = materializing_project(kit.live.mat)
 
         @asynccontextmanager
         async def yield_state(
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -18,11 +18,11 @@ from graphcore.graph import tool_state_update, tool_return, RawPromptInput, Cach
 from graphcore.tools.vfs import VFSAccessor, VFSState
 
 from composer.authoring.judge import PropertyFeedbackProtocol
-from composer.authoring.state import SkippedProperty, check_completion
+from composer.authoring.state import SkippedProperty
 from composer.authoring.tools import gated_give_up_tool, give_up_tool
 from composer.spec.cvl_generation import (
     cvl_guidance_tools, skip_tools, CVLGenerationExtra, FEEDBACK_VALIDATION_KEY,
-    validate_property_rules, CVL_JUDGE_KEY, run_cvl_generator,
+    CVL_JUDGE_KEY, run_cvl_generator,
     GeneratedCVL, PropertyRuleMapping, AppliedEdit, FeedbackToolBase, FeedbackServices,
 )
 from composer.prover.core import run_prover, CexHandler, ProverCallbacks, ProverReport
@@ -155,42 +155,32 @@ class PublishResultTool(
     Call to signal your completed cvl generation.
     """
     commentary: str = Field(description="Commentary on your generated spec")
-    property_rules: list[PropertyRuleMapping] = Field(
-        description="The property->rules mapping. For every property you did NOT skip "
-        "(referenced by its unique snake_case title from the batch listing), list the "
-        "name(s) of the rule(s)/invariant(s) in your spec that verify it. Every non-skipped "
-        "property must appear with at least one rule."
-    )
 
     @override
     async def run(self) -> Command | str:
+        # Completion requires every run-target buffer verified AND reviewed at its current digest
+        # (per-buffer stamps), with a clean property/rule partition across buffers. Each run-target
+        # declares its own property->rules on ``put_buffer``, so the published mapping is derived
+        # from the buffers. When every property is skipped there are no run-target buffers: the
+        # stamp check is then vacuous and ``validate_coverage`` alone decides whether publishing is
+        # allowed.
         buffers = self.state.get("buffers") or {}
-        if run_targets(buffers):
-            # Multi-buffer completion: every run-target buffer verified AND reviewed at its current
-            # digest (per-buffer stamps), plus a clean property/rule partition across buffers.
-            skipped_pairs = [(str(s.property_title), str(s.reason)) for s in self.state["skipped"]]
-            if (err := check_buffer_completion(
-                buffers, self.state["validations"], self.state["required_validations"],
-                skipped=skipped_pairs, version_history=self.state["version_history"],
-            )) is not None:
-                return err
-            with self.tool_deps() as titles:
-                skip_titles = {str(s.property_title) for s in self.state["skipped"]}
-                if (err := validate_coverage(buffers, all_properties=set(titles), skipped=skip_titles)) is not None:
-                    return f"Completion REJECTED: {err}"
-            if (err := validate_disjoint_rules(buffers)) is not None:
+        skipped_pairs = [(str(s.property_title), str(s.reason)) for s in self.state["skipped"]]
+        if (err := check_buffer_completion(
+            buffers, self.state["validations"], self.state["required_validations"],
+            skipped=skipped_pairs, version_history=self.state["version_history"],
+        )) is not None:
+            return err
+        with self.tool_deps() as titles:
+            skip_titles = {str(s.property_title) for s in self.state["skipped"]}
+            if (err := validate_coverage(buffers, all_properties=set(titles), skipped=skip_titles)) is not None:
                 return f"Completion REJECTED: {err}"
-            pr = [
-                PropertyRuleMapping(property_title=cast(PropertyTitle, p), rules=cast(list[RuleName], rs))
-                for b in run_targets(buffers) for p, rs in b.property_rules.items()
-            ]
-        else:
-            if (err := check_completion(self.state, self.state["version_history"])) is not None:
-                return err
-            with self.tool_deps() as titles:
-                if (err := validate_property_rules(self.property_rules, self.state["skipped"], titles)) is not None:
-                    return err
-            pr = self.property_rules
+        if (err := validate_disjoint_rules(buffers)) is not None:
+            return f"Completion REJECTED: {err}"
+        pr = [
+            PropertyRuleMapping(property_title=cast(PropertyTitle, p), rules=cast(list[RuleName], rs))
+            for b in run_targets(buffers) for p, rs in b.property_rules.items()
+        ]
         return tool_state_update(
             self.tool_call_id,
             "Accepted",
@@ -761,7 +751,7 @@ _SPEC_BUFFERS_GUIDANCE = """
 
 Instead of one spec, you author several named CVL **buffers**, each a self-contained spec verified and
 reviewed on its own. Use `put_buffer` / `edit_buffer` / `get_buffer` / `list_buffers` / `delete_buffer`
-to author them, and `submit_buffer` / `collect_results` (NOT `verify_spec`) to verify them.
+to author them, and `submit_buffer` / `collect_results` to verify them.
 
 **Strongly prefer splitting, and add over-approximating performance summaries preemptively.** A
 single spec has ONE global `methods{}` block, so every rule is verified under the *intersection* of
@@ -869,9 +859,7 @@ _PROVER_TOOLS = InjectingToolExtension(
 
 @dataclass
 class ProverTool:
-    lg_tool: BaseTool
-    #: The async multi-buffer tools (submit_buffer / collect_results), bound in place of ``lg_tool``
-    #: when spec-buffer authoring is enabled.
+    #: The async multi-buffer tools (submit_buffer / collect_results) the agent verifies with.
     buffer_tools: list[BaseTool]
     options: ProverOptions
 
@@ -999,9 +987,9 @@ async def batch_cvl_generation(
     if editing_tools is not None:
         task_host = TaskHost()
         kit = editing_tools.editing
-        # The same run-root strategy verify_spec uses (see ProjectDirectory): an
-        # empty working copy is read in-situ, a non-empty one against a temporary
-        # materialization whose lifetime is the contributed tool's invocation.
+        # Run-root strategy (see ProjectDirectory): an empty working copy is read
+        # in-situ, a non-empty one against a temporary materialization whose lifetime
+        # is the contributed tool's invocation.
         project_directory = materializing_project(source.project_root, kit.live.mat)
 
         @asynccontextmanager
@@ -1027,7 +1015,7 @@ async def batch_cvl_generation(
             async with project_directory(st.get("vfs") or {}) as run_root:
                 yield CVLAuthorState(
                     working_dir=pathlib.Path(run_root),
-                    curr_spec=st["curr_spec"],
+                    curr_spec=None,
                     prover_runner=WrappedProverRunner(
                         st["config"],
                         prover_tool.options,
@@ -1102,7 +1090,7 @@ async def batch_cvl_generation(
     # Multi-buffer authoring (opt-in): the agent partitions the spec into several named buffers,
     # Multi-buffer authoring is the only mode: the agent writes CVL through the buffer tools and
     # verifies each run-target buffer with the async submit_buffer / collect_results pair. Guidance-only
-    # CVL tools are bound (no put_cvl/edit_cvl/verify_spec).
+    # CVL tools are bound (no put_cvl/edit_cvl).
     buffer_authoring: list[BaseTool] = [
         put_buffer(SourceCVLGenerationState), get_buffer(SourceCVLGenerationState),
         edit_buffer(SourceCVLGenerationState), list_buffers(SourceCVLGenerationState),
@@ -1216,13 +1204,10 @@ async def batch_cvl_generation(
             # unformalizable" judgment — it's the budget talking. Keep the agent's account.
             return Curtailed(None, detail=res_state["result"])
         return GaveUp(reason=res_state["result"])
-    # In multi-buffer mode the artifact is the buffers combined into one document; else curr_spec.
+    # The published artifact is the buffers combined into one document (empty when every property
+    # was skipped, i.e. there are no buffers to combine).
     _buffers = res_state.get("buffers") or {}
-    if run_targets(_buffers):
-        d = combined_buffers_view(_buffers)
-    else:
-        d = res_state["curr_spec"]
-        assert d is not None
+    d = combined_buffers_view(_buffers)
     applied_edits: list[AppliedEdit] = []
     if editing is not None:
         for edit_id in res_state["version_history"]:
--- a/composer/spec/source/author.py
+++ b/composer/spec/source/author.py
@@ -717,13 +717,13 @@ class EditorAwareFeedbackTool(
 
 
 @tool_display("Getting feedback", "Feedback")
+# Buffer-aware feedback for the non-editing author (structural invariants, immutable source): reviews
+# each buffer via the property judge (no source snapshot) and stamps ``feedback:<buffer>``. The
+# agent-facing description is the shared FeedbackToolBase one.
 class BufferPropertyFeedbackTool(
     _BufferReviewFeedback,
     WithAsyncDependencies[Command, FeedbackServices],
 ):
-    """Buffer-aware feedback for the non-editing author (structural invariants, immutable source):
-    reviews each buffer via the property judge (no source snapshot) and stamps ``feedback:<buffer>``."""
-
     __doc__ = FeedbackToolBase.__doc__
 
     @override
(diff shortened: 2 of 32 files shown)


composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+    try:
+        n = int(raw.strip())
+    except ValueError:
+        return DEFAULT_MAX_SPEC_BUFFERS
+    return n if n >= 1 else DEFAULT_MAX_SPEC_BUFFERS
+
+
+class NamedBuffer(BaseModel):
+    """One named CVL spec buffer the agent authors. Frozen and pydantic so it is both the substrate's
+    algorithm type and the shape stored (serializably) in graph state."""
```
jtoman: wut?

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+
+    model_config = {"frozen": True}
+
+    #: Stable identifier.
+    name: str
+    #: The buffer's own CVL text — its rules, its ``methods{}`` block, and its ``import`` statements.
```
jtoman, line 61: I assume it is an invariant that `buffers[nm].name == nm` for all `nm`?

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+    for target in imports_in_cvl(buffers[name].cvl):
+        if (nm := by_path.get(posixpath.normpath(str(here / target)))) is not None:
+            out.append(nm)
+    return tuple(out)
+
+
+def import_closure(buffers: Mapping[str, NamedBuffer], name: str) -> list[NamedBuffer]:
+    """Buffer ``name`` plus every buffer reachable through its imports, transitively — deduped and
+    returned sorted by name. A run-target buffer imports a shared (``is_run_target=false``) buffer only
+    when it does, so a shared buffer belongs to the closure (and invalidation set) of exactly the
```
jtoman: "a run target buffer imports a shared buffer only when it does"

????? what?

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+    by_path = {posixpath.normpath(b.path): nm for nm, b in buffers.items()}
+    here = PurePosixPath(buffers[name].path).parent
+    out: list[str] = []
+    for target in imports_in_cvl(buffers[name].cvl):
+        if (nm := by_path.get(posixpath.normpath(str(here / target)))) is not None:
+            out.append(nm)
+    return tuple(out)
+
+
+def import_closure(buffers: Mapping[str, NamedBuffer], name: str) -> list[NamedBuffer]:
```
jtoman, line 137: is there a reason we don't have the prover create this inventory for us? The answer might be "because it doesn't produce the output we want, and this needs to land independently of that". I buy that answer, to be clear, but I want to make sure I understand the reason.

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def buffer_digest(
+    buffers: Mapping[str, NamedBuffer], name: str, *, extra_parts: Sequence[str] = ()
+) -> str:
+    """A content digest of buffer ``name`` and its transitive import closure, plus any ``extra_parts``
+    (e.g. skipped-property or conf-flag markers). Editing the buffer OR any buffer it imports changes
+    the digest, so it keys the buffer's cached verify/review. Mirrors
+    :meth:`ContentCache.compute_cache_key` (content-keyed, order-independent)."""
+    # TODO: a pure-comment edit (e.g. reframing a property's justification docstring) changes this
```
jtoman, line 162: WHAT IF WE USED THE CANONICAL TAC

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+    buffers: Mapping[str, NamedBuffer], *, all_properties: set[str], skipped: set[str]
+) -> str | None:
+    """Whether the run-target buffers cover the property space exactly once — the publish-time contract:
+    every non-skipped property assigned to exactly one buffer, and no unknown or skipped property
+    assigned. Returns None when valid, else one message enumerating every problem."""
+    assigned: list[str] = [p for b in run_targets(buffers) for p in b.properties]
+    seen: set[str] = set()
+    duplicated: set[str] = set()
+    for p in assigned:
+        (duplicated if p in seen else seen).add(p)
```
jtoman, line 277: cute

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+    naming the shared rules."""
+    seen: set[str] = set()
+    dup: set[str] = set()
+    for b in run_targets(buffers):
+        for r in b.owned_rules:
+            (dup if r in seen else seen).add(r)
+    return f"rules owned by more than one buffer: {sorted(dup)}" if dup else None
+
+
+_METHODS_BLOCK = re.compile(r"methods\s*\{([^}]*)\}", re.DOTALL)
```
jtoman, line 303: something something pumping lemma

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,320 @@
...
+    return " ".join(s.split())
+
+
+def _extract_decls(cvl: str) -> set[str]:
+    """The `methods{}` entries and simple `ghost` declarations in ``cvl``, normalized. A light-regex
+    heuristic: ``methods`` entries are split on ``;`` within each block; ghost decls are single-statement
+    (axiom-block ghosts and multi-line CVL function *bodies* are not extracted). Good enough to flag
+    verbatim copy-paste across buffers, not a CVL parser."""
+    decls: set[str] = set()
+    for block in _METHODS_BLOCK.findall(cvl):
```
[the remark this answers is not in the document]
jtoman: this is fine I guess.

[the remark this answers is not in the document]
jtoman: to be clear, this will cause a false positive for two instances of `function _.transfer() => cvlTransfer();` where the definition of `cvlTransfer()` is different. I'm not claiming this is likely to happen, but just making sure I understand the limitations.

composer/spec/source/prover.py
```diff
@@ -496,197 +505,415 @@ def setup_prover_config_in(
...
 
-            curr_state_digest = spec_digest(
-                spec, state["skipped"], state["version_history"]
-            )
+    submit_schema = create_model(
+        "SubmitBuffer", __doc__=_SUBMIT_BUFFER_DESCRIPTION,
+        name=(str, Field(description="The run-target buffer to submit for verification.")),
+        state=(Annotated[StateWithSkips, InjectedState], ...),
+        tool_call_id=(Annotated[str, InjectedToolCallId], ...),
+    )
```
jtoman: I'm confused, why are these schemas dynamic? There's nothing dynamic about then?

composer/spec/source/prover.py
```diff
@@ -496,197 +505,415 @@ def setup_prover_config_in(
...
-                    tool_call_id=tool_call_id, content=result.result_str,
-                    prover_link=result.link, validations=stamper(state, state["version_history"]),
-                    prover_history=prover_update, **nag_channel
-                )
-            return tool_state_update(
-                tool_call_id=tool_call_id, content=result.result_str, prover_link=result.link,
-                prover_history=prover_update, **nag_channel
+    @tool_display("Collecting prover results", None)
+    @tool(args_schema=collect_schema)
+    async def collect_results(**args) -> str | Command:
```
jtoman: so I am fairly certain this is catastrophically wrong. The prover tool right now is stateless: all of its deps are "static" and agnostic to the agent using the tool or they are passed in as tool call args. The "in flight" buffers are shared state, and there is absolutely nothing here that remembers which author submits which job. So if I have an author working on component A, and a separate author working on component B, and they both call `collect` at the same time, whoever gets lucky and woken up (call it component B) on submit will drain the *entire* queue. below, when looping over `drained`, it will just splat component A's result in, and then happily continue. This will not only be very confusing for B, but it can deadlock A. If `B` is now done (and decides to ignore A's spurious results), A is now waiting for a queue to be populated by a prover run result that is never coming back (it already finished and its result was routed to B).

NB I don't think you are saved by insisting the buffer names are unique across all authors (not that that is an invariant as far as I can tell anyway); the problem comes from having buffers being consumed unconditionally.

The fix is to either push these to dependent tools, which require you to provide a task "scope" for the each tool (so different authors get different task sets), or only drain known buffer names AND insist buffer names are unique across component workers. The latter seems pretty brittle tbh, and I would avoid it.

composer/spec/source/prover.py
```diff
@@ -496,197 +505,415 @@ def setup_prover_config_in(
...
+        needs_submit = [b.name for b in targets if b.name not in complete and b.name not in running]
+        board = [
+            "",
+            f"[buffers] complete: {sorted(complete)}",
+            f"[buffers] running: {sorted(running)}",
+            f"[buffers] needs (re)submission: {sorted(needs_submit)}",
+        ]
+        # Advisory: flag declarations duplicated verbatim across run-target buffers (once each) — likely
+        # belong in a shared buffer the duplicating buffers import.
+        fresh_dupes = {d: ns for d, ns in duplicated_declarations(buffers).items() if d not in reported_dupes}
```
jtoman: similar here, a reported dupe for some random component suppresses dupe flagging for every other component.

composer/spec/source/prover.py
```diff
@@ -496,197 +505,415 @@ def setup_prover_config_in(
...
+    names = list(buffers)
+    with ExitStack() as stack:
+        yield {
+            name: stack.enter_context(tmp_spec(
+                root=working_dir,
+                content=_retarget_buffer_imports(buf.cvl, names, tag),
+                name=f"{name}__{tag}",
+            ))
+            for name, buf in buffers.items()
+        }
```
jtoman, line 582: cut the gordian knot: always materialize the folder for every run. No race, no problem, no import surgery via string ops.

composer/spec/source/author.py
```diff
@@ -147,25 +155,37 @@ class PublishResultTool(
...
         with self.tool_deps() as titles:
-            if (err := validate_property_rules(self.property_rules, self.state["skipped"], titles)) is not None:
-                return err
+            skip_titles = {str(s.property_title) for s in self.state["skipped"]}
+            if (err := validate_coverage(buffers, all_properties=set(titles), skipped=skip_titles)) is not None:
+                return f"Completion REJECTED: {err}"
+        if (err := validate_disjoint_rules(buffers)) is not None:
+            return f"Completion REJECTED: {err}"
+        pr = [
+            PropertyRuleMapping(property_title=cast(PropertyTitle, p), rules=cast(list[RuleName], rs))
```
jtoman: instead of casting, just do `PropertyTitle(p)` and `[RuleName(rn) for rn in rs]`. Technically what you have is fine because these types are just typechecking fictions, but the minute they don't we get a mysterious type error. The casts are cheap.

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
 
+@tool_display("Getting feedback", "Feedback")
+class BufferPropertyFeedbackTool(
+    _BufferReviewFeedback,
+    WithAsyncDependencies[Command, FeedbackServices],
+):
+    """Buffer-aware feedback for the non-editing author (structural invariants, immutable source):
+    reviews each buffer via the property judge (no source snapshot) and stamps ``feedback:<buffer>``."""
+
+    __doc__ = FeedbackToolBase.__doc__
```
jtoman: love writing a doc comment and then immediately overwriting it

composer/spec/source/author.py
```diff
@@ -632,9 +652,49 @@ def judge_tools(self) -> tuple[BaseTool, ...]:
...
+    editor-aware (source editing) and property-only (structural invariants / immutable source) — build
+    on it, so per-buffer review works whether or not source editing is enabled. Subclasses supply
+    ``_get_feedback`` (how the judge is reached)."""
+
+    @override
+    async def run(self) -> Command:
+        buffers = self.state.get("buffers") or {}
+        targets = run_targets(buffers)
+        if not targets:
+            return await super().run()
```
jtoman: with shelly yeeting structural invariants, "no editing" as a consumer of all of this just doesn't exist anymore. Should simplify this...

composer/spec/source/author.py
```diff
@@ -632,9 +652,49 @@ def judge_tools(self) -> tuple[BaseTool, ...]:
...
+
+        stale = [b for b in targets if validations.get(f"feedback:{b.name}") != digest(b.name)]
+        if not stale:
+            return tool_return(
+                self.tool_call_id, "All buffers already reviewed and approved at their current state."
+            )
+        new_stamps: dict[str, str] = {}
+        blocks: list[str] = []
+        for b in stale:
+            verdict = await self._get_feedback(buffer_review_text(buffers, b.name), skipped)
```
jtoman: I'm... confused. How does the feedback agent know which properties a buffer purports to verify? Unless you've changed it (it doesn't look like you did) the feedback agent gets a giant master list of properties, but it has no way to evaluate whether the buffer actually formalizes the properties it claims. If `NamedBuffer` says "I formalize `Some important solvency property`" but never mentions that anywhere, the feedback agent has no way to judge "you're not formalizing that faithfully" because it doesn't know what is claiming. Right? what am I missing

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+
+
 _PropertyGenTemplate = TypedTemplate[PropertyGenParams]("property_generation_prompt.j2")
 
 class PropertyGenSystemParams(TypedDict):
     source_editing: bool
 
 _PropertyGenSysTemplate = TypedTemplate[PropertyGenSystemParams]("property_generation_system_prompt.j2")
 
+#: Appended to the CVL-generation system prompt: how to author and verify the spec as buffers.
```
jtoman: This is long enough and substantive enough it should be a `j2` file. Conditionally include via a new flag in PropertyGenSysParams

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+
+When to split:
+- Properties have **conflicting precision needs** — one buffer can summarize a function that another
+  keeps exact (each buffer has its own `methods{}`), with no global-methods-block conflict. This is
+  the primary axis: group the properties by the set of functions they genuinely need exact.
+- Separate **`assert`-only** properties (which can share aggressive over-approximations) from
+  **`satisfy`** properties (which need those functions exact) — see the soundness rule below.
+- To **isolate the hard/slow properties**: put the many easy properties in one buffer that verifies and
+  is approved once and never re-touched, while you iterate on a small hard buffer in isolation — its
+  re-verification and re-review cost only that buffer.
```
jtoman: "cost only that buffer" gestures at, but doesn't quite make explicit an important rule: buffers are stamped independently, so a spec that verifies good and gets stamped can be ignored for the rest of the run.

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+
+Structuring buffers:
+- Put infrastructure shared by two or more run-target buffers (ghosts, common invariants, helper CVL
+  functions, token/oracle models, and summaries several groups need) in a buffer with
+  `is_run_target=false`, and `import "<shared>.spec"` it from **each run-target that uses it** (and list
+  it in that buffer's `imports`). A shared buffer is pulled in only by the buffers that import it, so
+  editing it re-verifies only those — you can make one shared buffer for a subset of groups and another
+  for a different subset. **Import the shared buffer where you need it; never restate its contents in a
```
jtoman: This seems obvious? But agents consistently do obviously stupid stuff.

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+  Aggressive preemptive summarization is how the hard properties become tractable; a nonlinear-math,
+  hashing, or heavy-external function a property only reads through is the prime candidate.
+- **Soundness rule: an over-approximating summary is sound for `assert` rules but UNSOUND for `satisfy`
+  rules.** An over-approximation *adds* behaviors: an `assert` that holds over the larger behavior set
+  still holds over the real one (sound), whereas a `satisfy` may be witnessed only by an added, non-real
+  behavior (unsound — its witness need not correspond to a real execution). So over-approximate freely
+  in a buffer whose rules are **all `assert`**; in a buffer that contains any `satisfy` rule (or a
+  reachability check), keep the functions it exercises **exact**. This is a first-class reason to
+  partition: group `assert`-only properties (which share aggressive over-approximations) apart from
+  `satisfy` properties (which need those functions exact).
```
jtoman: OOOOOH. Yikes. You need to be very careful with this. Agents have historically used this as a license to NONDET away important functions with important side effects. I wrote a whole ass CVL RAG document explaining exactly when NONDET summaries are sound vs not (answer, basically never, except for view functions). I would significantly expand on this section, or just point the agent at the extensive CVL RAG documentation about how to summarize responsibly.

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+- **Put each summary where it is USED.** A summary several run-target buffers need goes in a shared
+  buffer they each `import`; a summary only one buffer needs goes in *that* buffer's own `methods{}` (a
+  run-target buffer has its own `methods{}` too). A single-buffer summary parked in a shared buffer that
```
jtoman: "may have its own"

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+      rule approvals_are_recorded { ... }
+
+  so editing this summary never invalidates the other buffers.
+- **Refine on a spurious counterexample.** A too-coarse over-approximation produces a *spurious* cex —
+  one reachable only because the summary admits behavior the real function cannot. When a buffer returns
+  VIOLATED, analyze the cex: if it hinges on behavior your summary allows but the real function forbids,
+  it is spurious → tighten that summary (add the missing constraint) or drop the over-approximation for
+  that function with `edit_buffer`, then re-`submit_buffer`. If the cex reproduces against the exact
+  function, it is a real bug. Start from the simplest sound over-approximation and tighten only as
+  spurious cexes force you to.
```
jtoman: again, this simplifies away a lot of the detail in the CVL RAG document, just point at it imo

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+
+Verifying — submit / collect (buffers prove in parallel; never wait on a slow buffer to work on a fast
+one):
+- **Settle the shared base before submitting anything that imports it.** Submit a run-target buffer only
+  once BOTH (i) that buffer is finished AND (ii) every shared buffer it imports is finished — you do not
```
jtoman: what does it mean for a run-target buffer to be "finished"? Like, finished editing?

composer/spec/source/author.py
```diff
@@ -656,13 +716,141 @@ def _version_history(self) -> Sequence[str]:
...
+  Submit each run-target buffer as soon as it and its shared imports are ready — they prove concurrently.
+- `collect_results()` returns the outcomes of jobs that have **finished so far** (without blocking) plus
+  a status board: which buffers are `complete`, `running`, or `needs (re)submission`. Process a finished
+  buffer immediately — fix its counterexample and `submit_buffer` it again — while the others keep
+  proving.
+- Work this priority order, and only ever block at the last step:
+  1. A finished result not yet processed → handle it. If the fix is in a **shared** buffer (e.g. an
+     under-approximation), edit the shared buffer; that invalidates every buffer importing it — the
+     board lists them under `needs (re)submission` (including ones already verified) — so re-submit
+     each of them.
```
jtoman: er, no it doesn't. This makes it sound like the board tells the agent what jobs need to be resubmitted based on the fix required for a shared spec fix. But it has no way of knowing a CEX remediation requires a shared spec fix?

composer/spec/source/author.py
```diff
@@ -823,7 +1015,7 @@ async def propose(
             async with project_directory(st.get("vfs") or {}) as run_root:
                 yield CVLAuthorState(
                     working_dir=pathlib.Path(run_root),
-                    curr_spec=st["curr_spec"],
+                    curr_spec=None,
                     prover_runner=WrappedProverRunner(
```
jtoman, line 978: UH. Wait, hang on. This breaks the plugin API in a **BIG** way that will break some other code. let's discuss

composer/spec/source/buffer_tools.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+
+
+def put_buffer[S: WithBuffers](ty: type[S]) -> BaseTool:
+    schema = create_model(
+        "PutBuffer", __base__=_PutBufferTemplate, __doc__=put_buffer_description,
+        state=(Annotated[ty, InjectedState], ...),
+        tool_call_id=(Annotated[str, InjectedToolCallId], ...),
+    )
+
```
jtoman: turns out you can subscript generic pydantic models with concrete types just fine, so this create model template dance is totally unnecessary.

composer/spec/source/buffer_tools.py
```diff
@@ -0,0 +1,196 @@
...
+    @tool_display_of(_put_display)
+    @tool(args_schema=schema)
+    def put_buffer(**args) -> str | Command:
+        if (err := cvl_syntax_error(args["cvl"])) is not None:
+            return err
+        buffers_now = args["state"].get("buffers") or {}
```
jtoman: I guess ... what's the point of the argument schema if you're gonna throw it all away at the argument typing level? But with the above understanding, just use the "WithImplementation" and what not to avoid all this boilerplate.

jtoman (changes requested):

composer/spec/source/spec_buffers.py
jar-ben: Exactly. I was checking it, the prover does not cheaply produce it. I am a bit worried about this too, i.e. I will add the functionality to the prover (but the integration is non-trivial (needing prover version release and bumping pointers at various places).

composer/spec/source/spec_buffers.py
jar-ben: Well, you know the tac is not really deterministic, and then there is also the "how to obtain it" question. The bigger problem perhaps is that we want to, at the end, provide the user with the final artifacts (including right cvl comments). 

This is a pre-existing issue. I saw in a run of AutoProver that 5-10 % of his prover runs were just due to CVL comment edits. I will look into this later (perhaps I was just unlucky with that run).

composer/spec/source/author.py
jar-ben: So, `CVLAuthorState` now exposes the buffer set directly (`buffers: Mapping[str, NamedBuffer]`) plus `spec_for_rule(rule) -> str | None`, which returns the CVL for the buffer that declares that rule (its text + transitive imports as one document, i.e. not .spec file). Is this sufficient?

composer/spec/source/prover.py
jar-ben: done

composer/spec/source/spec_buffers.py
jtoman: I was just kidding in any event

composer/spec/source/author.py
```diff
@@ -695,47 +752,55 @@ def digest(name: str) -> str:
...
+        with self.tool_deps() as judges:
             assert "vfs" in self.state
             snap = SourceSnapshot(
                 vfs=self.state["vfs"],
                 version_history=self.state["version_history"],
             )
-            return await judge(snap, spec, skipped, self.rebuttals, self.tool_call_id)
+            return await judges.for_buffer(name, properties)(
+                snap, spec, skipped, self.rebuttals, self.tool_call_id
+            )
```
jtoman, line 766: doesn't this mean every rebuttal goes to every feedback agent? even the ones not being rebutted?

composer/spec/source/author.py
```diff
@@ -1055,18 +1014,28 @@ async def propose(
...
     else:
-        judge_impl = source_feedback_judge(
-            judge_ctx, _LiveJudgeHost(env, editing), judge_prompt, props
+        judge_host = _LiveJudgeHost(env, editing)
+        source_judges = _PerBufferJudge(
+            build=lambda name, claimed: source_feedback_judge(
+                judge_ctx.child(CacheKey[CVLJudge, CVLJudge](name)), judge_host, judge_prompt, claimed
+            ),
+            properties=props,
         )
```
jtoman: given that we've dropped the structural invariant phase, which was the only user of this code that kept its source immutable, I don't see a particularly compelling reason to keep supporting a feature no one is using. I suspect it will make the footprint of this change much, much nicer; no more different mode juggling?

composer/spec/source/buffer_tools.py
```diff
@@ -61,108 +72,89 @@ class _PutBufferTemplate(BaseModel):
...
             if sum(1 for b in buffers_now.values() if b.is_run_target) >= cap:
                 return (
-                    f"Refusing to create run-target buffer {args['name']!r}: the run-target buffer cap "
+                    f"Refusing to create run-target buffer {self.name!r}: the run-target buffer cap "
                     f"({cap}) is already reached. Fold these properties into an existing run-target "
                     f"buffer instead."
                 )
-        existing = buffers_now.get(args["name"])
+        existing = buffers_now.get(self.name)
         # Keep the prior property->rule mapping when the agent re-puts text without restating it.
```
jtoman, line 92: is this mentioned in the tool documentation? I didn't see it.

composer/spec/source/buffer_tools.py
```diff
@@ -61,108 +72,89 @@ class _PutBufferTemplate(BaseModel):
...
 
 
-edit_buffer_description = """
-Make a surgical edit to one spec buffer instead of re-emitting it. Provide `name`, an exact
-`old_string` span copied from that buffer (must occur exactly once — include context to disambiguate),
-and `new_string`. The edited buffer is re-parsed; if it fails to parse the edit is rejected and the
-buffer is unchanged. Dramatically cheaper than `put_buffer` for a small change.
-"""
+def put_buffer[S: WithBuffers](ty: type[S]) -> BaseTool:
+    return tool_display_of(_put_display)(PutBuffer[ty].as_tool("put_buffer"))
```
jtoman, line 106: does the annotation form not work? how strange!

composer/spec/source/prover.py
```diff
@@ -281,6 +308,35 @@ def buffer_is_complete(
...
+    prover_history: list[ProverHistoryItem],
+    buffers: Mapping[str, NamedBuffer],
+    *,
+    skipped: Sequence[tuple[str, str]],
+    version_history: Sequence[str],
+) -> list[str]:
+    """The prover-run links whose completed results compose the run-target buffers at their current
+    digests — the same runs :func:`buffer_is_complete` considers. Rule-striping runs one buffer's rules
+    across several jobs, so its verdicts are spread over all of these links rather than carried by the
+    last one alone."""
```
jtoman, line 335: ...of course. oops

composer/spec/source/prover.py
```diff
@@ -540,33 +600,17 @@ def stuck_rule_nag(
...
-    ``import "<sibling>.spec"`` — retargeted to the tagged name — resolves), and yield ``name -> on-disk
-    spec path``; every file is removed on exit. The ``tag`` (unique per submission) isolates concurrent
-    jobs that share one run-root: on the in-situ path the run-root is the shared project directory, so
-    two jobs writing the deterministic ``{name}.spec`` would clobber each other and race on cleanup."""
-    names = list(buffers)
+    """Write every buffer as ``{name}.spec`` into the specs dir (all at once, so any buffer's
+    ``import "<sibling>.spec"`` resolves to its sibling), and yield ``name -> on-disk spec path``;
+    every file is removed on exit. The run owns its materialized project folder, so the deterministic
+    filenames never collide with a concurrent job's."""
     with ExitStack() as stack:
```
jtoman: rest in piss lol

composer/spec/source/prover.py
```diff
@@ -637,6 +712,9 @@ class _BufJob:
     #: so a job whose digest is now stale (its buffer or a shared import changed) can't mark it done.
     digest: str
     task: "asyncio.Task[None]"
```
jtoman: unclear why this is quoted by okay

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,110 @@
+## Splitting the spec into verification buffers
+
+Instead of one spec, you author several named CVL **buffers**, each a self-contained spec verified and
+reviewed on its own. Use `put_buffer` / `edit_buffer` / `get_buffer` / `list_buffers` / `delete_buffer`
+to author them, and `submit_buffer` / `collect_results` to verify them.
+
+**Strongly prefer splitting, and add over-approximating performance summaries preemptively.** A
+single spec has ONE global `methods{}` block, so every rule is verified under the *intersection* of
+what all rules need precise — one property that needs an expensive function kept exact forces EVERY
+rule to pay that cost, which is the usual source of a timeout. Splitting breaks that coupling: each
```
jtoman: I think you can trim this to be honest. You don't need to argue why splitting is expensive, fine to just say what splitting buys you.

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,110 @@
...
+- **Put each summary where it is USED.** A summary several run-target buffers need goes in a shared
+  buffer they each `import`; a summary only one buffer needs goes in *that* buffer's own `methods{}`. A
+  single-buffer summary parked in a shared buffer that others import makes every later edit re-verify all
+  of them — wasted work on buffers that never use it.
+- **Refine on a spurious counterexample.** A too-coarse over-approximation produces a *spurious* cex —
+  one reachable only because the summary admits behavior the real function cannot. When a buffer returns
+  VIOLATED, analyze the cex: if it hinges on behavior your summary allows but the real function forbids,
+  it is spurious → tighten that summary (add the missing constraint) or drop the over-approximation for
+  that function with `edit_buffer`, then re-`submit_buffer`. Start from the
+  simplest sound over-approximation and tighten only as spurious cexes force you to.
```
jtoman: there is a section in the context doc you can ref here I think

composer/spec/source/author.py
```diff
@@ -685,7 +739,10 @@ def digest(name: str) -> str:
         new_stamps: dict[str, str] = {}
         blocks: list[str] = []
         for b in stale:
-            verdict = await self._get_feedback(buffer_review_text(buffers, b.name), skipped)
+            claimed = [p for p in all_props if str(p.title) in b.property_rules]
+            verdict = await self._review(
+                b.name, buffer_review_text(buffers, b.name), skipped, claimed
+            )
             blocks.append(f"=== buffer {b.name} ===\nGood? {verdict.good}\nFeedback {verdict.feedback}")
```
jtoman, line 729: does this mean skipped is evaluated against every buffer? and changing a skip will invalidate all buffers? I'm not mad about this, just making sure I understand.

jar-ben: right - `skipped` was part of every buffer's digest, so changing a skip re-reviewed all buffers. Fixed in `d08fce25`: `skipped` is no longer in the per-buffer feedback digest. A skip belongs to no single buffer, so it's reviewed once as a standalone unit (`skips_review_digest` / `SKIPS_VALIDATION_KEY`). So now a skip change only re-triggers that single skip review, not every buffer.

composer/spec/source/author.py
jar-ben: right, that was the case at `b06cbe40`. Fixed in `d08fce25` - rebuttals are keyed by buffer now, so each unit's judge sees only the rebuttals filed against its own buffer:

```python
rebuttals = [r for r in self.rebuttals if r.buffer == name]   # in _review
```

composer/spec/source/author.py
jar-ben: agreed, done. Dropped the immutable-source path entirely - `BufferPropertyFeedbackTool` and the `if editing is None: ... else:` juggling are gone, there is just the single editing tool (`EditorAwareFeedbackTool`, `source_editing=True`) now.

composer/spec/source/buffer_tools.py
jar-ben: added it to the `put_buffer` description: *"The number of run-target buffers is capped; creating one past the cap is refused - fold those properties into an existing run-target buffer instead."*

composer/spec/source/buffer_tools.py
jar-ben: it works - switched all the buffer tools to the `WithImplementation` / `WithInjectedState[S]` idiom (`77eedf55`), so it's just `PutBuffer[ty].as_tool("put_buffer")` now, the `create_model` dance is gone.

composer/spec/source/prover.py
jar-ben: btw this got reworked in the master merge - the report reads these via `run_links` + `_fetch_covering` (newest-run-wins) now.

composer/spec/source/prover.py
jar-ben: needless quote, unquoted now (`task: asyncio.Task[None]`).

composer/templates/spec_buffers_guidance.j2
jar-ben: trimmed - the intro leads with what splitting buys you now, the expense is just a parenthetical. Also re-cut it to summarize the prover hotspots (nonlinear / bitwise / hashing / assembly), i.e. not every function.

composer/templates/spec_buffers_guidance.j2
jar-ben: done - the summaries section now points at the **"CVL Summarization — Knowledge Base"** context document (incl. its section B3 for the unresolved-call HAVOC ladder) instead of re-deriving the rules here. Kept only the assert-vs-`satisfy` soundness rule inline.

composer/spec/source/buffer_tools.py
jtoman: no, I meant the "passing an empty dict keeps whatever you declared before".

composer/spec/source/buffer_tools.py
jtoman: uh, that's not what I meant, I was talking about the `@tool_display_of(_put_display)` annotation directly on `PutBuffer`. Whatever, this is fine.

composer/spec/source/prover.py
```diff
@@ -588,193 +638,441 @@ def setup_prover_config_in(
...
+            if not targets:
+                return "No run-target buffers to collect. Author buffers and submit_buffer them first."
+
+            # Current digest per buffer is stable within this call (buffers/skips/edit-history are fixed);
+            # memoize it — each is a content hash over the import closure, read at several points below.
+            _digests: dict[str, str] = {}
+            def cur_digest(nm: str) -> str:
+                if nm not in _digests:
+                    _digests[nm] = _cur_digest(state, buffers, nm)
+                return _digests[nm]
```
jtoman, line 986: pretty sure `@functools.lru_cache` or something does exactly this, and works fine for local functions (probably)

composer/spec/source/prover.py
```diff
@@ -588,193 +638,441 @@ def setup_prover_config_in(
...
+
+            drained: list[_BufDone] = []
+            while not done_queue.empty():
+                drained.append(done_queue.get_nowait())
+            if not drained and wait and any(not j.task.done() for j in buffer_jobs.values()):
+                # Idle wait: nothing else to do, sleep until one job finishes. Unbounded, but every job is
+                # guaranteed to land on the queue — run_prover self-bounds its subprocess, and a crash is
+                # caught and enqueued as an error — so this can't hang on a wedged job.
+                drained.append(await done_queue.get())
```
jtoman, line 995: um, it can make it hard to find a wrap up! I'd put a timeout here and explicitly check the "wrap it up" pressure guard.

composer/spec/source/prover.py
```diff
@@ -588,193 +638,441 @@ def setup_prover_config_in(
...
+                while not done_queue.empty():
+                    drained.append(done_queue.get_nowait())
+
+            # Cancel jobs left running against a now-stale digest: a shared buffer they import was edited, so
+            # their result would be discarded anyway — and on local runs a doomed job needlessly holds the
+            # single prover slot. The agent re-submits them (they show under needs-(re)submission below).
+            for key, j in list(buffer_jobs.items()):
+                if not j.task.done() and j.name in buffers and j.digest != cur_digest(j.name):
```
jtoman, line 1003: shouldn't you also cancel jobs that name buffers that no longer exist?

composer/spec/source/prover.py
```diff
@@ -588,193 +638,441 @@ def setup_prover_config_in(
...
+                    j.task.cancel()
+                    buffer_jobs.pop(key, None)
+
+            # Retire finished jobs from the registry. A result that lands between the drain and here stays on
+            # the queue, so its buffer is picked up on the next collect even though its job is already gone.
+            for key in [key for key, j in buffer_jobs.items() if j.task.done()]:
```
jtoman, line 1009: to clarify, this is fine because `j.task.done()` returns true only after `_run_buffer_job` returns, at which point the results are guaranteed on the queue?

composer/spec/source/prover.py
```diff
@@ -588,193 +638,441 @@ def setup_prover_config_in(
...
+            fresh: dict[str, list[tuple[RulePath, StatusCodes]]] = {}
+            link: str | None = None
+            parts: list[str] = []
+            for d in drained:
+                if isinstance(d.result, str):  # compile/toolchain error: surface it, record no run
+                    parts.append(f"=== buffer {d.name} ===\n{d.result}")
+                    continue
+                results: list[tuple[RulePath, StatusCodes]] = list(d.result.raw_rule_status.items())
+                fresh.setdefault(d.name, []).extend(results)  # striped subsets of one buffer accumulate
+                link = d.result.link or link
```
jtoman, line 1022: ?? Why is it okay to use the run link from the last iteration of the loop the prover run record record of some other, completely unrelated task?

composer/spec/source/prover.py
```diff
@@ -588,193 +638,441 @@ def setup_prover_config_in(
...
-                # every declared rule was verified against exactly this authoring state
-                # (the state_digest match), so a piecemeal completion is as good as a
-                # full-run one.
-                return tool_state_update(
-                    tool_call_id=tool_call_id, content=result.result_str,
-                    prover_link=result.link, validations=stamper(state, state["version_history"]),
-                    prover_history=prover_update, **nag_channel
+                    "Every run-target buffer is verified at its current content. Once each also has "
+                    "feedback, you can publish."
                 )
```
jtoman, line 1069: if I verify all the rules in buffer `foo.spec` and haven't event started trying to run the rules in `bar.spec`, will this claim of "every run-target buffer is verified" trigger? It's probably worth just saying "buffers B1, B2, ... are fully verified at their current content" where Bi is taken from `complete`

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+class NamedBuffer(BaseModel):
+    """One named CVL spec buffer the agent authors. A frozen pydantic model, so the same type is both
+    what the buffer logic operates on and what is stored (serializably) in graph state."""
+
+    model_config = {"frozen": True}
+
+    #: Stable identifier, and the key this buffer is stored under: ``buffers[nm].name == nm`` holds.
+    name: str
```
jtoman, line 61: probably worth documenting this name needs to be a valid looking path component.

composer/spec/source/spec_buffers.py
jtoman: wait, don't we have a way to dump the AST of CVL? It would beat the handrolled imports_in_cvl parsing, but whatever

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,456 @@
...
+#: The parsed-AST command tag for a ``requireInvariant`` statement (an "assume invariant" command).
+_ASSUME_INVARIANT_CMD = "AssumeCmd.AssumeInvariant"
+
+
+def requireinvariant_citations(cvl: str) -> set[str]:
+    """The invariant names cited with ``requireInvariant`` in ``cvl``, read from the parsed CVL AST
+    (``ASTExtraction.jar`` via :func:`extract_cvl_ast`) rather than by scanning text. Each citation is
+    an ``AssumeInvariant`` command whose ``id`` is the invariant name. Empty when the source has no
+    AST (a hard syntax error — the buffer is rejected at the parse gate anyway)."""
+    ast = extract_cvl_ast(cvl)
```
jtoman, line 365: wait, you're reading the AST here!! Why are you using the jankass `imports_in_cvl` then lol

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,456 @@
...
+                walk(v)
+        elif isinstance(node, list):
+            for v in node:
+                walk(v)
+
+    walk(ast)
+    return cited
+
+
+# TODO(future): support shared invariants. Today an invariant must be declared (and so proved) in the
```
jtoman, line 386: "EVERY TODO COMMENT MUST CITE A JIRA TICK ---"

he shouts as he is dragged off to a padded room.

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,456 @@
...
+# resolution here plus a coverage model for an invariant-only (property-less) run-target buffer.
+def validate_requireinvariant_proved(
+    buffers: Mapping[str, NamedBuffer],
+    cited_by_buffer: Mapping[str, set[str]],
+    declared_by_buffer: Mapping[str, set[str] | None],
+) -> str | None:
+    """Every invariant a run-target buffer cites with ``requireInvariant`` must be declared in THAT
+    buffer, so it is proved in the same run under the same summaries. An invariant declared only in
+    another buffer — including an unproven shared buffer it imports — is not re-verified in this run, so
+    citing it would assume it unproven (unsound). Refuses, naming the offending citations.
```
jtoman, line 398: shouldn't we also check that the cited name isn't in `expected_to_fail`?

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+# TODO(future): support shared invariants. Today an invariant must be declared (and so proved) in the
+# buffer that cites it, forcing duplication when several run-targets need the same one. A follow-up will
+# let a shared buffer prove an invariant once and have importers cite it — which needs cross-buffer
+# resolution here plus a coverage model for an invariant-only (property-less) run-target buffer.
+def validate_requireinvariant_proved(
+    buffers: Mapping[str, NamedBuffer],
+    cited_by_buffer: Mapping[str, set[str]],
+    declared_by_buffer: Mapping[str, set[str] | None],
```
jtoman, line 393: worth using the `RuleName` phantom types here so we know that the codomain is a set of RuleNames.

composer/spec/source/spec_buffers.py
```diff
@@ -0,0 +1,456 @@
...
+    run-target buffer — copy-paste that likely belongs in a shared buffer the duplicating buffers import.
+    Returns ``{declaration: sorted buffer names}`` for each duplicated declaration (heuristic, text-based;
+    see :func:`_extract_decls`). Because it compares only the declaration text, it can false-positive:
+    two identical summary entries like ``function _.transfer() => cvlTransfer();`` are flagged even when
+    the CVL function ``cvlTransfer()`` they point at is defined differently in each buffer."""
+    where: dict[str, list[str]] = {}
+    for b in run_targets(buffers):
+        for decl in _extract_decls(b.cvl):
+            where.setdefault(decl, []).append(b.name)
+    return {d: sorted(names) for d, names in where.items() if len(names) > 1}
```
jtoman, line 456: TODO(future): just use the AST for this.

composer/templates/property_generation_prompt.j2
```diff
@@ -129,15 +129,12 @@ Similarly, you *must* have received positive feedback from the feedback judge on
...
-its unique snake_case title from the batch listing above), list the name(s) of the rule(s)/invariant(s) in your
-spec that verify it, including anything they rest on — a supporting invariant cited with `requireInvariant`,
-a justification rule behind a `require`. Every rule and invariant your spec declares must be named by some
-property: one that no property names is dropped from the report, so the reader would see the property's
-verdict without seeing what it rests on. Every non-skipped property must appear in this mapping with at
-least one rule; skipped
-properties must NOT appear. The result tool will be rejected if this mapping is incomplete or references a
-skipped or unknown property title.
+Each run-target buffer declares its own property->rules mapping on `put_buffer`, so the `result` tool derives
+the published mapping from your buffers — you do not restate it. For each property, name every rule and
```
jtoman, line 133: "you do not restate it" is a useless thing to say at this point; there is no way to restate it anyway assuming you adjusted the result tool accordingly.

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
+## Splitting the spec into verification buffers
+
+Instead of one spec, you author several named CVL **buffers**, each a self-contained spec verified and
```
jtoman, line 3: nit: shared buffers aren't self-contained, verified or reviewed.

composer/spec/source/author.py
```diff
@@ -638,33 +669,123 @@ def judge_tools(self) -> tuple[BaseTool, ...]:
...
+        # Skip quality is a whole-spec concern, so review the skip set ONCE (a skip is owned by no
+        # buffer; scrutinizing it in every buffer's judge would make any skip change re-review every
+        # buffer). The judge sees the whole spec (a skip's justification can rest on the code) and only
+        # the skips, with no per-buffer claim to cover.
+        skips_digest = skips_review_digest(buffers, skipped=skipped_pairs, version_history=vh)
+        if skipped and validations.get(SKIPS_VALIDATION_KEY) != skips_digest:
+            verdict = await self._review(SKIPS_VALIDATION_KEY, combined_buffers_view(buffers), skipped, [])
+            blocks.append(f"=== skipped properties ===\nGood? {verdict.good}\nFeedback {verdict.feedback}")
+            if verdict.good:
+                new_stamps[SKIPS_VALIDATION_KEY] = skips_digest
```
jtoman, line 742: this is probably nothing, but the review skip judge and a buffer named `skip_review` will end up sharing a judge (the collision is avoided on `new_stamps` thanks to the `feedback:` prefix). Worth adding a dedicated "skips_judge" to the `_PerBufferJudge` imo.

composer/spec/source/author.py
```diff
@@ -638,33 +669,123 @@ def judge_tools(self) -> tuple[BaseTool, ...]:
...
 ):
+    # Reviews each run-target buffer independently — against the properties that buffer claims — with its
+    # own persistent per-buffer judge (reached via _review), stamping feedback:<buffer>; the skip set is
+    # reviewed once as a standalone unit (see run()). With no run-target buffers there is nothing to review.
     __doc__ = FeedbackToolBase.__doc__
 
     @override
-    async def _get_feedback(
-        self, spec: str, skipped: list[SkippedProperty]
+    async def run(self) -> Command:
```
jtoman, line 702: considering you're overriding `run` anyway, and `get_feedback` is a hard error, I think there is little value in the `FeedbackToolBase` exporting shared execution infrastructure? As a common tool call schema: sure!

composer/spec/source/author.py
```diff
@@ -793,16 +915,18 @@ async def batch_cvl_generation(
     })
 
     sys_prompt : list[RawPromptInput | type[CacheMarker]] = [
-        lambda load: load(_PROPERTY_GEN_SYS_PROMPT)
+        lambda load: load("property_generation_system_prompt.j2"),
+        f"\nCreate at most {max_spec_buffers()} run-target buffers; fold further properties into "
+        f"existing ones. A single run-target buffer is the one-spec case.",
```
jtoman, line 920: This last sentence is so .... who cares.

composer/spec/source/author.py
```diff
@@ -864,11 +988,17 @@ async def propose(
...
     })
     protected = focus.protected if focus is not None else ()
-    judge_impl = source_feedback_judge(
-        judge_ctx, _LiveJudgeHost(env, editing), judge_prompt, props
+    # One persistent judge per buffer, bound to that buffer's claimed properties, on its own child
+    # context (memory scoped to the buffer, and kept across a rebuild when the claim changes).
+    judge_host = _LiveJudgeHost(env, editing)
+    source_judges = _PerBufferJudge(
+        build=lambda name, claimed: source_feedback_judge(
+            judge_ctx.child(CacheKey[CVLJudge, CVLJudge](name)), judge_host, judge_prompt, claimed
```
jtoman, line 996: RIP cache exploration lol

composer/spec/source/author.py
```diff
@@ -994,8 +1135,10 @@ async def propose(
             # unformalizable" judgment — it's the budget talking. Keep the agent's account.
             return Curtailed(None, detail=res_state["result"])
         return GaveUp(reason=res_state["result"])
-    d = res_state["curr_spec"]
-    assert d is not None
+    # The published artifact is the buffers combined into one document (empty when every property
+    # was skipped, i.e. there are no buffers to combine).
+    _buffers = res_state.get("buffers") or {}
+    d = combined_buffers_view(_buffers)
```
jtoman, line 1141: wait ... doesn't `combined_buffers_view` return a string? How does this end up written out to the FS? or to the web front end? it's not in the VFS that's for source edits...

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+- **complex loops**, or heavy **branching** (path-count explosion).
+
+The method: (1) find these hotspots in the code the properties reach; (2) work out which properties
+genuinely need which hotspots *exact*; (3) group the properties into buffers on that basis, and in each
+buffer summarize the hotspots its own properties do not need exact. Do this partition up front — do not
+start with one monolithic buffer and wait for a timeout to force the split.
+
+When to split:
+- Properties have **conflicting precision needs** — one buffer can summarize a hotspot that another
+  keeps exact (each buffer has its own `methods{}`), with no global-methods-block conflict. This is
```
jtoman, line 30: > each buffer has its own `methods{}`

you said this already?

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+  of the run — you can ignore it and only the buffer you actually edit pays re-verification and re-review.
+
+Structuring buffers:
+- Put infrastructure shared by two or more run-target buffers (ghosts, helper CVL functions,
+  token/oracle models, and summaries several groups need) in a buffer with
+  `is_run_target=false`, and `import "<shared>.spec"` it from **each run-target that uses it**. A shared
+  buffer is pulled in only by the buffers that import it, so
+  editing it re-verifies only those — you can make one shared buffer for a subset of groups and another
+  for a different subset. **Import the shared buffer where you need it; never restate its contents in a
+  run-target** (a duplicate `methods{}` entry or ghost is a compile error).
```
jtoman, line 46: Nit: this is naming too different problems; the former is around copy-paste, the latter is about importing a methods block and then redeclaring an conflicting one right?

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+- **A `requireInvariant`'d invariant must be declared in the run-target buffer whose rule cites it** —
+  NOT in a shared (`is_run_target=false`) buffer. A shared buffer is never run, and importing an
+  invariant does not verify it in the importing run, so citing an unproven invariant is an unsound
+  assumption and is refused at publish. If two run-target buffers each need the same invariant, declare
```
jtoman, line 50: oh no one cares why, just state the rule.

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+
+Structuring buffers:
+- Put infrastructure shared by two or more run-target buffers (ghosts, helper CVL functions,
+  token/oracle models, and summaries several groups need) in a buffer with
+  `is_run_target=false`, and `import "<shared>.spec"` it from **each run-target that uses it**. A shared
+  buffer is pulled in only by the buffers that import it, so
+  editing it re-verifies only those — you can make one shared buffer for a subset of groups and another
+  for a different subset. **Import the shared buffer where you need it; never restate its contents in a
+  run-target** (a duplicate `methods{}` entry or ghost is a compile error).
+- **A `requireInvariant`'d invariant must be declared in the run-target buffer whose rule cites it** —
```
jtoman, line 47: however, it is worth emphasizing that this is the exception to the "put shared infra in a shared buffer".

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+  NOT in a shared (`is_run_target=false`) buffer. A shared buffer is never run, and importing an
+  invariant does not verify it in the importing run, so citing an unproven invariant is an unsound
+  assumption and is refused at publish. If two run-target buffers each need the same invariant, declare
+  it in each of them.
+- Each **run-target** buffer declares its `property_rules` (the properties it verifies + their rule
+  names). Across all run-target buffers, every non-skipped property must appear in **exactly one**
+  buffer, and each rule lives in exactly one buffer.
+
+Summaries — your main lever for tractability:
+- For each run-target buffer, **summarize now**, in its `methods{}`, the hotspot functions its
```
jtoman, line 57: "now" is mushy for a long context agent, just say "up front", although you have already mentioned this bit a few times already.

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+  a status board: which buffers are `complete`, `running`, or `needs (re)submission`. Process a finished
+  buffer immediately — fix its issue if it has any (CEXs, timeouts, errors, ...) and `submit_buffer` it
+  again — while the others keep proving.
+- Work this priority order, and only ever block at the last step:
+  1. A finished result not yet processed (a counterexample, but also a timeout, sanity failure, or
+     error) → handle it. Diagnose whether the fix belongs in this buffer or in a **shared** buffer it
+     imports (e.g. an under-approximation in shared infra). Once you edit a shared buffer, every buffer
+     importing it goes stale, so the board then lists them (including ones already verified) under `needs
+     (re)submission` — re-submit each. The board reflects that staleness only after your edit; it does
+     not infer that a result needs a shared fix.
```
jtoman, line 104: These last two sentences are still nonsense; the board received from the prior `collect_buffers` call doesn't reflect anything; its static text. It also doesn't update on an edit, because again, its static text.

Say what actually happens. "Edits to shared buffers invalidates the results for already verified buffers that import them, and the results of any inflight jobs that depend on those shared buffers. Future status board results will reflect this by marking the affected jobs as needing resubmission"

composer/templates/spec_buffers_guidance.j2
```diff
@@ -0,0 +1,109 @@
...
+  2. Else a buffer still to author/submit → author it and `submit_buffer` it.
+  3. Else everything is submitted and running with nothing finished to process → call
+     `collect_results(wait=true)` to sleep until the next job finishes.
+- Editing a buffer (or a shared buffer it imports) makes its prior verification stale; re-submit it. A
```
jtoman, line 108: this is like ... the 4th time you've said this

jtoman (commented):
Honestly, we're really close. The most glaring issue is how outputs are managed, I think right now you'll just end up with all of the files concatenated together in `autospec_ComponentName.spec` which won't work like ... even a little bit.

composer/spec/source/author.py
jar-ben: hm - the base isn't only `EditorAwareFeedbackTool`. `VanillaFeedbackTool` (the natspec feedback tool,
bound via `property_tools`) still uses the base's `run` + `_get_feedback`; only the editor-aware/buffers
subclass overrides them. so the base is genuinely shared execution infra for the vanilla path, not just
the tool-call schema (we can split it in a follow-up if it makes sense?)

composer/spec/source/author.py
jar-ben: done - the skip review has its own `_skips_judge` keyed on `SKIPS_VALIDATION_KEY`, so a buffer literally
named `skip_review` can't collide with it.

composer/spec/source/author.py
jar-ben: dropped.

composer/spec/source/author.py
jar-ben: right, fixed. `combined_buffers_view`
is only the reviewable concat now; the delivered/proved artifacts are the per-buffer `.spec` files
(`spec_files`), written by `write_artifact`. nothing publishes the concatenation as a runnable spec.

composer/spec/source/buffer_tools.py
jar-ben: ah - documented on `put_buffer`: an empty/omitted `property_rules` keeps the previously-declared mapping.

composer/spec/source/prover.py
jar-ben: switched to `@functools.lru_cache` on the local `cur_digest`.

composer/spec/source/prover.py
jar-ben: good point - now bounded: it waits
for a job with a per-tick timeout and re-checks `budget_pressure()` each tick, so a firing wrap-up
interrupts it instead of blocking.

composer/spec/source/prover.py
jar-ben: fixed.

composer/spec/source/prover.py
jar-ben: yep - `_run_buffer_job` puts its result (or a caught error) on the queue as its last step before
returning, so `task.done()` implies the result is already on the queue.

composer/spec/source/prover.py
jar-ben: it wasn't - fixed. each `ProverRunLog` carries its own `link`, so we don't carry a link across different
buffers in the drain loop anymore.

composer/spec/source/prover.py
jar-ben: it was already gated on `not needs_submit and not running`, so an un-submitted `bar` in `needs_submit`
blocked the message - but naming them is clearer, so it now says "buffers B1, B2, ... verified at their
current content" from `complete`.

composer/spec/source/spec_buffers.py
jar-ben: noted - `name` is the on-disk spec stem, so it has to be a valid path component.

composer/spec/source/spec_buffers.py
jar-ben: well, it's a time efficiency question. Do we want to run certoraRun everytime
we compute the import_closure? Note the lightweight AST jar we already use for
requireInvariant does NOT give us the import info (it rejects/blanks import lines),
so getting imports reliably means the heavier resolving parse - runtime seems quite
high. If we get a lightweight execution option for this, then yes. I wanted to defer
this; I'll look into a reliable and fast solution later.

composer/spec/source/spec_buffers.py
jar-ben: ha - `requireinvariant_citations` does read the AST, but it's a one-shot: it only runs on the publish
gate, once per buffer at completion. `imports_in_cvl` runs inside `buffer_digest`, recomputed all the
time - that's the path I don't want to put a jar spawn on. and the import case can't just reuse this
AST anyway: the cheap `syntax-check` mode rejects/blanks `import` lines, so `importedSpecFiles` comes
back empty - getting imports means the heavier resolving parse.

composer/spec/source/spec_buffers.py
jar-ben: typed the citation/declaration sets as `set[RuleName]`.

composer/spec/source/spec_buffers.py
jar-ben: good catch - done

composer/spec/source/spec_buffers.py
jar-ben: could, but I'm worried about runtime: this dup check is very frequent (it runs on the status render),
unlike the requireInvariant check which we only do on submit. also low stakes - it's a heuristic hint
we surface to the agent ("these look copy-pasted, consider sharing them"), not a soundness gate and not
even a hard gate, just a recommendation. leaving the TODO and deferring.

composer/templates/property_generation_prompt.j2
jar-ben: dropped it.

composer/templates/spec_buffers_guidance.j2
jar-ben: right - scoped "self-contained, verified and reviewed" to run-target buffers; shared buffers are
imports-only.

composer/templates/spec_buffers_guidance.j2
jar-ben: dropped the repeat.

composer/templates/spec_buffers_guidance.j2
jar-ben: right - split them.

composer/templates/spec_buffers_guidance.j2
jar-ben: called the `requireInvariant`-in-the-run-target rule out as the explicit exception to "put shared infra
in a shared buffer".

composer/templates/spec_buffers_guidance.j2
jar-ben: trimmed to the rule.

composer/templates/spec_buffers_guidance.j2
jar-ben: "now" -> "up front", and de-duped.

composer/templates/spec_buffers_guidance.j2
jar-ben: agreed - adjusted

composer/templates/spec_buffers_guidance.j2
jar-ben: dropped the repeat.

jar-ben:
Thanks for the thorough round — responses are inline on each thread. Quick summary:

**Output management (your headline concern) is fixed.** yep, this was the real gap. the delivered artifact is now the per-buffer specs the run actually proved -
each buffer as its own `certora/specs/<slug>/<buffer>.spec` + a `.conf` per entrypoint, importing the
shared `certora/specs/summaries/` a level up. `combined_buffers_view` stays only as the reviewable
concat (I am not sure how we use this one / if we need it).

I also validated the delivered layout end-to-end on a real run: per-buffer `certora/specs/<slug>/<buffer>.spec` + a per-buffer `certora/confs/<slug>/verify_<buffer>.conf`, shared summaries imported via `../summaries/`, shared buffers import-only, and static spec-coverage on par with the single-spec baseline — so the split is coverage-neutral.

**One unrequested change to flag** (`artifact_text` / `write_quarantined`): new here, flagging since it's unrequested. `artifact_text` is the `FormalResult` single-text *view* of
the result (only reader now is the CLI cache printer) - not the deliverable. the actual multi-file
`.spec`/`.conf` output is written by the overridden `write_artifact`/`write_quarantined`. `FormalResult`
assumes one primary text artifact, which fits the other (single-artifact) backends; the prover is the
only multi-file one and absorbs that locally via overrides + this joined-text view, rather than
broadening the cross-backend protocol. deliberate.

The AST-for-imports / dup-detection suggestions and the `FeedbackToolBase` slimming I've answered inline — deferred or pushed back with rationale. Re-requesting review — thanks!

composer/spec/source/artifacts.py
```diff
@@ -68,26 +92,36 @@ def _artifact_dir(self) -> Path:
...
-        the fixed run overlay (shared with the live prover run). No-op if no
-        base config."""
-        if base_config is None:
-            _log.warning("no base config for %s; skipping conf dump", spec.stem)
-            return
-        conf = prover_config_overlay(
-            base_config,
-            main_contract=self._main_contract,
-            verify_target=f"{self._main_contract}:{spec_path}",
+        specs_root = under_project(self._project_root, i.specs_dir)
```
jtoman: do we record anywhere stable (besides the langgraph checkpoitn db) the mapping of properties to spec files? Just saying "this property is covered by these rules" doesn't tell you what files those are in, right?

composer/spec/source/author.py
```diff
@@ -200,7 +204,8 @@ async def run(self) -> Command | str:
         # lives only in another buffer — e.g. an unproven shared buffer — is an unproven assumption.
         cited_by_buffer = {b.name: requireinvariant_citations(b.cvl) for b in run_targets(buffers)}
         if (err := validate_requireinvariant_proved(
-            buffers, cited_by_buffer, declared_by_buffer
+            buffers, cited_by_buffer, declared_by_buffer,
+            expected_to_fail=set(self.state["rule_skips"]),
```
jtoman: ... wait, sorry, I'm just now realizing; are rule skips also not buffer namespaced? So expecting the failure of "transfer_integrity" will ignore all failures across all buffers? That's **probably** fine because rule names tend to be uniqueish, but probably worth addressing...

Given all of the state around these buffers, it might be worth adding an amend buffer data tool, so an agent can amend the property -> rule mapping or this "skip list" on a per buffer basis, without having to re-emit the whole buffer text.

composer/spec/source/buffer_tools.py
```diff
@@ -28,6 +30,15 @@ class WithBuffers(TypedDict):
     buffers: dict[str, NamedBuffer]
 
 
+def _escapes_component_dir(name: str) -> bool:
+    """True if ``name`` — the buffer's on-disk stem under ``certora/specs/<component>/`` — resolves
+    outside that dir. A leading ``/`` or a ``..`` that climbs out escapes; a plain subdir does not. This
+    is what confines every buffer to its own component, keeping the shared ``summaries/`` and other
+    components' spec dirs immutable to the agent."""
+    n = posixpath.normpath(name)
+    return n.startswith("/") or n == ".." or n.startswith("../")
```
jtoman: prefer `x = pathlib.PurePosixPath(name); len(x.parts) == 1 and not x.is_absolute()` TIOLI I guess, I just have no idea what `posixpath.normpath` does so I can't say if this is right.

composer/spec/source/prover.py
```diff
@@ -403,8 +409,9 @@ def completing_run_links(
             link = elem.get("link")
             if link and link not in seen:
                 seen.add(link)
-                links.append(link)
-    return links
+                spec = (component_specs_dir(slug) / f"{b.name}.spec").relative_to(SPECS_DIR).as_posix()
```
jtoman: this is the third place you've added code to compute "this is the place on disk specs go", this will drift and drive us insane. centralize please.

composer/spec/source/prover.py
jtoman: I didn't realize an unverified `bar` landed it in `needs_submit`, but I agree with the conclusion anyway

composer/spec/source/prover.py
jtoman: so why is this `link` variable still here? I see it's used down below to update the `prover_link` field in the author state, but that field is basically meaningless now (and, arguably, after my striping change, was already). Drop this variable, and drop the `prover_link` field from the state, unless that will be a "whole thing".

composer/spec/source/spec_buffers.py
jtoman: oh I didn't know that about the require invariant AST jar. never mind then.

jtoman (commented):
two final small-ish things.

PR #223, opened 2026-09-08
autoprover: re-fetch a completed cloud job on transport failure

jar-ben:
A transient failure while downloading a *completed* (SUCCEEDED) cloud job's results used to bubble up and re-run the entire (often hours-long) proof. Retry the fetch itself with capped exponential backoff against the already-finished job instead — never re-prove. Complements the ProverCLI request-level retry/timeout: POU retries individual requests; this covers a whole-fetch failure that outlives those.

### DIFF
--- a/composer/prover/cloud.py
+++ b/composer/prover/cloud.py
@@ -7,6 +7,7 @@ job URL (so waiting needs no credentials), then retrieves its results through
 
 import asyncio
 import logging
+import shutil
 import tempfile
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


composer/prover/cloud.py
```diff
@@ -159,6 +160,42 @@ def _results_api() -> ProverOutputAPI:
...
+    Each attempt starts from an empty ``dest`` so a partially-written archive from a failed attempt
+    cannot leave a truncated file behind. Never re-runs the job — only re-reads it.
+    """
+    for attempt in range(1, _FETCH_MAX_ATTEMPTS + 1):
+        try:
+            await asyncio.to_thread(
+                _results_api().fetch_sources_and_treeview_files, job_id, dest
+            )
+            return
+        except Exception as exc:
```
ericeil: I wonder if this should be more selective?  See my (otherwise nearly identical) change here: https://github.com/Certora/AutoProver/pull/238/changes#diff-739b06c0facd657e613efd97617f3caf229985cd660d24a05097c9deb53d73f5R195-R197

ericeil (approved):
Looks good!

PR #226, opened 2026-09-08
autosetup: robust CVL import parsing (fixes prover-cache invalidation)

jar-ben:
parse_imports_from_spec was a line-based regex that missed a second import on the same line and any import preceded by an inline block comment. Those imports feed the prover cache key (enhanced_config_manager -> cloud_runner), so a missed import let a stale cached result survive an edit to that import.

Replace it with imports_in_cvl(): a char-level port of the CVL lexer's import rule (a string literal is one whole token, comments are trivia), so an import inside a string or a // or /* */ comment is never picked up and a real import is never missed regardless of line structure. Over-including is sound; under-including is not.

### DIFF
--- a/certora_autosetup/parsers/spec_imports.py
+++ b/certora_autosetup/parsers/spec_imports.py
@@ -1,9 +1,61 @@
 """Utilities for parsing CVL spec file imports."""
-import re
 from collections import deque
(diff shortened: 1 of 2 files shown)


shellygr (approved):
oh dear

PR #227, opened 2026-09-08
Take graphcore 85be3db, which bounds the no-tool-call scolding loop

shellygr:
## What this is

The graphcore submodule and the `pyproject.toml` pin move from `2177f57` to `85be3db`. That
range is Certora/graphcore#38 and nothing else. `uv.lock` stays as it is, because
`[tool.uv.sources]` resolves graphcore to the local path, so the SHA lives in those two places
only. It is the same file set the previous bump touched.

## What #38 changes

An AI turn with no tool call cannot advance a graphcore graph, so `_scolding_node` appends a
message saying every turn must end with a tool call and routes back for another try. Nothing
counted the retries. A model that answers every scolding the same way cycles until langgraph's
recursion limit, hundreds of paid calls later.

That happened on a cloud run here. One agent returned an empty message after writing a rough
draft and went round 494 times over about fifteen minutes, taking most of the run's LLM spend,
almost all of it cache-write on re-sending the growing scolded conversation. Three other agents
in the same run reached the scolding node once each and recovered on the next turn, which is
where the new limit of three comes from.

## What it does not fix

The reason that run died rather than losing a single phase is separate. The
structural-invariant task is gathered without `return_exceptions=True`, so its failure cancels
AutoSetup and takes the process down. #214 removes that phase and the fatal path with it. This
bump only stops the agent from spinning in the first place.

## Checks

`scripts/check_submodule_pin.py`, pyright and pytest all run clean locally.


jar-ben (approved):
sure

PR #228, opened 2026-09-12
Let a component lose property extraction without losing the run

shellygr:
## What happened

When one of five components hit an Anthropic 500, a dev run died 35 minutes in:

```
19:37:33 task failed: phase=BUG_ANALYSIS task_id=extract-3 wall=1393.66s error=APIStatusError
APIStatusError: {'type': 'error', 'error': {'type': 'api_error', 'message': 'Internal server error'}}
```

`extract-0` and `extract-1` had finished (844s and 1271s) and never became CVL. `extract-2` and
`extract-4` were 15 and 9 minutes in and were destroyed mid-flight. System analysis, the `HARNESS`
phase, the prover setup phase and structural invariants had all succeeded. No report, no spec, no prover runs.

This is a recurring class, not a one-off. Two earlier runs went the same way, one of them after
about eight hours of prover time.

## Why the retry did not fire

The run-wide retry did not fail. It was not in the image. The dev AutoProver image that ran this
job pinned `f1f62a55`, which is older than #217 and #220. At that commit `should_retry` looked
only at the status code. An error the server reports mid-stream arrives on the open 200 response
of the stream. The predicate returned False and the task died on its first attempt. The log has no
"(Retry 1)" line. #220 reached that image four days later, with the next pin bump.

Two cloud images pin AutoProver separately. An earlier version of this description said the
shipped pin already had #217 and #220. That was true of the other image only.

## The fan-out

`core.py` gathered the per-component extraction tasks with no `return_exceptions`, so the first
exception propagated out of `run_pipeline` and killed the process. The sibling tasks were orphaned
at loop teardown. Formalization, twelve lines up the file, does the opposite: it gathers
tolerantly, wraps each settled value in a `ComponentOutcome`, and lets `_tally` turn a
`BaseException` into a failure string. That asymmetry was the bug.

`_settle_extraction` splits a finished fan-out into an `ExtractionResult`: the batches it produced
and the components that failed, each an `ExtractionFailure` of the unit and its error. It is a
module-level helper rather than inline so the partition is testable without standing up a backend.

A component lost at extraction never reached the formalizer. It stays out of `outcomes`, and the
formalizer is never handed one it did not see. It goes into the report inputs and into `_tally`
instead, so a lost component is visible in both the report and the run verdict.

If every component fails, the run still fails, but the error now names them rather than only
saying that nothing was extracted.

## The retry predicate

Not what killed this run, but the same family. An error the server reports mid-stream arrives on
the already-open 200, so the status code cannot speak for the error. For that case, #220 taught
`AnthropicService.should_retry` to fall through to the payload's `error.type`. `openai.py`
and `openrouter.py` were still status-only and fail identically on the same payload.

The extractor moves to `provider.py` as `payload_error_type`: the `{"error": {"type": ...}}` shape
is common to the Stainless SDKs. The roster is not. Which type values count as transient is each
provider's own vocabulary, so the OpenAI-family set is separate and deliberately narrow
(`server_error`, `rate_limit_exceeded`). `insufficient_quota` and the `invalid_request_error`
family are deterministic, and a type the set does not name stays deterministic rather than being
guessed at.

Worth flagging: unlike the Anthropic side, this half is not backed by an observed incident. I have
not seen an OpenAI mid-stream error payload in a real run. The roster is reasoned from the
documented taxonomy rather than from a trace.

## Tests

`tests/test_pipeline_extract_failures.py` covers the partition. One component failing leaves the
others. A component with no properties is still not a failure. Every component failing is
reported. `CancelledError`, `BudgetExceeded` and `BudgetPressureAbort` still propagate. A
component failure that settled first does not change that. The wipeout message naming the failures is driven through
`run_pipeline`.

`tests/test_openai_retry.py` is the twin of `test_anthropic_retry.py`, parametrized over both
OpenAI-family services.

Two existing fakes of `_extract_all` now return an `ExtractionResult`: `test_pipeline_overlap.py`
and `test_pipeline_staged_formalizer.py`.

1314 pass, and pyright is clean over the CI scope.

### DIFF
--- a/composer/llm/anthropic.py
+++ b/composer/llm/anthropic.py
@@ -2,7 +2,6 @@
 ``ModelProvider`` implementation that mints ``ChatAnthropic`` instances."""
 
 from typing import Literal, TypeGuard, Any, TYPE_CHECKING, override, cast
-from collections.abc import Mapping
 from io import BytesIO
 from dataclasses import dataclass, field
 import asyncio
@@ -14,7 +13,8 @@ import httpx
 from composer.input.files import UploaderBase, ContentRenderer
 from composer.input.types import ModelConfiguration
 from composer.llm.provider import (
-    ProviderServiceBase, ProviderSpec, compaction_threshold, standard_callbacks
+    ProviderServiceBase, ProviderSpec, compaction_threshold, payload_error_type,
+    standard_callbacks
 )
 from composer.llm.pricing import PriceProvider, price_provider_for
 from .types import CacheLevel
@@ -219,19 +219,6 @@ RETRYABLE_ERROR_TYPES = frozenset({
 status code cannot speak for them."""
 
 
-def _payload_error_type(body: object) -> str | None:
-    """The ``error.type`` discriminator of an Anthropic error payload, or ``None``
-    for a body of any other shape (the SDK leaves it as the raw text when the
-    payload does not parse)."""
-    if not isinstance(body, Mapping):
-        return None
-    error = body.get("error")
-    if not isinstance(error, Mapping):
-        return None
-    error_type = error.get("type")
-    return error_type if isinstance(error_type, str) else None
-
-
 class AnthropicService(ProviderServiceBase):
     def __init__(self):
         from graphcore.tools.memory import anthropic_async_memory_tool
@@ -278,7 +265,7 @@ class AnthropicService(ProviderServiceBase):
             # 200 response, and the SDK builds the exception from that response — so the
             # status code carries none of the meaning and the class stays the base
             # APIStatusError. The transient/deterministic split lives in the payload.
-            return _payload_error_type(exc.body) in RETRYABLE_ERROR_TYPES
+            return payload_error_type(exc.body) in RETRYABLE_ERROR_TYPES
         return False
 
 @dataclass
--- a/composer/pipeline/core.py
+++ b/composer/pipeline/core.py
@@ -73,7 +73,9 @@ from .keys import (
     POST_PROPERTY_KEY, PRE_PROPERTY_KEY, PRIORITIZATION_KEY, PROPERTIES_KEY,
     SYSTEM_ANALYSIS_KEY, PLUGIN_FORMALIZATION_KEY, candidates_digest
 )
-from composer.diagnostics.budget import total_budget, named_budget_or_nop, time_budget
+from composer.diagnostics.budget import (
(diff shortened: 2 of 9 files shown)


composer/pipeline/core.py
```diff
@@ -871,7 +881,7 @@ async def _extract_all[P: enum.Enum, H, Main, U: FeatureUnit](
     # ``Main``/``U`` (matching the caller's), so there's nothing to tie it to.
     ecosystem: Ecosystem[Any, Main, U],
     plugins: PluginPhaseManager[P, U],
-) -> list[_Batch[U]]:
+) -> tuple[list[_Batch[U]], list[tuple[U, BaseException]]]:
```
shellygr, line 884: fever-inducing types again

composer/llm/anthropic.py
```diff
@@ -278,7 +265,7 @@ def should_retry(self, exc: Exception) -> bool:
             # 200 response, and the SDK builds the exception from that response — so the
             # status code carries none of the meaning and the class stays the base
             # APIStatusError. The transient/deterministic split lives in the payload.
-            return _payload_error_type(exc.body) in RETRYABLE_ERROR_TYPES
+            return payload_error_type(exc.body) in RETRYABLE_ERROR_TYPES
```
jtoman, line 268: Sorry I'm confused. Do we know why this didn't save the run? Here it doesn't look like you're actually changing the vocabulary or anything, so the retry mechanism which (by my reading) should have just gracefully retried failed. Why?

composer/pipeline/core.py
```diff
@@ -953,8 +963,41 @@ async def budgeted_task(u: U) -> _Batch | None:
         with named_budget_or_nop("property_extraction"):
             return await _one(u)
 
-    got = await asyncio.gather(*[budgeted_task(u) for u in ecosystem.units(main)])
-    return [b for b in got if b is not None]
+    units = list(ecosystem.units(main))
+    got = await asyncio.gather(
+        *[budgeted_task(u) for u in units], return_exceptions=True
+    )
+    return _settle_extraction(units, got)
```
jtoman, line 986: its funny you wrote this function, I have a local version too hah

jtoman (approved):
the actual resilience is fine, I'm more annoyed/concerned that the retry mechanism didn't seem to work!

composer/llm/anthropic.py
shellygr: it wasn't part of the run it seems.

PR #229, opened 2026-09-12
diagnostics: attribute every billed LLM call to its origin

jar-ben:
## The problem

A run's billed token totals and the message history reachable through `ap-trail`
cannot currently be reconciled, because nothing records *where a call came from*.

On a recent 4h25m dev run (5 components, ~$780), only **$168 of the spend traces
to a recorded LLM call**. The other ~$575 — 91M Opus cache-write tokens, 73% of
the run — is billed to `CVL_GEN` and appears in neither the executor log nor the
run's message history.

This is not a counting error. Per phase, an `ap-trail` export and
`ap_report/job_info.json` agree **to the token**:

| cost_center | billed cache_write | in message history | coverage |
|---|---|---|---|
| `DISCOVER_DESIGN_DOC`, `COMPONENT_ANALYSIS`, `HARNESS`, `BUG_ANALYSIS` | — | — | **100%** |
| `formalization_preparation` | 3,507,156 | 1,545,879 | 44% |
| `formalization` | 96,348,840 | 6,278,951 | **7%** |

`by_phase` sums exactly to `totals`, so nothing is orphaned into a stray task
lane — the calls run inside a real task's scope and are billed to it correctly.
They simply leave no trace of *which code path* made them. `cache_read` is 94%
covered while `cache_write` is 7%, so the missing calls write a large fresh
prefix and read almost nothing: repeated cold starts on a big context.

A second run (9h06m, 2 components, ~$565) shows the same shape — $282 of $565
unattributable, again concentrated in the two formalization cost centers. So it
is not a one-off.

Candidates we currently cannot distinguish:

- `with_structured_output(...).ainvoke(...)` call sites — one-shot calls that
  never enter message state, and at least one sits in the formalization path;
- retried calls — `install_retry_policy` resumes from the last checkpoint, so a
  failed attempt is billed by the provider but never checkpointed.

Already ruled out, so the PR is not chasing these: cache-TTL expiry (4 of 665
Opus calls followed a >300s gap, carrying 1% of writes); server-side context
editing (`context_management: {"applied_edits": []}` on all 625 formalization
calls, i.e. it never fired); and the agentic CEX analyzer in
`composer/prover/agentic_analyzer.py` — its
`run_to_completion(..., within_tool=…)` becomes `ThreadMeta.from_tool_id`, so it
*would* register threads, and there are none.

Rather than keep guessing, make the run say it.

## The change

`UsageCallback` is attached at model construction, so it is the one place that
sees every billed call — which is exactly why `by_phase` sums to `totals`. That
makes it the right place to capture a call's **origin**.

**`composer/diagnostics/timing.py`**

- New `CallOrigin(thread_id, node)` — the LangGraph thread and node a call ran
  under, or `None`/`None` for a call made outside any graph. Absence is the
  signal, so it is recorded as absence rather than back-filled from ambient
  context.
- `RunSummary` gains `token_usage_by_origin` / `calls_by_origin`, keyed by
  `(origin, task_id)`. Deliberately a separate axis from `PhaseRecord`: the
  question it answers ("is this call in a thread the trail recorded?") cuts
  across phases and outlives them.
- `record_token_usage(...)` takes `origin=`.
- `token_usage_summary()` gains **`by_origin`**, sorted heaviest-first, so it
  rides out through `ap_report/job_info.json` next to `by_phase`.
- Bounded at 2000 distinct origins, overflowing into a single
  `(origin-overflow)` bucket, so a pathological run cannot grow the summary
  without bound.

**`composer/diagnostics/usage_callback.py`**

- Metadata reaches the *start* hooks only, so the origin is stashed at
  `on_chat_model_start` / `on_llm_start` and re-joined at `on_llm_end` by
  LangChain's per-call `run_id`. Single-threaded thanks to `run_inline`, so a
  plain dict is safe; capped at 4096 in-flight entries.
- `on_llm_error` releases the stashed origin, so a failed call cannot leave an
  entry behind to be mis-joined onto whichever call reuses the slot.
- A per-call `DEBUG` line (model, task, cost center, thread, node, four token
  buckets) for live triage.

No behaviour change to what is billed or to any existing field: `by_origin` sums
to `totals`, and `totals` / `by_model` / `by_phase` are untouched.

Illustrative output — two calls billed to the same task, one graph-driven and
one not, which `by_phase` cannot tell apart:

```
{'thread_id': None,                 'node': None,    'task_id': 'formalize-0', 'calls': 2, 'cache_write': 90002}
{'thread_id': 'autoprove_abc-cvl',  'node': 'tools', 'task_id': 'formalize-0', 'calls': 1, 'cache_write':   200}
```

## Tests

Three added to `tests/test_token_usage.py` (12 → 15):

- `test_origin_separates_in_graph_from_out_of_graph_calls` — two calls billed to
  the same task land in distinct origin buckets.
- `test_by_origin_sums_to_totals` — including a call whose start event was never
  seen and one with no `run_id` at all; neither may go missing from accounting.
- `test_failed_call_releases_its_origin`.

Full suite: **1293 passed, 10 skipped** on this branch vs **1290 passed, 10
skipped** on `master` — exactly the three new tests. The 27 errors are identical
on both and are environmental (the RAG-DB tests need postgres; one Solana test
needs a real program).

## What this buys

One run answers the question. Tooling that reconciles billed spend against the
trail currently reports only "unattributed cache-write spend: $575" and cannot
say more; with `by_origin` it can name the thread — or state definitively that
the call belonged to no thread, and show which task and node produced it.

## Not in this PR

- Fixing whatever `by_origin` turns out to indict. This only makes it visible.
- Registering additional threads with `thread_logger`. If the spend turns out to
  be out-of-graph structured-output calls, they have no thread to register, and
  `by_origin` is the right permanent home for them regardless.

### DIFF
--- a/composer/diagnostics/timing.py
+++ b/composer/diagnostics/timing.py
@@ -66,6 +66,37 @@ class TokenTotals:
         }
 
 
+@dataclass(frozen=True)
+class CallOrigin:
+    """Where one LLM call came from, independent of which task was billed for it.
+
+    ``thread_id`` / ``node`` are the LangGraph run's, read off the call's metadata;
+    they are ``None`` for a call made outside any graph (a bare
+    ``with_structured_output(...).ainvoke(...)``, a summarizer, a retried attempt
+    that never reached a checkpoint). That distinction is the whole point of the
+    type: a run's *billed* totals and the message history reachable through
+    ``ap-trail`` can only be reconciled if each call says whether it belongs to a
+    recorded thread. Calls with no ``thread_id`` are exactly the ones no
+    transcript will ever account for.
+    """
+    thread_id: str | None = None
+    node: str | None = None
+
+    @property
+    def in_graph(self) -> bool:
+        return self.thread_id is not None
+
+    def as_dict(self) -> dict[str, str | None]:
+        return {"thread_id": self.thread_id, "node": self.node}
+
+
+#: Cap on distinct origins tracked per run. Origins are (thread_id, node) pairs,
+#: so a normal run has a few hundred; the cap only stops a pathological run from
+#: growing the summary without bound. Overflow folds into ``_ORIGIN_OVERFLOW``.
+_MAX_ORIGINS = 2000
+_ORIGIN_OVERFLOW = CallOrigin(thread_id="(origin-overflow)", node=None)
+
+
 @dataclass
 class PhaseRecord:
     task_id: str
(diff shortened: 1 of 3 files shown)


shellygr (approved):
lgtm

PR #232, opened 2026-09-15
Read verdicts from every run that accounts for the published spec

shellygr:
## The bug

A run proved 25 rules and the report showed 23, with two blank UNKNOWN rows carrying no link
and no duration. The rules were fine. The report was reading one prover run.

`make_prover_fetcher` fetched verdicts from `formalized.run_link`, the component's last run.
That is the wrong assumption about how a spec gets proved. `_is_completion_history` accepts
the contiguous newest-first stretch of runs recorded against the current authoring state, and
says so in its own comment: "a piecemeal completion is as good as a full-run one". An author
may therefore park a stubborn rule, prove the rest in a full run, and come back to that one in
a scoped re-run. The last run is then not the whole story, and anything only the scoped run
proved reads in the report exactly like a rule the prover could not decide.

On the run that prompted this, the two rows were a lemma that had timed out three times and
the helper lemma the author introduced to get it through. Both verified in a 289-second scoped
run. The property whose proof rests on them was published green, above two blank rows.

## What changed

The report now reads the same stretch of history completion is judged on, via
`covering_run_links` over the runs `_iterate_history` walks, and stamps each verdict with the
run that produced it. `ProverRunLog` gained the run's link so that attribution is possible at
all; the field is `NotRequired`, so an older checkpoint replays.

Across runs the newest verdict wins. This is deliberately *not* `Verdict.merge`, whose
most-terminal-wins rollup is right within a single run and would otherwise let a stale TIMEOUT
outrank the VERIFIED that superseded it. That is the one thing in here most likely to be got
wrong, so it is tested directly.

A rule that genuinely no run reported still renders UNKNOWN, since nothing else would be
honest, but `coverage.warnings` now names it.

## The gate is untouched, on purpose

My first reading blamed the publish gate, and that was wrong. `all_rules` comes from
`declared_rules_list`, a separate typechecker pass, so the coverage check is not circular, and
a run's rule selection lives in the generated conf rather than the CLI args, which is what made
five runs look unscoped in the log. The gate accepted this run because the rules really were
all covered. Piecemeal completion saves whole re-runs and should stay.

## Third commit, separate concern

Two things a run did not record, both of which cost us time on the run above.

The stuck-rule nag fires on the third identical failure and offers only "change your approach"
or "skip the property". It now also lists the tools the plugins gave the author, pulled from
`ProvidedTools.tools`, so nothing plugin-specific enters this repo. It says nothing about what
those tools are for: the author reads each description and decides whether one fits the
failure. And `report.json` records `active_plugins`, without which a finished run leaves no
trace of which plugins took part.

## Review round

Three commits answer the review. The last-run fallback in `covering_output_links` is gone,
because a published result always has a run at the spec state it published. `_fetch_covering`
now says that its argument must arrive newest first, which is what makes the first verdict
found the right one to keep. And the nag no longer calls every plugin tool a way to hand a
rule off.

## Checks

`pytest -m "not expensive"`: 1330 passed. `pyright`: 0 errors. With current master merged in
locally: 1392 passed, 0 errors. New tests cover `covering_run_links` (newest-first, stops at a
state change, transparent to nag markers, skips link-less runs), the two fetcher behaviours
above, a result with no covering run, the reminder text, and the list of active plugins.


composer/spec/source/report/schema.py
```diff
@@ -331,6 +331,11 @@ class AutoProverReport(BaseModel):
     run_mode: str | None = None
     #: What a prioritized run set aside. Empty for a comprehensive run.
     deprioritized: list[DeprioritizedProperty] = Field(default_factory=list)
+    #: Plugins active for the run, as the manifest that feeds every per-component cache key.
+    #: Without it a finished run leaves no record of which plugins participated, and whether an
+    #: installed plugin was even loaded can only be inferred from the container image. Empty for
+    #: a run with no plugins, and on reports written before the field existed.
+    active_plugins: list[str] = Field(default_factory=list)
```
jtoman, line 338: I swear I fixed this??? It might be one of those "kicking around on a branch" things... *sigh*

composer/spec/source/prover.py
```diff
@@ -237,6 +273,28 @@ def _iterate_history(
...
+    """
+    links: list[str] = []
+    for elem in reversed(l):
+        if elem["sort"] != "run":
+            continue
+        if elem["state_digest"] != curr_digest:
+            break
+        if (link := elem.get("link")) is not None:
+            links.append(link)
+    return links
```
jtoman, line 295: is it worth actually recording the rules executed as part of the run? That's available in the run history, and it leaves no confusion about which run link yielded which result right? although I guess a VERIFIED in run a is just as good as a VERIFIED in run b; this only matters for CEX analysis I guess. hmm

composer/spec/source/prover.py
```diff
@@ -682,16 +746,11 @@ async def run_in(run_root: str) -> str | Command:
...
-                    "You may need to significantly change your approach, or skip the property if this is a persistent issue (you may need to use rebuttals to communicate"
-                    " these failures to the feedback judge)."
-                ]
-                if seen_post_compaction_history:
-                    nag_channel["reminders_channel"].append(
-                        "(NB: Some of these prover calls happened before your most recent task history summarization)"
-                    )
+                nag_channel["reminders_channel"] = stuck_rule_reminder(
+                    to_warn,
+                    delegation_tools=state.get("delegation_tools") or (),
```
jtoman, line 751: this kind of assumes every tool provided by a plugin is for resolving prover failures. That's not true in general, you could imagine a tool that's able to search a corpus of example CVL rules which would help it *formulate* a rule, not necessarily deal with a timeout. But this is fine until we want to whip up a "what is this tool for" expression.

composer/spec/source/report_prover.py
```diff
@@ -44,21 +44,38 @@ def _fetch(api: ProverOutputAPI, link: str) -> dict[RuleName, Verdict]:
...
             c.duration or None,
             Path(loc.file).name if (loc and loc.file) else None,
+            link=link,
         )
         name = RuleName(c.rule_name)
         verdicts[name] = cand.merge(verdicts.get(name))
     return verdicts
 
 
+def _fetch_covering(api: ProverOutputAPI, links: list[str]) -> dict[RuleName, Verdict]:
```
jtoman, line 54: oh, right, I forget this information is recoverable from the POU, ignore what I said earlier about recording the rules.

composer/spec/source/report_prover.py
```diff
@@ -44,21 +44,38 @@ def _fetch(api: ProverOutputAPI, link: str) -> dict[RuleName, Verdict]:
...
+    """rule_name -> `Verdict` across the runs that account for one spec, ``links`` newest first.
+
+    Newest run wins per rule, rather than the most terminal outcome: a rule that timed out in
+    one run and verified in a later scoped re-run is verified. Within a single run the rollup
+    stays `Verdict.merge`'s — that is where "most terminal wins" is the right answer.
+    """
+    verdicts: dict[RuleName, Verdict] = {}
+    for link in links:
+        for name, v in _fetch(api, link).items():
+            verdicts.setdefault(name, v)
```
jtoman, line 64: EHHHH this is relying on the fact that `covering_run_links` returns the links in *reverse* execution order (most recent first). That's *fine* I guess, but I'd document loudly, otherwise this looks wrong!

composer/spec/cvl_generation.py
```diff
@@ -119,6 +124,15 @@ def output_link(self) -> str | None:
         ``None`` when no run link was produced. Drives the report's ``run_link``."""
         return _output_link(self.final_link)
 
+    @property
+    def covering_output_links(self) -> list[str]:
+        """``covering_links`` in ``/output/`` form, newest first, falling back to the last run
+        when the covering set is empty. What the report reads verdicts from."""
```
jtoman, line 130: truly, why would the last run link be available by the covering link empty set here be empty? Even having this code suggests that this state reachable, it shouldn't be, and if it is, that's a bug we should fix, not carry around crufty code that doesn't serve anything.

Honestly, we might want to drop the `output_link` field entirely, including the "last_run" field from the cvl author's state; it's superseded by the (richer) representation of multiple possible runs.

jtoman (approved):
approving with nits I trust you to change/address

composer/spec/source/report/schema.py
shellygr: Claude answers: Not done. Master records the active plugins in the run tags, and those live in a database inside the VPC. The report is the only artifact a user gets, so the field stays here too.

composer/spec/source/prover.py
shellygr: Claude answers: Not done, and your later comment says the same: the output library already returns the rules of each run, so the link list does not have to carry them.

composer/spec/source/prover.py
shellygr: Claude answers: Done in 8d090ec4. The reminder now lists the tools the plugins gave the author and says nothing about what they are for, and the state field and the keyword are called plugin_tools.

composer/spec/source/report_prover.py
shellygr: Claude answers: Noted, I dropped that one.

composer/spec/source/report_prover.py
shellygr: Claude answers: Done in f19cb8c9. The parameter is called newest_first, its docstring names covering_run_links as the function that produces that order, and a comment at the setdefault says why the first verdict found is the one to keep.

composer/spec/cvl_generation.py
shellygr: Claude answers: The fallback is gone in 331a5c6a: covering_output_links now returns only the covering links, because a published result always has a run at the spec state it published. Dropping output_link and the author's last-run field is a bigger change, so not in this PR: run_link is part of the Formalized protocol that every backend implements, it feeds the per-component prover links in the report and the console summary, and a budget-curtailed partial has no covering run but still shows its last run link. A local warm cache written before this PR has no covering links and now renders UNKNOWN rows, named in the coverage warnings. The cloud is not affected, because the root cache key includes the build.

PR #234, opened 2026-09-16
CloudRunner: unified early-stop infra (preprocessing watchdog + stop-on-first-violation)

jar-ben:
Supersedes #84 and #178 — merges both onto one shared mechanism.

## Problem

Both PRs add the *same* thing: while polling a still-running cloud job, inspect its partial state each tick and **cancel it early** instead of waiting out the client's 150-minute wait loop.

- **#84 — preprocessing watchdog:** a job stuck in prover preprocessing sits `RUNNING` with no rule results until the prover's ~2h global timeout. Cancel it once its treeview shows no rules past a budget.
- **#178 — stop_on_first_violation:** one `VIOLATED` rule is enough to trigger a result revision, so cancel instead of finishing the rest.

They edit the same poll branch and change the same return signature incompatibly. This builds the common infra once and plugs both in.

## Approach — shared early-stop layer (`utils/early_stop.py`)

- **`EarlyStop`** — a checker's decision: `reason`, `cancel` (loop-owned, opt-in), `payload` (data a post-cancel refetch can't recover, e.g. the partial checks), `job_status`, `drop_submission_cache`.
- **`EarlyStopChecker`** — a stateful per-job instance bound to `(prover_api, job_url)`; `observe(job_info) -> EarlyStop | None`. Holds its own clocks/latches; runs in the poll executor.
- **`PreprocessingCheck`** — adapter over the unchanged `PreprocessingWatchdog` (treeview state machine).
- **`FirstViolationCheck`** — cancels on the first `VIOLATED` rule, carrying the partial checks. Rate-limited (30s) and self-gating: `get_all_checks` is empty during preprocessing, so it no-ops until rule checking starts — no cross-checker signal needed.

### `cloud_runner.py`

- `_wait_for_job_completion_with_api` returns `(success, early_stop)`; start/finish times are no longer threaded back — the caller reads them from the `get_job_info` it already fetches.
- One place consults an ordered checker list per tick (each `observe` under a 60s guard, first decision wins) and performs the cancel when asked; one place builds the `ProverResult` generically from the `EarlyStop` fields.
- New `stop_on_first_violation` ctor arg. Env knobs: `AUTOSETUP_PREPROCESSING_BUDGET_SECONDS` (1800; `0` disables), `AUTOSETUP_PREPROCESSING_PROBE_SECONDS` (90), `AUTOSETUP_VIOLATION_PROBE_SECONDS` (30).

### Downstream

- `JobStatus.PREPROCESSING_TIMEOUT` + `ProverResult.is_preprocessing_timeout`.
- `on_job_problem` skips conf workarounds for any early-stopped result (not a conf-fixable failure).
- Reporter emits a dedicated preprocessing-timeout row and no longer crashes on a cancelled job's missing report.
- `prover_runner._checks_to_rule_results` reuses the checks→rule-results conversion for captured partial checks.

Kept client-side (per discussion): the prover's own `global_timeout` doesn't reliably cut off preprocessing stalls, so the client watchdog is what actually bounds them.

## Testing

- `test_preprocessing_watchdog.py` (state machine) + `test_early_stop.py` (both checkers, wait-loop integration incl. both-enabled, retry suppression, reporter rows).
- Full non-expensive suite green; `pyright` clean on changed files.


certora_autosetup/utils/cloud_runner.py
```diff
@@ -50,6 +56,17 @@ class CloudProverRunner(ProverRunner):
     # Timeout for job completion including job queue time - 150 minutes
     JOB_TIMEOUT_SECONDS = 7200 + 1800
 
+    # Early-stop defaults (see early_stop.py). All env-overridable.
+    #
+    # Preprocessing watchdog: budget is time spent in RUNNING without the prover producing a
+    # treeview. Healthy jobs produce one within minutes; a preprocessing blowup never does and
+    # would otherwise burn the whole JOB_TIMEOUT_SECONDS. Budget <= 0 disables the watchdog.
+    PREPROCESSING_BUDGET_SECONDS = 1800
+    PREPROCESSING_PROBE_SECONDS = 90
```
jar-ben, line 66: we don't need an env knob for `PREPROCESSING_PROBE_SECONDS` nor for `VIOLATION_PROBE_SECONDS` = just define these as constants in the code. 

env knob for `PREPROCESSING_BUDGET_SECONDS` is useful.

certora_autosetup/utils/cloud_runner.py
```diff
@@ -821,22 +851,76 @@ async def _wait_and_parse_job_results(
...
+                # loop cancelled it if the checker asked). Build a non-success result carrying
+                # the checker's reason and any partial results it captured.
+                job_handle.status = early_stop.job_status
+                error_msg = (
+                    f"Job stopped early ({early_stop.reason}): {early_stop.message}; "
+                    f"job {'cancelled' if early_stop.cancel else 'left running'} after {duration:.1f}s"
+                )
+                log_with_contract(
+                    self.component,
+                    "error",
```
jar-ben: why "error"? if this is the terminate on first violation, it's not an error. violations are not always bad. also, long preprocessing time is also not always bad (this is generic cloud_runner infra, not just autosetup)

certora_autosetup/utils/cloud_runner.py
```diff
@@ -821,22 +851,76 @@ async def _wait_and_parse_job_results(
...
+                )
+
+                # Drop the cached submission so a deliberate future run submits fresh instead
+                # of re-attaching to the cancelled job.
+                if early_stop.drop_submission_cache and not self.disable_cache:
+                    await self._remove_submission_cache(cache_key)
+
+                return ProverResult(
+                    job_handle=job_handle,
+                    success=False,
```
jar-ben, line 892: why do we set `success` to False? again, getting a quick violation might not mean a failure. perhaps adding a `stopped_early` boolean parameter would be much better.

certora_autosetup/utils/cloud_runner.py
```diff
@@ -996,19 +1138,31 @@ async def _wait_for_job_completion_with_api(
...
-                        return False, prover_start, prover_finish
-                    # If status is 'RUNNING', 'QUEUED', etc., continue waiting
+                        return False, None
+
+                    # Status is non-terminal ('RUNNING', 'QUEUED', ...). Consult the early-stop
+                    # checkers; if one fires, cancel the job (when it asked) and end the wait.
+                    early_stop = await self._run_early_stop_checkers(early_stop_checkers, job_info)
+                    if early_stop is not None:
+                        self.log(
+                            f"⏱ Early stop ({early_stop.reason}) for {job_url}: {early_stop.message}",
```
jar-ben, line 1147: not a warning, INFO at most.

certora_autosetup/utils/early_stop.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python3
+"""
+Shared early-stop infrastructure for cloud prover jobs.
+
+While a cloud job is still RUNNING, the poll loop consults an ordered list of
+``EarlyStopChecker``s once per tick. A checker inspects the job and may decide the wait
+should end now — before the job reaches a terminal status and before the loop's own
+150-minute budget runs out. Each checker is a per-job instance bound at construction to
```
jar-ben: > 150-minute
Don't talk about constants here that might change. it's irrelevant. just describe what early_stop does.

certora_autosetup/utils/early_stop.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@dataclass
+class EarlyStop:
+    """A checker's decision to stop waiting on a still-running job.
+
+    Attributes:
+        reason: Short machine tag identifying the stop condition, one of the ``REASON_*``
+            constants. Recorded in the result's ``output_data``.
+        cancel: When True the poll loop cancels the cloud job (the common case); False stops
```
jar-ben: >the common case
common now, but can change. drop this commented assumption

certora_autosetup/utils/early_stop.py
```diff
@@ -0,0 +1,174 @@
...
+            waiting but leaves the job running on the cloud.
+        payload: Data captured at detection that a post-stop refetch may no longer return
+            (e.g. the partial checks holding the violation, since ``get_all_checks`` on a
+            just-cancelled job can come back gappy). Opaque to the loop; the result builder
```
jar-ben: > can come back gappy
Have you tested this claim? do you have a proof? I would be surprised if this is the case. Clean solution would be to collect the results later same as we do in the standard pipeline; don't pass any untyped payload unless we need it and unless is has a dedicated consumer (in that case, it would have to be tied to the particular EarlyStopChecker which I don't think we need now).

certora_autosetup/utils/early_stop.py
```diff
@@ -0,0 +1,174 @@
...
+            )
+        return None
+
+
+class FirstViolationCheck:
+    """EarlyStopChecker: cancel the job the moment any rule is VIOLATED.
+
+    One violation already suffices to trigger a result revision, so there is no value in
+    waiting for the remaining rules. The partial checks seen at detection are carried in the
+    :class:`EarlyStop` payload, because a ``get_all_checks`` refetch on a just-cancelled job
```
jar-ben: > The partial checks seen at detection are carried in the
    :class:`EarlyStop` payload, because a ``get_all_checks`` refetch on a just-cancelled job
    can come back gappy

again the unverified claim

certora_autosetup/utils/job_problem_fixes.py
```diff
@@ -140,6 +140,11 @@ def on_job_problem(
     """
     if result.success:
         return False
+    if result.output_data.get("early_stop_reason"):
```
jar-ben: as said above, I am not convinced result.success should be set to false on early stop. it depends on how we currently interpret result.success - it is that the prover run successfully finished, or that the ProverJob (our CloudRunner definition that includes also early stops) successfuly finished?

certora_autosetup/utils/runner_types.py
```diff
@@ -46,6 +46,10 @@ class JobStatus(Enum):
     COMPLETED = "completed"
     FAILED = "failed"
     CANCELLED = "cancelled"
+    # Cancelled by the preprocessing early-stop check: the prover never produced a treeview
+    # (never finished preprocessing) within the watchdog budget. Distinct from FAILED so
+    # retry logic can skip conf workarounds that cannot help.
+    PREPROCESSING_TIMEOUT = "preprocessing_timeout"
```
jar-ben: there is inconsistency. we show teh preprocessing timeout case but not the first violation (early stop) case. either we include both, or not, or one "EARLY_STOPPED" entry with metadata in the result (which we perhaps already do I think?)

certora_autosetup/utils/cloud_runner.py
jar-ben: Claude answers: kept `success` meaning "the prover ran to a normal terminal completion" — its established meaning across ~30 call sites (it gates result caching at cloud_runner.py:212/305 and the PASSED/FAILED counts in parallel_autosetup / conf_runner). An early stop is not that, so `success=False` keeps a first-violation run from being counted as a pass or cached as a full result, and keeps retry logic from treating it as a normal completion.

To carry the distinction you are asking for, I added a typed `ProverResult.stopped_early` (+ `early_stop_reason`) — so a consumer that wants to act on the violation rather than retry can, without overloading `success`. And per the runner_types thread, JobStatus is now a single `EARLY_STOPPED` with the reason as metadata (no per-reason status). `on_job_problem` now keys off `result.stopped_early`.

If you would rather `success` mean "the ProverJob (incl. deliberate early stops) finished as intended", that is a broader semantic change across those ~30 consumers — happy to do it as a follow-up, but I kept this PR to the narrower, lower-risk definition.

certora_autosetup/reporting/reporter.py
```diff
@@ -602,7 +602,36 @@ def _collect_sanity_rows(
             except Exception as e:
                 self.log(f"Failed to read config for {contract_name}: {e}", "WARNING")
 
-            jr = self._prover_api.get_job_report(result.job_url)
+            if result.is_preprocessing_timeout:
+                # The job was cancelled by the preprocessing watchdog before producing any
+                # results — there is no job report to fetch; emit a dedicated row instead.
+                rows.append(_SanityRow(
```
shellygr, line 608: nit: wondering why we can't just dump the whole config instead of destructuring it here

shellygr (approved):
one nit above

certora_autosetup/reporting/reporter.py
jar-ben: We could, but it can be large. So, we are selecting only flags that are usually relevant. We used this schema in previous autosetup reports, so I would not change it now (though it's not so important). 

But there was a code duplication - I moved the fields extraction code to a shared helper.

PR #236, opened 2026-09-16
Pass exclude_rules through the plugin-facing prover runner

shellygr:
`WrappedProverRunner.run` is the `ProverRunner` a plugin receives through `CVLAuthorState`. It called `setup_prover_config_in` without `exclude_rule`, a required keyword-only argument, so every plugin prover call raised `TypeError: setup_prover_config_in() missing 1 required keyword-only argument: 'exclude_rule'` before it staged anything. The dz-strategy plugin's `lemma_prover` tool is the caller that hit this.

The runner and the `ProverRunner` protocol now take `exclude_rules` next to `rules`, the way `verify_spec` scopes its own runs, and pass it through. `tests/test_wrapped_prover_runner.py` drives the runner against a mocked `run_prover`: all four cases fail on master with the error above and pass with this change. Pyright reports no errors.


composer/spec/source/author.py
```diff
@@ -692,6 +692,7 @@ async def run(
         callbacks: ProverCallbacks,
         tool_call_id: str,
         rules: list[str] | None = None,
+        exclude_rules: list[str] | None = None,
```
jtoman, line 695: I was going to ask why the typechecker didn't help us here. Answer: `**config` means that the typechecker just sort of assumes that any required kwargs that aren't provided explicitly come through that `**config`. that's disappointing and the only way to fix it is to change the calling convention which is a "whole thing".

*sigh*. Alright then.

jtoman (approved):

PR #237, opened 2026-09-16
Add the timeout resolver's run workspace and its adapters

shellygr:
Stacked on #236 (the `exclude_rules` pass-through). Merge that first.

This is the AutoProver half of the standalone timeout resolver. When there is no pipeline, the DZ chain needs a workspace, and this rebuilds one from a prover run. It is a new package, `composer.resolver`, with no change to existing code.

`workspace.py` fetches a run's sources and tree view through ProverOutputUtility. The files land in the layout that a tarball extraction produces. It finds the run directory by its `.cwd` marker and reads the canonical `run.conf`. It records the run's verdict per rule path and its global timeout. Anything missing or ambiguous is a `WorkspaceError` that names it.

`runner.py` implements the plugin-facing `ProverRunner` over that workspace. The pipeline's runner forces `parametric_contracts`, `optimistic_loop` and `rule_sanity` onto the conf, and it stages the spec under `certora/specs/`. That changes the settings the rule timed out under, and it changes the location its CVL imports resolve against. This runner keeps the conf as the run recorded it. It removes only the keys that belong to one invocation: `msg`, `rule`, `exclude_rule`, `compilation_steps_only`, `global_timeout`, `server`, and `wait_for_results`. It writes the spec buffer over the spec at its original path for the duration of the run, and it verifies under the original timeout.

`preflight.py` compiles the run once with `compilation_steps_only` before any search. `artifacts.py` writes registered artifacts to a directory. `patches.py` turns the agent's VFS overlay and spec buffer into unified diffs against the fetched run. `author_state.py` assembles the `CVLAuthorState` with a proposer that only records proposals.

`tests/test_resolver_workspace.py` covers all of it against a fetched layout written into `tmp_path`, with `run_prover` and `certoraRun` mocked: 25 tests. The non-expensive suite passes (1348) and pyright reports no errors.


shellygr:
Claude answers: closing. This package belongs in dz-strategy, which is where it moves; only #236 needs to land in AutoProver.

PR #238, opened 2026-09-16
prover: retry a dropped connection when fetching a finished cloud job

ericeil:
## What

Wraps the cloud results fetch (`fetch_sources_and_treeview_files`) in a bounded retry: three attempts, doubling backoff.

## Why

Retrieving a succeeded job's sources and tree view is a pure read of immutable artifacts, but it makes hundreds of requests. `prover_output_utility` mounts a urllib3 retry adapter, which covers a request that fails *before its response begins*. It does not cover a peer that goes away mid-transfer, or one that outlives those attempts. Above that, nothing retried at all: `install_retry_policy` classifies LLM-provider failures, so a prover-API blip propagates out of the authoring graph and takes the unit's run down with it.

This is not hypothetical. A four-hour run lost one of its three parallel formalization units to `ProverAPIError: Remote end closed connection without response` while fetching a job that had **already succeeded**, on that unit's first submission. The cost is asymmetric — the work is discarded immediately, and the loss is discovered whenever the rest of the run finishes.

## What it does and does not retry

- **Retried:** `ProverAPIError` and `OSError`. Both, because `requests.RequestException` is an `IOError` subclass, so transport failures that escape the library's wrapping land in the same place.
- **Not retried:** `JobNotFoundError`, `AuthenticationError`, `InvalidJobError`, `ParseError` — a second attempt cannot change any of them, and retrying spends the backoff to fail identically.
- **Safe to re-ask:** the artifacts are immutable, and `prover_output_utility` writes a completion marker for each half as it lands, so a retry resumes from those marks rather than re-downloading what already arrived.

## What is unchanged, deliberately

Exhausting the attempts still propagates. Reporting a fetch failure to an authoring agent as a *verification* failure would have it rewrite a specification that was never wrong — the same reasoning the `CloudJobError` message in `run_prover` already carries. Making a fetch failure legible to the agent is a separate decision, not this PR's.

## Tests

`tests/test_cloud_fetch_retry.py` — five unit tests, no network. They assert the shape of the retry rather than any timing: a transient failure is re-asked, the backoff grows, an unreachable endpoint still raises after the attempts are spent, a missing job is not retried at all, and a transport error outside the library's wrapper is.

Gate: `pytest -m "not expensive"` 1334 passed, `pyright` 0 errors.


ericeil:
Oh, I see Jaroslav already opened an almost identical PR https://github.com/Certora/AutoProver/pull/223

PR #239, opened 2026-09-16
sandbox: one scratch directory, a readable git config, an unreadable output

ericeil:
Some updates to the way we handle sandboxed cargo builds, for the Solana and Soroban Prover work.

## One scratch directory

`composer/layout.py` declares `CERTORA_DIR` and `INTERNAL_DIR` — what a user keeps, and what a user ignores. They live there rather than in `spec/gen_types.py` (which declares the per-backend subdirectories under them) because `composer.sandbox` needs `INTERNAL_DIR` and stays pydantic-free, the convention already recorded at `composer/sandbox/config.py:20`.

The sandbox's private `CARGO_HOME` and tmp dir move under it (`.sandbox_cargo` → `.certora_internal/sandbox/cargo`), so everything a run generates sits beneath the one directory a project already ignores.

## An unreadable output

`RUST_FORBIDDEN_READ` now withholds that directory **anywhere in the tree**, not only at the root, and matches the **directory entry itself**, not only its contents.

A confined build's private `CARGO_HOME` holds an entire cargo registry — measured at 730 MB on one real Solana program — and it lands inside whatever workdir the build runs in, which need not be the project root. With only the anchored `^.certora_internal/` pattern, one `list_files` returned 28,904 lines, 28,739 of them from that tree, and the agent's next request was 2,272,575 tokens against a 1,000,000 limit. Nested `target/` goes with it, for the same reason the anchored pattern missed it; a generated crate's own `target/` was ~900 MB in that case.

Matching the bare entry is a separate fix from matching at any depth, and it is the one that makes the pruning work. graphcore short-circuits a subtree by testing the directory's own name — `DirBackend.dump_to` hands `copytree`'s `ignore` callback the entry names — so a contents-only pattern like `^\.certora_internal/.*` gets the right answer by walking all 730 MB and rejecting it file by file. `fs_forbidden_read` matches `.certora_internal` itself (via `fs_withheld_subtree`, which exists for exactly this) and prunes at the root; the Rust rule now does too, so the two really are the same rule rather than merely looking like it.

The directory is withheld whole rather than subdirectory by subdirectory, because the next scratch tree to grow without bound is the one nobody remembers to add.

This extension was previously deferred, in a `NOTE:` on `RUST_FORBIDDEN_READ`, to "the backend that runs confined Rust builds" — on the grounds that nothing in the front half, and nothing in the Rust application framework, creates these directories. That backend now exists and its CLI hands this pattern to its source tools, so the note becomes the rule and the `NOTE:` is gone.

### The plumbing that carries it

`cli_pipeline` grows a required `ecosystem` parameter and takes `forbidden_read` from `ecosystem.language.default_forbidden_read`, replacing the hardcoded `fs_forbidden_read` at three sites (`cli.py`, `foundry/entry.py`, `spec/source/autoprove_common.py`). The exclusion rule can then no longer disagree with the ecosystem the pipeline is driven with — the Solidity rule on a Cargo project admits `target/` and withholds the `.json` an Anchor IDL needs.

`ecosystem` moves *up* from `Continuation.__call__` to the context manager, since it is now needed before the continuation runs; `Continuation` and `cli_pipeline` carry the `App`/`Main`/`U` type parameters to match. Both existing callers pass `EVM`, whose `default_forbidden_read` **is** `fs_forbidden_read`, so behaviour is unchanged for EVM.

## A readable git config

Cargo resolves a `[patch.crates-io]` git source through libgit2, and libgit2 reads the global git config before it will open the cached repository at all. Denied that read it does not degrade to "no user config" — it reports the source as unopenable, which cargo surfaces as:

```
Unable to update https://…: can't checkout from '…': you are in the offline mode (--offline)
```

A message about the network, for a cache that is fully warm, and one no amount of pre-fetching fixes. `git_config_ro_paths` grants the three spellings of that file read-only — `$GIT_CONFIG_GLOBAL`, `$XDG_CONFIG_HOME/git/config`, `~/.gitconfig` — and nothing around them: the files, not the home directory. Landlock's PathBeneath is hierarchical, so granting `~/.config` would hand an untrusted `build.rs` every other application's configuration, and granting `$HOME` would hand it everything. A host with no global config is the ordinary case, not an error.

The residual exposure is the git config itself, which can name a credential helper and, in a badly configured checkout, carry a token in a `url.*.insteadOf`. That is a real if narrow leak and the price of building a git dependency at all. The alternative considered and rejected was a private `$HOME`: it silently relocates rustup's and `cargo-build-sbf`'s toolchain lookups and fails much later, mid-build, complaining about downloading Rust.

## The compose overlay

`--profile sandbox` is now required alongside the consuming service's profile. A `depends_on` no longer implicitly enables the target's profile, and the refusal — "service autoprove depends on undefined service run-confined-build" — names neither this overlay nor the profile it wants, so the header documents it.

## Tests

27 tests across the two files, 9 of them added here: the sandbox's `CARGO_HOME` and tmp dir are unreadable at the root and nested, a withheld directory matches as a bare entry so a walk can prune it, `src/targets.rs` and `my_target/` stay readable, the git-config grant covers each spelling and drops what is absent, and the policy grants the config rather than the home.

Gate on this branch: `pytest -m "not expensive"` 1338 passed, `pyright` 0 errors.


composer/sandbox/recipes.py
```diff
@@ -93,6 +103,36 @@ def shared_cargo_ro_paths(cargo_home: str | Path) -> tuple[Path, ...]:
...
+    Cargo resolves a ``[patch.crates-io]`` git source through libgit2, and libgit2 reads the global
+    config before it will open the cached repository at all. Denied that read it does not degrade
+    to "no user config": it reports the source as unopenable, which cargo surfaces as
+
+        Unable to update https://…: can't checkout from '…': you are in the offline mode (--offline)
+
+    — a message about the *network* for a cache that is fully warm, which no amount of pre-fetching
+    fixes. Any project patching a dependency to a git repo hits this, an Anchor fork included.
+
+    Files, never ``$HOME`` and never ``~/.config``: Landlock's PathBeneath is hierarchical, so a
```
chandrakananandi: could we make this comment a bit simpler? It's a bit hard for me to understand because the models tend to convolute words too much

chandrakananandi (approved):
Looks good, just one remark about a code comment being a bit confusing and hard to understand.

composer/sandbox/recipes.py
ericeil: Done

PR #240, opened 2026-09-16
Support for invoking the Solana and Soroban Provers

ericeil:
Second in a series splitting long-lived Solana/CVLR work into independently reviewable pieces (the first is #239). Two halves of one seam: **which Prover CLI a run submits to**, and **how that chain's counterexamples render**. Everything between those two points — cloud polling, the treeView parse, the verdict roll-up — stays chain-neutral, which is what keeps this small.

## Which CLI

`certora_cli` ships one entry point per Prover — `certoraRun`, `certoraSolanaProver`, `certoraSorobanProver` — each with its own build step, since one compiles Solidity and another compiles a Rust crate. They share the signature `list[str] -> CertoraRunResult | None`, so the entry point is a lookup rather than an import:

- `ProverApp` names the three; `_PROVER_ENTRIES` maps each to its module and function.
- `import_prover_entry(app)` resolves one, honoring `$CERTORA` exactly as `import_run_certora` did.
- `ProverEntry` names the signature they share, over the `CertoraRunResult` protocol in `stubs/certoraRun.pyi`, so the wrapper's use of the result stays checked and the one unverifiable step — `getattr` on a dynamically imported module — is a single explicit cast.
- `prover_app(name)` narrows an untrusted string to the literal. One caller: the subprocess wrapper, which reads the app out of its own `argv`. The parse belongs at that boundary so every in-process caller keeps a checked literal and only the process boundary pays for a runtime check.

**This half is inert for EVM.** The app is not defaulted — it selects the CLI and therefore the build step, where the chains differ most, and only the caller knows which ecosystem the spec was written for. So every caller names it: the autoprove pipeline passes `EVM.name`, following the ecosystem already driving the run rather than asserting it alongside, and the codegen executor and the munge compile check say `"evm"` outright. The wrapper resolves the same `certoraRun` entry it always did.

## Which frames

A counterexample stops being a rendered string on `RuleResult` and becomes data. `Counterexample` carries the trace, the assertion, and its source span — kept as fields rather than only inside the rendering, because whether a violation says anything about the *program* is decided from the assertion (`classify_violation`), and re-reading it out of the rendered text would be the same fact stored twice.

`TraceShape` is the per-chain part: which frames of a trace are worth showing. A trace is mostly runtime the property is not about, and each chain buries the counterexample under its own scaffolding. `dropped` removes a frame and its subtree; `elided` keeps the frame and replaces its subtree with a count of what went. A chain with no measured counterexample (soroban, and `move` on the analyzer CLI's wider ecosystem axis) gets `GENERIC_TRACE`, which claims nothing chain-specific.

Frames are named by the prover's own format template rather than by parsing rendered prose. A frame's `message.text` *is* a template — literal pieces interleaved with `'{n}'`, the counterexample's values riding alongside in `arguments` (`report/calltrace/sarif/Sarif.kt`) — so `MessageModel.frame` collapses the placeholders to `{}` and `render` fills them from `arguments`, and a shape entry is written the way the prover's format string writes it: `"__rust_alloc: {}"`. That replaces a `": '"` heuristic that guessed at where a value began, covered only one of the prover's conventions, and on a miss silently named a frame after its own arguments. The cost is that a shape entry is coupled to its frame's arity; every frame named in a shape today is fixed-arity, and a future miss surfaces as un-dropped noise rather than as a wrong name.

## How a run is configured

`ProverOptions` is the whole of what a run needs before its arguments: which CLI takes it (`app`), which server it goes to (`server`), and what budget the Prover was given (`global_timeout`).

The latter two used to live in a `list[str]` of CLI flags that the options then read their own meaning back out of: `cloud` was `"--server" in extra_args`, and `global_timeout` found the flag's index and parsed `extra_args[idx + 1]` — an `IndexError` on a trailing `--global_timeout`. That list was never a passthrough; `make_prover_options` is its only producer, per-run flags reach `run_prover` separately as `args`, and across every reference it held `--global_timeout <n> --server <s>` or nothing. So the two values it encoded are now fields. `cloud` stays a derived property, and `runtime_bound` names what the callers actually wanted — the budget the Prover was given, or our own patience when it was given none — so `DEFAULT_GLOBAL_TIMEOUT` reads as ours rather than as a flag.

**No timeout behavior changes here.** `make_prover_options` still gives a cloud run the resolved timeout and leaves a local run unbounded, `AUTOPROVER_GLOBAL_PROVER_TIMEOUT` resolves as before, and the subprocess and polling backstops extend the same number they always did. `cli_args()` renders the same tokens in the same order the old `extra_args` held.

There is one `ProverOptions`, not two. The codegen prover tool takes the same type `run_prover` consumes, carried on `ProverDeps` rather than reassembled from parts — a copy between two look-alike types is a shape the type system cannot check, since both directions are constructor calls where everything unmentioned takes a default, and `app` is exactly the sort of field such a copy drops. `ProverDeps` also names `keep_folder`, which is the VFS materializer's debug flag rather than a prover argument.

Two incidental cleanups ride along: `SpecBufferWithDraft` and its `WithCurrSpecAndDrafted` alias are gone (their only binders were the `get_spec_tool` / `get_cvl` overloads the write-side draft gate already removed), and one annotation in `thinking.py` is tightened.

## Review this as if it changed EVM behavior

It does, in three ways, and all are visible in the diff's test churn:

1. Every EVM trace renders through `TraceShape` (`EVM_TRACE`) — the same four frames dropped as the old hard-coded list, now matched by template. Rendering also becomes a single `re.sub` pass instead of a loop of `str.replace` in argument order, which fixes a latent bug: a value containing another argument's placeholder would be clobbered by the next iteration.
2. `cex_dump` is a derived property over `Counterexample`, and its text gains `<assert>` and `<source>` elements ahead of the trace, inside the same `<counterexample>` envelope it always had. An assertion in `tests/test_autoprove_report.py` moved for exactly that.
3. Every EVM run's options are a different shape. `cli_args()` renders the same tokens in the same order the flags were passed in before, which is what keeps the autosetup cache key — it hashes that joined list — stable rather than stranding every existing entry.

The claim is that no EVM verdict, link, report field or cache key changes — only the rendering an analysis prompt reads. That is the claim worth attacking.

## Fixtures

`tests/data/solana_cex/` holds unedited output from two real Solana Prover runs: a whole treeView (one VIOLATED, one ERROR, one VERIFIED) and one rule's output carrying a genuine property violation. They are what makes `tests/test_solana_cex_trace.py` the first test of `composer.prover.results` that does not submit a cloud job — `test_tree_parsing` is `expensive` and costs a cloud run per assertion.

That the Solana treeView needs no chain-specific parsing is measured here rather than assumed: these outputs parse through the same code EVM does.

## Deliberately not here

- The Anchor program the fixtures came from. Nothing in this PR compiles or loads it, and re-capturing the fixtures needs the Solana platform tools and a cloud job regardless; it lands with the end-to-end gate.
- Two unrelated improvements to `run_prover` that sit in the same file on the source branch: an analysis-free CEX handler, and putting the job link into the `CloudJobError` message.
- The autosetup cache key omits `app`. Latent while autosetup is EVM-only, and worth its own change rather than a silent cache invalidation buried here.

Gate: `pytest -m "not expensive"` 1349 passed (master baseline 1329, +20 from the tests here), `pyright` 0 errors.

One caveat worth naming: `tests/test_codegen_integration.py` is the only test that exercises the executor path `ProverDeps` is bound on, and it is `expensive`. It cannot pass where `solc` is not 0.8.29 — the mismatch surfaces as a `HarnessFakeLLM` tape exhaustion rather than a compiler error, which is how it presents on my machine. It fails identically at `origin/master`, so it is not a regression, but that path is unverified by execution here and wants a run somewhere with the right toolchain.


composer/authoring/judge.py
```diff
@@ -145,7 +145,7 @@ class JudgeInput(FlowInput, _JudgeExtra):
     pass
 
 
-def _wrote_rough_draft(s: JudgeState, _: Any) -> str | None:
+def _wrote_rough_draft(s: JudgeState, _: object) -> str | None:
```
jtoman: ???????

jtoman:
> a timeout is meaningless without a server to enforce it

This is so breathtakingly wrong? Timeouts are interpreted as flags by the prover jar and it will terminate itself in process if the timeout is exceeded??

composer/prover/core.py
```diff
@@ -58,20 +59,58 @@
 DEFAULT_GLOBAL_TIMEOUT: float = 7200.0
 
 
+@dataclass(frozen=True)
+class LocalRun:
+    """The run is verified by a prover on this machine, and is bounded only by our own patience:
```
jtoman: again, whether we want a global timeout locally is one question, but pretending a local run can't have a timeout bound is just wrong...

composer/prover/core.py
```diff
@@ -58,20 +59,58 @@
...
+    no ``--global_timeout`` reaches the CLI, and :data:`DEFAULT_GLOBAL_TIMEOUT` serves as the
+    subprocess backstop."""
+
+
+@dataclass(frozen=True)
+class CloudRun:
+    """The run is submitted to ``server``, which enforces ``global_timeout`` itself.
```
jtoman: > which enforces ``global_timeout`` itself

**incoherent grumbling**

composer/prover/results.py
```diff
@@ -211,41 +291,82 @@ def read_and_format_run_result(s: Path) -> dict[str, RuleResult] | str:
...
-    """
-    # Extract and format the message
+def _frame_name(text: str) -> str:
+    """A frame's identity, with whatever value it held stripped off.
+
+    The prover formats a value-bearing frame as ``name: '{0}'`` — the name is the frame, the
+    argument is the value it took in this counterexample. Matching a :class:`TraceShape` against the
+    whole text would make every single allocation its own frame."""
+    head, sep, _ = text.partition(": '")
+    return head if sep else text
```
jtoman: noooo, come on what? please let's do something different...

composer/spec/cvl_research.py
```diff
@@ -71,7 +71,7 @@ class _CVLResearchST(MessagesState, RoughDraftState):
 widget; pass ``None`` for top-level invocations."""
 
 
-def _wrote_draft(s: _CVLResearchST, _: Any) -> str | None:
+def _wrote_draft(s: _CVLResearchST, _: object) -> str | None:
```
jtoman: ??

composer/prover/ptypes.py
```diff
@@ -53,18 +53,95 @@ def pprint(self) -> str:
...
+#: entry costs a spuriously reported finding and can never suppress a real one.
+#:
+#: ``Cannot overflow`` is the Solana Prover's sound-signed-math pass. Under
+#: ``-solanaTACSoundSignedMath`` it annotates pointer arithmetic it believes cannot exceed 64 bits
+#: and then asserts that belief (``sbf/tac/TACModSimplifier.kt``, ``removeNoOverflow``); the
+#: assertion fails where a pointer is computed from a value the analysis could not pin down,
+#: typically an account's ``data_len`` that no rule bounded. An author can act on it — pin the
+#: length, or fall back to ``-solanaTACOptimisticOverflowOptimization`` — but a findings synthesizer
+#: handed it writes the prover's limitation up as a defect in the program under verification.
+_GENERATED_ASSERTIONS = ("Unwinding condition in a loop", "Cannot overflow")
```
jtoman, line 123: so, the reason we don't run into this with the evm prover is we never let it run without optimistic loops. Do soroban/solana not have a similar flag? do we expect to also run them under asserted loops?

composer/prover/core.py
```diff
@@ -92,16 +131,16 @@ def _resolved_global_prover_timeout() -> int:
         return default
 
 
-def make_prover_options(*, cloud: bool) -> ProverOptions:
+def make_prover_options(*, cloud: bool, app: ProverApp = "evm") -> ProverOptions:
```
jtoman: saying "this is a disaster if we let a default compile the wrong way" and then having a default that leads to exactly the same footgun here is a choice

composer/tools/prover.py
```diff
@@ -151,8 +157,7 @@ async def run(self) -> Command:
 
 async def _run_certora_prover(
     tool: CertoraProverTool,
-    cex_handler: CexHandler,
-    prover_opts: ProverOptions,
+    deps: ProverDeps,
```
jtoman, line 160: yeah it was silly to pass these as separate args...

jtoman (changes requested):

composer/authoring/judge.py
ericeil: I thought I'd reverted this! (this happened when I asked Claude to strengthen the type annotations in this change)

composer/prover/ptypes.py
ericeil: All of the Solana spec examples I have, have used asserted loops.  I'm not sure there's any real Solana-specific reason for this; it's more that the people building those specs had strong opinions about this.  But this isn't just about those assertions; the Solana prover adds other assertions for things like validating overflow assumptions in "sound math" mode.

composer/prover/core.py
ericeil: Yeah, this is nonsense; will fix

composer/prover/core.py
ericeil: indeed

composer/authoring/judge.py
ericeil: (And Claude got confused by some spurious unrelated changes that I hadn't fully separated from this change at the time...)

composer/prover/results.py
ericeil: I think this is better now

composer/prover/core.py
```diff
@@ -131,16 +101,16 @@ def _resolved_global_prover_timeout() -> int:
...
     certoraRun ``--server`` resolved from the deployment env."""
-    target: RunTarget = LocalRun()
-    if cloud:
-        target = CloudRun(
-            server=cloud_server_for_env(),
-            global_timeout=_resolved_global_prover_timeout(),
-        )
-    return ProverOptions(app=app, target=target)
+    if not cloud:
+        return ProverOptions(app=app)
```
jtoman, line 108: no global timeout tho?

composer/prover/results.py
```diff
@@ -50,10 +50,31 @@ class SarifArgs(BaseModel):
...
+        A counterexample's values live in ``arguments``, never in ``text``, so the template already
+        *is* the identity: all 33 ``__rust_alloc`` frames of one measured trace carry the same
+        ``text`` and differ only in the argument they took."""
+        return _PLACEHOLDER.sub("{}", self.text)
+
+    def render(self) -> str:
+        """The template with this counterexample's values substituted for its placeholders."""
+        return _PLACEHOLDER.sub(
+            lambda placeholder: f"'{self.arguments[int(placeholder.group(1))].value}'", self.text
+        )
```
jtoman, line 76: holy crap this is cursed. Not a criticism of your code, just ... what IS this language/stdlib dude.

jtoman (approved):

composer/prover/results.py
ericeil: yeah...

composer/prover/core.py
ericeil: Adding that for local runs seemed a bit out of scope for this PR (and I didn't feel like testing it :) )

PR #243, opened 2026-09-18
Build and inspect a Solana crate from Python

ericeil:
`composer/cargo/` reads, builds, and inspects a Solana crate from Python.

- **metadata** — typed `cargo metadata`. Which crate owns a source file, and which version of a dependency the build resolves (`RUST_FORBIDDEN_READ` hides `Cargo.lock` from agents).
- **session** — a reused workdir and a host-target `cargo check`. Each run gets a private `CARGO_HOME`, so the graph is fetched once per session rather than once per compile.
- **sbf** — the verification build (`cargo certora-sbf`) and the JSON manifest it prints. The backend owns the prover's `build_script` so the build stays sandboxed.
- **depinfo** — which sources a build compiled, from rustc's `.d` files. An edit in a file the `certora` feature gates out does not reach the build.
- **symbols** — functions a built program defines, demangled as the prover's tuning files name them.

Does not import `composer/spec/cvlr/`. `tests/test_cvlr_symbols.py` checks demangling against `rustfilt` on recorded `llvm-readelf` names; it does not need a toolchain.


composer/cargo/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Cargo workspace, session, and chain-specific builds.
```
chandrakananandi, line 1: why is it in `composer` vs `scripts`? I ask because julian's been putting his stuff here: https://github.com/Certora/AutoProver/tree/julian/soroban-setup and I am trying to decide which is the right place. Maybe composer?

chandrakananandi (commented):
Thanks for the PR! I am trying to understand the overall goal especially in the context of what's happening for soroban: is this a sort of "autosetup" for solana?

composer/cargo/__init__.py
ericeil: I interpret `scripts` as a place for scripts that are used to "build" AutoProver itself, not code that runs as part of AutoProver in the cloud.

ericeil:
Closing this PR; I'll pull this code in as it's needed in more feature-oriented PRs later.

PR #246, opened 2026-09-21
Log the category behind Anthropic streaming refusals

jar-ben:


composer/llm/anthropic.py
```diff
@@ -268,6 +269,54 @@ def should_retry(self, exc: Exception) -> bool:
             return payload_error_type(exc.body) in RETRYABLE_ERROR_TYPES
         return False
 
+_logger = logging.getLogger(__name__)
```
jtoman, line 272: I am almost positive this can be accomplished with a model callback alongside the `standard_callbacks` we instantiate.

jtoman (changes requested):

